PR #6, opened 2017-12-18
Updated button so it can be applied to a button or anchor and removed…

jtrujill:
I still need to update the button's unit tests and I also need to update the button demo's documentation, but this pull request will allow others to start making conversation


src/app/lib/button/button-demo/button-demo.component.html
```diff
@@ -1,75 +1,83 @@
...
+        <button hc-button title="Primary-Alt3 Button" color="primary-alt3">Primary Alt3 Button</button>
+    </div>
+    <h5>
+        Secondary Buttons
+    </h5>
+    <button hc-button title="Secondary Button" color="secondary">Secondary Button</button>
+    <h5>
+        Disabled Buttons
+    </h5>
+    <button hc-button title="Disabled Button" color="primary" disabled>Disabled Button</button>
```
corytak: Love how semantic the markup is and how you can use the native apis like disabled. Very nice.

src/app/lib/button/button-demo/button-demo.component.ts
```diff
@@ -1,90 +1,15 @@
...
+    styleUrls: ['./button-demo.component.scss'],
+    encapsulation: ViewEncapsulation.None
 })
-export class ButtonDemoComponent implements OnInit {
+export class ButtonDemoComponent {
+    showTemplate: boolean = true;
 
-  public showTemplate: boolean = true;
-
-  public primaryButton1Options: ButtonOptions = {
```
corytak: Love the removal of all of these unnecessary objects.

src/app/lib/button/button.component.ts
```diff
@@ -1,16 +1,42 @@
...
-  selector: 'hc-button',
-  templateUrl: './button.component.html',
-  styleUrls: ['./button.component.scss']
+    selector: 'button[hc-button], a[hc-button]',
+    template: '<ng-content></ng-content>',
+    styleUrls: ['./button.component.scss']
 })
-export class ButtonComponent {
-  @Input() public options = new ButtonOptions();
-  @Output() public onButtonClick = new EventEmitter<boolean>();
```
corytak: This is a nice change that the consumer can use the native `(onClick)` functionality rather than a pass through EventEmitter

src/app/lib/button/button.component.ts
```diff
@@ -1,16 +1,42 @@
-import { Component, EventEmitter, Output, Input } from '@angular/core';
-import { ButtonOptions } from './button-options';
+/* tslint:disable:component-selector */
```
corytak: At some point we'll want to establish our tslint rules and perhaps have an [.editor config file](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig) that we check in with the library pull requests don't have as much noise from formatting.

src/app/lib/button/button.component.ts
```diff
@@ -1,16 +1,42 @@
...
+// https://github.com/mgechev/codelyzer/issues/178#issuecomment-265154480
+
+import {Component, ElementRef, Input, OnChanges, Renderer2, SimpleChanges} from '@angular/core';
+
+export function throwErrorForInvalidButtonColor() {
```
corytak: Any reason in particular why this function exists outside of the ButtonComponent class?

corytak:
@jtrujill This looks really great. I like the direction of using native HTML elements as much as possible. The removal of bulky options objects will make this so much easier to use. 

One thing we'll want to look at as a team is creating some formatting standards so the code changes are easier to tease apart from formatting changes.

src/app/lib/button/button.component.ts
jtrujill: The reason for this is so that I can do unit testing to make sure that the error gets thrown appropriately

src/app/lib/button/button.component.ts
jtrujill: I totally agree and didn't realize until @waldnercharles had said that I'm using spaces instead of tabs and I'm using 2 spaces per tab. I created a new issue #8 that is intended to fix this issue. I'm going to hold off on getting this pull request merged in until that is resolved, since we're in no hurry

jtrujill:
I'll remove the font-awesome dependency and get that pushed to this pull-request. In general I think we should avoid using any other library within our own library to avoid version conflicts with the users of our library and also to reduce the overall size of the library. For example I don't believe that I library like underscore or ramda would be beneficial for this project. So I think we should have it as a policy that any external dependency should be frowned upon and requires a very good reason to make an exception to that

src/app/lib/button/button-demo/button-demo.component.html
```diff
@@ -1,75 +1,83 @@
...
-    </div>
-    <pre><code>&#x3C;hc-button
+    <div class="headerName">Angular Component</div>
+    <br>
+    <div id="HTML" *ngIf="showTemplate">
+        <div class="codeToggle">
+            <span class="toggleOptionOn" (click)="viewToggle('html')">HTML</span>
+            <span class="toggleOptionOff" (click)="viewToggle('ts')">TypeScript</span>
+        </div>
+        <pre><code>&#x3C;hc-button
```
benjanderson: this should be changed to show the proper usage of the button
ie instead of <hc-button> it should be <button hc-button

src/app/lib/button/button.component.scss
```diff
@@ -1,618 +1,137 @@
 @import '../colors.scss';
```
benjanderson: This is amazing how you condensed all of the styles into a reusable and maintainable file. Great work!

jtrujill:
I now have unit tests written for the button and anchor component. I had to create two components because the disabled property had to be handled differently for anchor since it doesn't natively support it. 

I also edited the demo documentation so that it shows the correct html usage and gives the correct parameters.

PR #9, opened 2017-12-23
Travis CI 

jtrujill:
I added a .travis.yml file so that we can get a CI pipeline in place to make sure all pull requests pass lint, build and unit tests. I'll need someone with more privileges to enable travis ci for Fabric.Cashmere before CI will work


jtrujill:
I would like to use slack notifications to notify us of the CI pipeline. However, I don't know if we want to use the cashmere channel or if we'd like to create another channel.

benjanderson:
A CI pipeline has already been created https://i.imgur.com/7omTLkt.png through vsts that runs lint and tetss and is being deployed to http://cashmere.azurewebsites.net/

You can see that the build is already associated with the check in to master so I opted not to create a slack integration

I am unfamiliar with travis, and am open to the idea

jtrujill:
@benjanderson
Travis ci is an open source CI pipeline that is used by probably most open source software on github. The reasons I like it is that the process is saved in code and is available for everyone to view and make edits to the process.

Travis will also run CI against all pull requests and will indicate to the reviewers whether it has passed or failed. This I think is the best reason for moving to Travis instead of vsts. The approval process would be smoother if we could validate that a pull request passes our pull request process before we try to make a commit to master. Also it requires that reviewers use both github and VSTS in order to manage the project, which I don't like the idea about doing

Another benefit is that Travis is free to use and there are no limits as long as the software is open source. We might as well save the company some money and also allow others such as myself the ability to easily edit the pipeline without requiring that everyone be given privileges to access it through VSTS

benjanderson:
I like your reasoning and it sounds like we have much to gain and nothing to lose by using travis ci. I still think it is reasonable to keep the existing CI pipeline to deploy to the website whenever master is checked into.

package.json
```diff
@@ -12,7 +12,8 @@
     "start": "ng serve",
     "build": "ng build",
     "test": "ng test",
-    "test-single-run": "ng test --single-run",
+    "ci:build": "ng build --prod --no-progress",
+    "ci:test": "ng test --single-run --code-coverage --no-progress",
     "release": "ng build --target production --build-optimizer",
```
corytak: We may want to have some different commands for building the demo app, e.g. `ci:build` and commands for building the library. Right now the `npm run package` command will run `node build.js` and then to publish the library to npm you would run `npm publish dist` with whatever prerelease tags we want to set.

jtrujill: I agree with that. I thought I remembered that people had agreed to wanting to do the npm packaging manually, at least at first. That would be an easy addition that we could do later, but want I wanted to achieve immediately was better info for reviewers on whether or not a pull request passes the basic set of tests that we have.

jtrujill:
@benjanderson 
If you make that change to vsts sometime today and let me know through slack when that’s done I’ll merge pull request #9.

This pull request will also require that either @benjanderson, @corytak or someone else with full access to the repo enable Travis CI for this repo. This link tells you how to do that https://docs.travis-ci.com/user/getting-started/

benjanderson:
Sounds good, @jtrujill I will enable it after the pull request is completed

PR #13, opened 2018-01-07
fix linting errors

benjanderson:


benjanderson:
sweet, as this fixes the build I am going to just complete this

PR #21, opened 2018-01-18
added npm-font-open-sans to package.json

jacquie:


corytak:
This gets to @joeskeen 's point about how things can get confusing when the docs and the lib files are all enmeshed into one directory structure. Styles.scss isn't packaged into the library build process and only styles the demo app. If we want to include this in a sass file that could be imported by apps using the library we should import it into src/app/lib/colors.scss or add a new .scss file in the lib folder or rename the colors.scss file to be something that's more generic.

andrew-frueh:
Ah that's a great callout @corytak - I see what you're saying about "styles.scss" being only for the demo app.  So this is a great topic for conversation at our Cashmere Open Space - and I think @joeskeen makes a good point about separating the demo from the library better.

But as to the point of this pull request, I like your suggestion Cory.  For the sake of clarity, my suggestion would be that we add another file called fonts.scss and import Open Sans and Font Awesome there.  Then adjust the demo site accordingly.  @jacquie - is that a switch that you can make?  Or @corytak it looks like this is just in a branch not in a separate fork, so maybe you can take care of it?

PR #23, opened 2018-01-19
Button UX Cleanup

andrew-frueh:
Updated pull request with some additional work I added to the button component.  Have added a "glyph" and "dropdown" option for the button and included examples and documentation on the demo page.  After checking the 40px height in some existing apps like MeasuresManager and PatientIntake, I realized it looks a bit too big.  So I've adjusted the standard size to 35px height, and have included a "large" boolean option that should be used primarily for modals - where the 40px size looks better.


jtrujill:
What if instead of using a boolean for large or small we just created another component that represented a small button? I feel this is similar to function design and that it's better to break something out into its own thing rather than passing a boolean to change its function. In the future we could choose to style the small button radically different from the large button. By keeping them as the same component it increases the difficulty of maintaining the button.

So right now the large button has a tag of <button hc-button></button>
What I'm proposing is adding this new component <button hc-button-sm></button>

Sorry for the late response

andrew-frueh:
I would be okay with that @jtrujill - honestly I'm not entirely sure there is a need for the large size.  The only place where I thought it would still make sense is as the Save button on modals.  But I'm not sure that's really necessary.  I think we just need to start dropping this component into our apps and see how it's working in context.  Then we'll know if we need more than the 35px size.  So I'm gonna remove the "large" tag for now until we decide we need it.  Btw - couldn't help but notice the "Comment" button here on GitHub is 35px...so I'm feeling good about that decision :-)

src/app/lib/button/button.component.ts
```diff
@@ -23,6 +23,7 @@ export class ButtonComponent implements OnChanges {
   private supportedColors = ['primary', 'primary-alt1', 'primary-alt2', 'primary-alt3', 'secondary', 'tertiary'];
 
   @Input() color: ButtonColor = 'primary';
+  @Input() size = false;
```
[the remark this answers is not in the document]
andrew-frueh: Hey @benjanderson - as per the above comments with @jtrujill - I decided to remove the size size elements altogether.  So these elements have been removed in the most current commit.

src/app/lib/button/button-demo/button-demo.component.html
```diff
@@ -121,6 +129,13 @@ <h5>
             </td>
             <td width=50%>Boolean value that enables/disables the button</td>
         </tr>
+        <tr>
+            <td style="font-weight: 600">class</td>
```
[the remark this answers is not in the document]
andrew-frueh: Good call - I've switched this column to use th instead of td and adjusted the css accordingly. Committing that change now.

src/app/lib/button/button.component.scss
```diff
@@ -94,17 +94,17 @@
...
-    border-color: $secondary-button-hover-border;
+    border: 2px solid $secondary-button-hover-border;
   }
   &:active {
     background-color: $secondary-button-pressed;
     color: $secondary-button-hover-text;
-    border-color: $secondary-button-pressed-border;
+    border: 2px solid $secondary-button-pressed-border;
   }
   &:focus {
```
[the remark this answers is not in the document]
andrew-frueh: Let's hold on this for now and talk more at OpenSpace.  I have some concern that adding transition effects on hover can make the buttons feel less responsive - almost laggy.  And effects on click can be perceived as cute or gimmicky.

src/app/lib/colors.scss
```diff
@@ -17,7 +17,7 @@ $slate-gray-400: #708090;
 $white: #fff;
 $black: #000;
 $third-column-gray: #5e676f;
-$primary-blue: #00AEEF;
+$primary-blue: #00AEFF;
```
corytak: This isn't really relevant to this pull requests. But what are your thoughts about the color names in this colors.scss? If there were ever some major change in or corporate color scheme `primary-blue` locks us into blue as our primary color. Would it make more sense for this name to be just `primary`?

src/app/lib/button/button.component.ts
```diff
@@ -12,7 +12,9 @@ type ButtonColor = 'primary' | 'primary-alt1' | 'primary-alt2' | 'primary-alt3'
 
 @Component({
   selector: 'button[hc-button]',
-  template: '<ng-content></ng-content>',
+  template: `<i *ngIf="glyph" class="fa {{glyph}} button-glyph"></i>
+  <ng-content></ng-content>
+  <i *ngIf="dropdown" class="fa fa-chevron-down fa-fw button-dropdown"></i>`,
```
corytak: I know there's been some discussion about not having font awesome as a dependency. I'm wondering what a solution would look like if we didn't have font awesome. I remember there being talk about shipping with our own font. That may be the way to go without having an explicit dependency. But this should work if we list FontAwesome as a PeerDependency.

benjanderson: I really like this idea, I propose we wait a little for the library to mature before we do something like this. As we are probably going to have quite a bit of churn until we have a solid set of icons

src/app/lib/button/button-demo/button-demo.component.html
```diff
@@ -19,7 +27,6 @@ <h5>
         Disabled Buttons
     </h5>
     <button hc-button title="Disabled Button" color="primary" [disabled]="true">Disabled Button</button>
-    <a hc-button title="Disabled Link" color="primary" [disabled]="true">Disabled Link</a>
     <h5>
         Tertiary Buttons
```
benjanderson: It would be nice to have the anchor component near here to compare to the tertiary buttons

src/app/lib/button/button.component.ts
andrew-frueh: I totally agree @corykon - a few of us were talking about it this morning and that's definitely where we want to get to.  I think as a starting point for this library we'll include FontAwesome as a dependency - because honestly I don't think I could give you a comprehensive list of all the icons we'd need in a HCGlyphs font yet.  I've made a note to make that a discussion topic at our next Cashmere OpenSpace.

src/app/lib/colors.scss
andrew-frueh: My suggestion @corykon would be to take the bull by the horns here and create a fork of Cashmere for yourself, then make a branch for updates to the colors.scss file.  And we can talk it through as a group if you have a draft for the OpenSpace meeting.  Or submit a pull request and we can talk it through there.

src/app/lib/button/button-demo/button-demo.component.html
andrew-frueh: Completely agree @benjanderson - I'll get that added today and double check the styling on that anchor component.  Then hopefully this pull request will be ready to be merged.

andrew-frueh:
@benjanderson @corytak @jacquie @corykon - I made one more commit on this pull request to incorporate the anchor link into button demo and cleaned up its styling.  As Cory pointed out, we should probably get this pull committed soon so it doesn't start resulting in collisions.  So throw an approval on if you're good for me to merge this (and I think I'll probably merge & squash on this one because I have dozens of commits here).

Issue #17, opened 2018-01-17
Cashmere Demo Page


benjanderson:
The demo site was based off the site @andrew-frueh example of a [style guide](https://spark.healthcatalyst.com/community/ux/pages/components) on spark. The current version loosely resembles that example. Our demo of our components needs improvement to bring it up to modern style guide standards 

This is a list of things that need to be done, but is not comprehensive:
- [x] landing page (example: https://material.angular.io/ 
- [x] global navigation
- [x] link back to our github repo
- [x] global styles for documentation
- [ ] tabbed code (eg. html / typescipt) component
- [x] component list side bar styling
- [x] version number / alpha release warning

We desperately need help, if you wish to contribute please post your intention here or on #cashmere channel on slack. Please make pull requests small

joeskeen:
Before jumping into tabbed code, I'd like to add code syntax highlighting.  [ngx-highlightjs](https://murhafsousli.github.io/ngx-highlightjs/) looks promising, what do you think @benjanderson?

corytak:
My .02 if it's just for the demo project, I think it would be fine to add a library as a devDependency. If it's going into a component that we're shipping we'll probably need to evaluate libraries more strictly.

joeskeen:
As a n00b to the project, how are we separating out the demo project (cashmere.azurewebsites.net) from library code?  IMHO the demo project should have its own `package.json` so we don't inadvertently add dependencies to the library.

benjanderson:
I am ok with libraries in the demo project as long as they are not included in the main npm package. I am not sure how that works and will defer to @corytak if that is possible

corytak:
The package.json file in the dist folder is generated in build.js
```
shell.cp(`-Rf`, [`package.json`, `LICENSE`, `README.md`], `${NPM_DIR}`);
shell.sed('-i', 'dependencies','peerDependencies','dist/package.json');
```
Basically, it's copying the package.json to the dist folder and replacing `dependencies` with `peerDependencies`. To be included in the npm package/library, the new module needs to be added to src/app/lib/cashmere.ts (the public api). The app.module.ts is the demo app and each component so far has a demo folder as part of its code. 

I'm open to the idea of a separate package.json file for the demo app but am unsure about what it's implementation might look like or the best way to keep the two in sync.

joeskeen:
[...]

PR #24, opened 2018-01-19
Demo page improvement (landing page with some reorganization)

nickrflynn:
It's not pretty yet but it's a step towards something like [https://material.angular.io/](https://material.angular.io/)

(#17 - demo page needs improvement)


benjanderson:
I like the organization and the direction. I think we need some sort of global navigation even if its just the home page and the components. Its a little disorienting to be thrown into the components page without a way back

andrew-frueh:
Cool, good start @nickrflynn.  One small change I would suggest: can you give your <h2> line a font-weight of 300?  So it looks like the attached image.
screen shot 2018-01-19 at 5 09 08 pm

nickrflynn:
I know @benjanderson used https://material.angular.io/ as an example but do we have something else more specific in terms of a vision/mock up for the demo @andrew-frueh?

andrew-frueh:
Nope, nothing more specific than https://material.angular.io/ at the moment.  I'll give some more thought to content for this landing page.  But as far as the browsing experience for the actual components, I think the Angular spec is a great target.  Ultimately, we want this page to replace our style guide PDF.  Let's discuss this more at our next Cashmere Open Space meeting.

andrew-frueh:
Cool - changes look good @nickrflynn.  So @benjanderson - I know we're still working out our governance but for straight forward updates like this can we say if it gets UX signoff and a technical signoff from someone like yourself we can merge?  There is plenty more we'll want to do on the demo page, but I'm assuming the goal isn't to leave pull requests open forever.  Seems like the intent of this one was to introduce a landing page - and thumbs up from me on what Nick has checked in.

benjanderson:
Good point @andrew-frueh, this is a great incremental change. I will approve this knowing its just a base

andrew-frueh:
@benjanderson I added an approval of my own for posterity.  So what is standard practice now - is "Squash and Merge" usually the best way to merge to reduce the list of commits?  Or is that not a big deal?

benjanderson:
I don't have a strong opinion on whether to squash or not. Currently on my team we never squash to keep a better history and merges can be easier. However I can see the benefit of a squash with such a simple change like this. 

I do feel that the person who created the pull request should complete it

andrew-frueh:
Cool, that works for me @benjanderson - makes sense to me to not squash for the sake of preserving history.  And I also like the idea of the person submitting the pull request having the satisfaction of merging it.  So take it away @nickrflynn - go ahead an hit that big shiny merge button :-)

nickrflynn:
I actually can't complete it: `Only those with write access to this repository can merge pull requests.`

benjanderson:
you should have been sent an email or you can go here
https://github.com/HealthCatalyst/Fabric.Cashmere/invitations

PR #29, opened 2018-01-25
Checkbox v1

nickrflynn:
More will need to be done for full support in forms but I believe this is a good v1. 

Note: I also standardized the singular vs. plural component title and route for the button.  I think we all want them singular -- or at least consistent.


src/app/lib/checkbox/checkbox.component.html
```diff
@@ -0,0 +1,7 @@
+<div class="hc-checkbox-container" [ngClass]="{'disabled': disabled}">
+    <input type="checkbox" [attr.id]="name" [disabled]="disabled" [(ngModel)]="model" />
+    <label class="hc-checkbox-overlay" [attr.for]="name" [ngClass]="color"></label>
```
andrew-frueh: Lint caught this one error.  We got rid of the color option but forgot to remove it from the html.

src/app/lib/select/select-demo/select-demo.component.html
```diff
@@ -35,19 +38,26 @@ <h5>Disabled Select Component</h5>
...
       <tr>
-        <td style="font-weight: 600">[options]</td>
+        <th>[options]</th>
         <td>
             <span>Type:</span> Array
         </td>
         <td width=50%>An array of strings for the select options</td>
     </tr>
+    <tr>
+        <th>[disabled]</th>
```
corykon: thanks for going back and updating these 👍

src/app/lib/checkbox/checkbox.component.html
nickrflynn: Running `ng lint` locally didn't catch this.  I'll have to investigate why.

PR #34, opened 2018-01-26
add url rewrite for azure spa

benjanderson:


corykon:
Thanks for knocking this kind of stuff out.

PR #36, opened 2018-01-27
Update SelectComponent to be compatible with Angular Forms

corytak:
Implemented the `ControlValueAccessor` interface to ensure compatibility with angular forms. In order to properly test this, I created a test component that used the select component and used a reactive form. 

I did sneak in some exports to `cashmere.ts`

### DIFF
--- a/src/app/lib/select/select.component.ts
+++ b/src/app/lib/select/select.component.ts
@@ -25,7 +25,7 @@ export class SelectComponent implements ControlValueAccessor {
   onTouched: any = () => { };
 
   get alpha() {
-    return this.disabled ? 0.4 : 1.0;
+    return this.disabled ? this.disabledAlpha : this.enabledAlpha;
   }
 
   get value() {
--- a/src/app/lib/select/select.component.ts
+++ b/src/app/lib/select/select.component.ts
@@ -18,7 +18,7 @@ export class SelectComponent implements ControlValueAccessor {
   @Input() placeholder: string;
   @Input() options: Array<string> = [];
   @Input() disabled: boolean = false;
-  _value: any;
+  _value: string;
   onChange: any = () => { };
   onTouched: any = () => { };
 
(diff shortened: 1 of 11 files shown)


src/app/lib/select/select.component.ts
```diff
@@ -1,22 +1,56 @@
...
+    return this.disabled ? this.disabledAlpha : this.enabledAlpha;
+  }
 
-  constructor() { }
+  get value() {
+    return this._value;
+  }
+
+  set value(val: string) {
+    this._value = val;
```
jacquie: Do we need to check for the existence of onChange and onTouch here or is it guaranteed that they will be set by this point?

src/app/lib/select/select.component.ts
```diff
@@ -1,22 +1,56 @@
...
+  }
+
+  registerOnTouched(fn: any) {
+    this.onTouched = fn;
+  }
 
-  ngOnInit() {
-    if (this.disabled) {
-      this.alpha = 0.4;
+  writeValue(value: string) {
```
benjanderson: Im trying to figure out what this is for?

src/app/lib/select/select.component.ts
corytak: This would get called when the user has selected a new element and the change event is dispatched: `(change)="value = $event.target.value"`. onChange and onTouch are initialized with the component and so I believe it's guaranteed they'll be set by the time the user chooses a new selection.

src/app/lib/select/select.component.ts
corytak: `writeValue` takes a value from a reactive form and sets it in the component. So if we created a formGroup in our host component and patched the values from an ajax call for example, the select element's value would be replaced with the new form value.

corytak:
@andrew-frueh the getting started button issue is resolved

Issue #41, opened 2018-01-27
HC-Icons Font


andrew-frueh:
Develop an HC-Icons font to remove our dependency on FontAwesome, GlyphIcons, or any other 3rd party glyph set.  This document represents the current best understanding of what icons should be included in this font:

https://spark.healthcatalyst.com/docs/DOC-3157

Please look it over and let us know us know if anything is missing that you currently make use of in your apps.

jtrujill:
I'm working on this right now I'll get it in today

andrew-frueh:
That's awesome @jtrujill - thank you!  Let me know need svgs for any of the icons in that Spark doc.

corykon:
Renamed, as it looks like this issue was repurposed for an icon component

PR #47, opened 2018-01-27
Added icon component

jtrujill:
This pull request is for issue #41 to create an icon component


jtrujill:
@corytak I've made the change and I'm now exporting the component and module

PR #48, opened 2018-01-28
Checkbox better support for angular forms and native properties

jtrujill:
Updated checkbox so that it is compatible with angular's reactive and template forms. Added support for more of the input element's native attributes and properties. I didn't mess with any of the component's styles. The alterations to the css is a format change, and I don't know why tslint seems to be letting these differences get through. This is purely a functional update

### DIFF
--- a/src/app/lib/checkbox/checkbox.component.ts
+++ b/src/app/lib/checkbox/checkbox.component.ts
@@ -1,21 +1,158 @@
-/* tslint:disable:component-selector */
+/* tslint:disable:no-use-before-declare */
 
-import { Component, Input, OnInit } from '@angular/core';
+import {
+  Attribute,
+  Component,
+  EventEmitter,
+  forwardRef,
+  HostBinding,
+  HostListener,
+  Input,
+  Output
+} from '@angular/core';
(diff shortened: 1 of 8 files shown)


src/app/lib/checkbox/checkbox.component.ts
```diff
@@ -1,21 +1,158 @@
...
+  stopChangeEvent(event: Event) {
+    event.stopPropagation(); // prevent native change event from emitting its own object through output 'change'
+  }
+
+  private emitChangeEvent(): void {
+    this.onChangeFunc(this.checked);
+    this.change.emit(new CheckboxChangeEvent(this, this.checked));
+  }
+
+  private anyToBoolean(value: any): boolean {
```
nickrflynn: Much better than my `<any>this.disabled === 'true'`

joeskeen: I could see this being useful outside of this class as well; would it be good to extract this out as some sort of utility method?

jtrujill: Exactly what I was thinking. I was going to do that in a different commit. I don't know what the best name would be for a folder and how we want to structure these util methods. I could just go with the generic folder name of util or core.

Issue #42, opened 2018-01-27
Filter Drawer


andrew-frueh:
For apps that make use of robust filtering, a left side filter drawer is used to expose what filters are available allow users to make selections.  Within the drawer users can save Filter Sets to easily flip between previously saved sets of filter selections.  Clicking on a filter row typically triggers a modal with a picklist where specific selections can be made.

The filter drawer integrates with filter chiclets that are visible in header row of the page content next to the button used to open/close the filter drawer.  The chiclets tell the user what filters are currently active and allow the user to clear specific filters without opening the drawer.  

A clickable prototype to demonstrate the functionality described above is available here:
https://xd.adobe.com/view/5260b164-4d93-4238-8fef-8bde1575f248/screen/9180b57f-a454-46eb-bf46-fe1bbc8a005a/Dashboard-Filters

![filterdrawer](https://user-images.githubusercontent.com/22795893/35475134-4ac49752-0356-11e8-97d7-031f4a047147.png)

jtrujill:
I have a completed component for this that I created for Cafe. Here's what it looks like 

<img width="1001" alt="screen shot 2018-01-27 at 11 57 03 am" src="https://user-images.githubusercontent.com/5596515/35475309-3b57a0f4-0359-11e8-8a79-9d030cc0be3e.png">

I'll get this converted by the end of the weekend or Monday

andrew-frueh:
You're a rockstar @jtrujill - the CAFE (or should I say "Touchstone") filters drawer looks perfect.  I think CAFE was actually the first place we came up with that concept, but it's being used all over now.  The only difference is the "Filter Sets" thing that we added in later, and I think Leading Wisely may have added that.  Am I remembering that right @jacquie?  In any case, go ahead and do a PR for the version exactly like you have it in CAFE, and we can worry about adding the Filter Sets later.

jtrujill:
PR #49 is now merged

PR #49, opened 2018-01-28
Added drawer component

jtrujill:
Closes issue #42


jtrujill:
@corytak I've exported the drawer component and module

Issue #53, opened 2018-01-29
Dropdown/Select component arrow toggle


jonest:
Looking at the style guide v1.3 page 42, the dropbown/select component should have the arrow toggle depending on the open or closed state of the menu:

![image](https://user-images.githubusercontent.com/3075414/35530212-a31ecbde-04f0-11e8-8c01-18dd63cbac88.png)

jonest:
I don't have permissions to assign this to myself, but I've got a fix for this

andrew-frueh:
I'm going to close this one for now while our select component uses the native DOM select.  If we get to a point where we create our own dropdown (as in the image above), then toggling the arrow is definitely something we should do.

PR #54, opened 2018-01-29
Toggle dropdown arrow

jonest:
This resolves issue #53

dropdown

### DIFF
--- a/lib/src/select/select.component.html
+++ b/lib/src/select/select.component.html
@@ -1,6 +1,6 @@
 <label (click)="toggleArrow($event.target)" (focusout)="resetArrow($event.target)" class="closed" [style.opacity]="alpha">
     <select [disabled]="disabled" (change)="value = $event.target.value" required>
-      <option *ngIf="placeholder" [value]="null" selected disabled hidden>{{placeholder}}</option>
+      <option *ngIf="placeholder" [value]="null" selected disabled>{{placeholder}}</option>
       <option *ngFor="let option of options" [value]="option" [selected]="option === value">{{option}}</option>
     </select>
 </label>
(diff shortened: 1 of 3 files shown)


jonest:
@andrew-frueh Good find! I see it on Firefox on Windows as well.

jonest:
After looking into it, it's specifically related to this bit of CSS:

```css
&.closed::after {
    content: "\f078";
}
&.open::after {
    content: "\f077";
}
```

If I remove that, it won't toggle, but it won't close immediately. Something about changing this content causes it to close. I tried some various other things, but I'm stumped. I'm not sure what to do to get it to work in Firefox :(

jonest:
@andrew-frueh It looks like it has something to do with the `hidden` attribute. If I remove that it toggles just fine. How important is it to keep that in there for the placeholder?

image

corytak:
@jonest FYI I'm still seeing the error in Firefox where the dropdown closes immediately after opening.

andrew-frueh:
Thanks for persisting on this one @jonest - the good news is, that on my laptop it looks like your change did fix the issue with the dropdown not staying open.  Unfortunately, what I'm still seeing on Chrome and Safari is that after you close the dropdown or make a selection, the arrow stays upside down until I click away from the select box (remove focus from it).  I think it has something to do with what fires the (focusout) command.  Like Chrome keeps focus on the select item even after you make a selection.

I would say mess around with other events to catch that, but I was just talking with @corytak and he was seeing some other strange behavior on his computer with the select box still not staying open.  I'll let Cory elaborate on what he was seeing.

Bottom line, I hate to see you spinning your wheels on this one when there is other stuff you could be picking off that are higher priority (like say #78 breadcrumbs, etc).  It seems like the click handlers kind of open a can of worms across browsers.  So maybe just better to put this one on the shelf and come back to it later?

jonest:
> So maybe just better to put this one on the shelf and come back to it later?

It's no big deal! I haven't really been putting too much effort into it :P

> @jonest FYI I'm still seeing the error in Firefox where the dropdown closes immediately after opening.

@corytak Are you on Windows? I was on a Mac last night looking at this.


I'll take a peek at the issues at some point.

corytak:
@jonest yes I'm on windows.

PR #59, opened 2018-01-30
Incorporate <hc-icon> into Button Component

andrew-frueh:
Want to make sure and take advantage of the fancy new hc-icon component from @jtrujill.  So I've adjusted the button component to use it for the optional glyph and dropdown indicator.


andrew-frueh:
Well @jtrujill it looks like we were working on the same thing at the same time.  I think the only thing that I did differently than you was add a class to handle the vertical alignment and padding when an hc-icon is added.  Without it the glyph doesn't end up centered.

vertical

jtrujill:
@andrew-frueh awesome it looks good. You actually don't need that deep selector for the hc-icon so you can remove that

andrew-frueh:
Cool, thanks @jtrujill.  I actually tried doing that class with /deep/ first, but it wasn't getting applied to hc-icon.  I just checked it again to confirm and that seems to be the case.  I'm assuming the reason I need it is because hc-icon lives within <ng-content>?  But I'm far from an angular expert, so let me know if there's better way to do that.

jtrujill:
Okay if you get that merged in I can take a look and see what's up at a later point

PR #64, opened 2018-01-30
Genericize tabs component to include horizontal tabs

corytak:


src/app/lib/tabs/tab-set.component.scss
```diff
@@ -0,0 +1,62 @@
...
+  .tab-horizontal {
+    background-color: inherit;
+    color: $offblack;
+    font-size: 14px;
+    padding: 10px 20px;
+    &:hover{
+      cursor: pointer;
+    }
+    &.active {
+      border-bottom: $blue 3px solid;
```
andrew-frueh: I should have caught this before @corytak - but can you make the horizontal bar 4px thick rather than 3?  It just looks a little thin now.  Other than that, I'm think this PR looks great.

corytak:
@andrew-frueh I updated it to be 4px on the horizontal active tab

andrew-frueh:
Thanks @corytak - awesome work on this!

PR #65, opened 2018-01-30
Add Navbar, Popover, and App Switcher components

benjanderson:
This took much longer than I intended. The documentation still needs improving and it could use more tests, but I thought I would get it out sooner so I can get feedback


benjanderson:
notice all the `merge master`, it was out a little too long

corytak:
@benjanderson when I ran the `node build.js` to bundle the library strict null checking caught some aot compilation errors: 
image

.angular-cli.json
```diff
@@ -23,6 +23,7 @@
         "styles.scss"
       ],
       "scripts": [
+        "../node_modules/popper.js/dist/umd/popper.min.js"
```
corytak: Not sure if popper is used directly in the code yet but it may need to be added to the globals object in `rollup.config.js`

src/app/lib/navbar/navbar.component.scss
```diff
@@ -0,0 +1,119 @@
...
+
+$navbar-color: $charcoal-blue !default;
+$navbar-brand: $blue !default;
+$navbar-height: 53px !default;
+$navbar-app-padding: 0 75px 0 25px !default;
+$navbar-text: $white !default;
+$navbar-text-inactive: $gray-300 !default;
+$navbar-z-index: 1030 !default;
+
+.navbar.fixed-top::ng-deep {
```
corytak: It's my understanding that `ng-deep` [is deprecated. ](https://angular.io/guide/component-styles#deprecated-deep--and-ng-deep). However, I'm not sure of a different way to handle this.

.angular-cli.json
andrew-frueh: Yup, I'm getting the same error on popper.js as well.  Not able to compile and test without it.

src/styles.scss
```diff
@@ -4,12 +4,15 @@ $fa-font-path : '../node_modules/font-awesome/fonts';
 html,
 body {
     font-family: 'Open Sans', 'Helvetica Neue', Helvetica, Arial, sans-serif;
-    max-width: 1540px;
-    margin: 0 auto;
-    height: 100%;
+    margin: 0;
```
corytak: I know a handful of apps internally have max widths and gutters on the side. I'm not sure if we need to replicate that functionality in cashmere's docs or not.

benjanderson:
Thanks @corytak I think it may be a good idea to add `node build.js` step to the travis ci check

corytak:
Good idea @benjanderson . I'll create an issue.

jonest:
I also had problems with getting `popper.js` installed, too, @corytak . Looks like that was also an issue for you from your screenshot. I had to change my `.npmrc` to get it to work. Perhaps we should include one in here so that it pulls from the npm repository? I think it was trying to pull from our private repository.

jonest:
@benjanderson I see that the text for the app name shrinks as you change size, but not the tabs. Is this desired?

image

jonest:
I'm guessing for a second pass, we'd want to have the tabs collapse, correct?

andrew-frueh:
Correct @jonest - long term it would be great to get the navBar responsive so that it collapses to something like this:

responsivemenu

src/app/lib/navbar/navbar-demo/navbar-demo.component.ts
```diff
@@ -0,0 +1,49 @@
...
+        <a href="https://community.healthcatalyst.com/" target="_blank">Health Catalyst Community</a>
+    </li>
+    <li>
+        <button (click)="aboutClick($event)">About</button>
+    </li>
+  </ul>
+</hc-popover-content>`;
+
+  fixedTopNavbar: string = `body {
+  padding-top: 53px;
```
jonest: Is there a way this can be done automatically instead of having a user adjust their padding? Can we target the body ourselves from this component?

jonest:
Great job on this by the way @benjanderson ! 🙌

src/app/lib/navbar/navbar-demo/navbar-demo.component.ts
benjanderson: changing the body styles is not the cleanest thing to do for a component. I am open to suggestions though

benjanderson:
@jonest correct, for v1 I was just supporting the minimum width of 1024, but I can see us adding responsive behavior pretty cleanly in the future

jonest:
> @jonest correct, for v1 I was just supporting the minimum width of 1024, but I can see us adding responsive behavior pretty cleanly in the future

Completely understandable :) Just wanted to keep that in mind as I was taking a peek at things from a QA perspective :)

src/app/lib/app-switcher/app-switcher.component.scss
```diff
@@ -0,0 +1,50 @@
+@import '../colors.scss';
```
corykon: App switcher look really good 👍

jonest:
> Awesome stuff @benjanderson - only thing I have is on the rollover states for items in the navBar. We haven't given clear direction on this in the past btw. My personal feeling is that the rollover and click states for the navBar items should just be the dark blue background color change (no blue underline, no bold text). And only when the item switches to its active state do we add the underline and bold. It just separates rollover from active better in my opinion.

Don't know about you guys, but I'd be totally fine with these things coming through in a second pass :) We could have more than one person working on them that way, too!

src/app/lib/navbar/navbar-demo/navbar-demo.component.ts
jonest: Sounds reasonable 👍

andrew-frueh:
Totally true @jonest.  @benjanderson I'd be happy to make that rollover change myself and submit as a pull request after this one is checked it.  There is a lot of code here to get merged, so maybe Ty is right.  Up to you Ben!  I'll put my approval in either way.

benjanderson:
thanks to @corykon, your navbar rollover changes are done :)

andrew-frueh:
@benjanderson @corykon - you guys are amazing; this is a huge leap forward for Cashmere.  Very exciting!

AdamGMiller:
The app switcher is pretty awesome.  My one worry is that I think the discovery service is secured by a windows login.  In that case, if someone is authorized to an application but hasn't been added as a Discovery Service User in the EDW, they'll be unauthorized to the discovery service.  For Measures Manager, we've created an endpoint in our application so that we can use our service account to hit the discovery service and work around that issue.  

Not sure what the best solution is here - we could have the component simply take an array of applications, we could open up the discovery service endpoint to everyone, or we could allow a full custom discovery service url so we can pass application-specific endpoints.

PR #71, opened 2018-01-31
Button cleanup

jtrujill:
Added a util file, used hostbinding where possible, updated button class names and added a unit test for a missing case

### DIFF
--- a/src/app/lib/button/anchor.component.ts
+++ b/src/app/lib/button/anchor.component.ts
@@ -1,25 +1,34 @@
 /* tslint:disable:component-selector */
-/* tslint:disable:use-host-property-decorator */
 // https://github.com/mgechev/codelyzer/issues/178#issuecomment-265154480
 
-import { ChangeDetectionStrategy, Component, ElementRef, HostListener, Renderer2 } from '@angular/core';
+import {
+  ChangeDetectionStrategy,
+  Component,
+  ElementRef,
+  HostBinding,
+  HostListener,
+  Renderer2,
+  ViewEncapsulation
+} from '@angular/core';
 import { ButtonComponent } from './button.component';
 
 @Component({
   selector: 'a[hc-button]',
   template: '<ng-content></ng-content>',
   styleUrls: ['./button.component.scss'],
-  host: {
-    '[attr.tabindex]': 'disabled ? -1 : 0',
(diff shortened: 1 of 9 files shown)


src/app/lib/button/anchor.component.ts
```diff
@@ -1,25 +1,34 @@
...
   template: '<ng-content></ng-content>',
   styleUrls: ['./button.component.scss'],
-  host: {
-    '[attr.tabindex]': 'disabled ? -1 : 0',
-    '[attr.disabled]': 'disabled || null',
-    '[attr.aria-disabled]': 'disabled.toString()'
-  },
-  changeDetection: ChangeDetectionStrategy.OnPush
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  encapsulation: ViewEncapsulation.None
```
corykon: Do we really need to turn off view encapsulation? Would be great to avoid this as much as possible.

jtrujill: The reason for turning that off is so that people can style it differently if needed. Maybe they want to increase the size of a button icon, align the text differently, change the color to something not supported. 

There is no issue with turning view encapsulation off for a library component and the reason is that when the user of a library uses a component within their own component which should have ViewEncapsulation.Emulated any changes that they make to the component will not leak out of that component.

What we are doing by having the color property is giving the user access to preset css values that match our company's design doc, but there will be situations where the user will need to change the component slightly.

I will make an issue to discuss this

jtrujill: Also I agree that this spacing issue is beginning to get irritating because the project is currently set at 2 spaces. If people don't like that we need to discuss as a team what we want for formatting and stick to that. I'll bring this up in the meeting tomorrow

andrew-frueh: It looks like Charles Waldner's poll on slack showed 4 spaces as the clear winner on indent.  And I think there was a strong sentiment that we just need to pick something and go with it.  I don't think anyone was lobbying hard for one option over another.  So I can check with Charles it that's something he was planning on submitting a PR for.  Or if anyone else wants to that's fine too.

jtrujill: I agree I don't care I just want consistency because it makes merging difficult

PR #73, opened 2018-01-31
Table Styles

corykon:
Added some base css classes to use on `<table>` elements. Could use your eyes on this @andrew-frueh. 

For now, this is just styles, but I'd like to chat more with people about componentizing things at open space. Lots of people have built table components, and I think there are some good learnings to gather.


andrew-frueh:
Wow.  For a visual perspective this is exceptional @corykon - very impressed.

How would your styles currently handle include a column of our hc-checkbox components (which would in turn highlight the row)?  Also how would your styles handle a column of hc-icon components (similar to [something like this](https://xd.adobe.com/view/f0c739fa-f83c-4cf5-ad04-d8f39a419fe7/) in Measures Manager)?

src/styles.scss
```diff
@@ -1,10 +1,16 @@
...
+@import './app/lib/colors.scss';
+@import './app/lib/tables.scss';
+
 
 html,
 body {
     font-family: 'Open Sans', 'Helvetica Neue', Helvetica, Arial, sans-serif;
-    margin: 0;    
+    margin: 0;
+    height: 100%;
```
benjanderson: this height 100% causes there to always be a scroll bar 

image

src/app/lib/tables.scss
```diff
@@ -0,0 +1,144 @@
+@import "./colors.scss";
```
benjanderson: I would recommend pulling some of the customizable parts into variables with !default so they can be customized if so desired https://robots.thoughtbot.com/sass-default.

src/app/lib/tables.scss
```diff
@@ -0,0 +1,144 @@
...
+//-----------------------------------------
+
+table.hc-table {
+    position: relative;
+    width: 100%;
+    max-width: 100%;
+    color: $text;
+
+    td, th {
+        padding: 8px 16px;
```
benjanderson: $cell-padding

src/app/lib/tables.scss
```diff
@@ -0,0 +1,144 @@
...
+            &:nth-child(2n) {
+                background-color: $white;
+            }
+
+            &:nth-child(2n + 1) {
+                background-color: $slate-gray-100;
+            }
+
+            &:not(.empty-table-row):hover {
+                background-color: tint($primary-brand, 75%);
```
benjanderson: $hover-color

src/app/lib/tables.scss
```diff
@@ -0,0 +1,144 @@
...
+        border-left: 1px solid $slate-gray-200;
+        border-right: 1px solid $slate-gray-200;
+
+        tr {
+            &:nth-child(2n) {
+                background-color: $white;
+            }
+
+            &:nth-child(2n + 1) {
+                background-color: $slate-gray-100;
```
benjanderson: $row-color-alternate

src/app/lib/tables.scss
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    tbody {
+        font-size: 15px;
+        border-left: 1px solid $slate-gray-200;
+        border-right: 1px solid $slate-gray-200;
+
+        tr {
+            &:nth-child(2n) {
+                background-color: $white;
```
benjanderson: $row-color

src/app/lib/tables.scss
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    // Sorting
+    th.hc-col-sortable, th.hc-col-sortable-left {
+        position: relative;
+        cursor: pointer;
+        padding-right: 35px;
+
+        &:hover {
+            background-color: tint($primary-brand, 75%);
```
benjanderson: reuse

src/app/lib/tables.scss
benjanderson: also aliasing the font-awesome characters with variables would make it more readable

src/app/lib/tables.scss
```diff
@@ -0,0 +1,144 @@
...
+
+    td, th {
+        &.hc-cell-borders {
+            border: 1px solid $slate-gray-200;
+        }
+    }
+
+    // Condensed
+    &.hc-table-small {
+        td, th { padding: 5px 16px; font-size: 13px;}
```
benjanderson: $cell-padding-condensed

corykon:
> How would your styles currently handle include a column of our hc-checkbox components (which would in turn highlight the row)? Also how would your styles handle a column of hc-icon components (similar to something like this in Measures Manager)?

Shouldn't have any problems implementing either of those cases on top of these styles!

src/app/lib/tables.scss
corykon: Great suggestions, I'll work on that

PR #75, opened 2018-01-31
Changed project indent size to 4 spaces

jtrujill:


jtrujill:
Ha, yeah. That wasn't the point but ka-ching

PR #80, opened 2018-02-01
Add Github link to navbar

nickrflynn:


nickrflynn:
Small piece of #17 (Cashmere Demo Page)

benjanderson:
Its a tragedy we couldn't use the navbar-icon element to do this. When you check this in I may try to modify the navbar to work with this

jtrujill:
@benjanderson What does the navbar-icon do? Could we just modify the navbar so that it styles the hc-icon the same as the navbar-icon

nickrflynn:
Done @andrew-frueh

PR #85, opened 2018-02-02
NavBar App Switcher Hover/On State

andrew-frueh:
Minor adjustment to @benjanderson's amazing navBar.  Adds a hover state for the app switcher button to darken the background and keeps the dark background on until the popover is closed.  Thanks for your help with this one Ben!

### DIFF
--- a/src/app/lib/app-switcher/app-switcher.component.scss
+++ b/src/app/lib/app-switcher/app-switcher.component.scss
@@ -7,12 +7,18 @@
 
 .brand {
     cursor: pointer;
+    transition: background-color 0.25s ease;
+}
+
+.brandOn {
+    background-color: $dark-blue;
 }
 
 .popover-content {
     max-width: 410px;
     & > h4 {
         margin: 10px 0;
+        padding-bottom: 5px;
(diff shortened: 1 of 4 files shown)


src/app/lib/app-switcher/app-switcher.component.ts
```diff
@@ -16,6 +16,7 @@ export class AppSwitcherComponent implements OnInit, OnDestroy {
     public subscription: Subscription;
 
     private ngUnsubscribe: any = new Subject();
+    private brandBg = 'brand';
```
benjanderson: you need to make this public to pass the travis ci check. private properties cannot be accessed by the markup 

`app-switcher.component.html(5,6): : Property 'brandBg' is private and only accessible within class 'AppSwitcherComponent'`

src/app/lib/app-switcher/app-switcher.component.scss
```diff
@@ -7,20 +7,25 @@
 
 .brand {
     cursor: pointer;
+    transition: background-color 0.25s ease;
```
benjanderson: should we put more animations into our hover states?

PR #87, opened 2018-02-02
Separate documentation from library code

### LINKED ISSUE #20: Separate documentation site from library code
As per discussion on #17, separating out the documentation site from the library code has advantages including, but not limited to:

* dependencies can be added to the documentation site without bloating the library
* it's easier to understand when making changes what impact those changes will have
* more approachable for new contributers

Making changes to this will require care to preserve the NPM packaging process and the documentation site publishing process.  Changes to these may be necessary.

joeskeen:
Fixes #20 

With these changes, the Angular CLI webapp code is now found in /docs, which includes the reference implementation of the library, also known as the documentation.  Meanwhile, the library code has been separated out into the /lib/src folder.  This makes it easier for you to make changes to the demo or library code without inadvertently impacting the other.  Additionally, this makes it possible to separate library dependencies from demo app dependencies, as each has their own package.json file.


debug.log
```diff
@@ -0,0 +1 @@
+[0202/145128.544:ERROR:crash_report_database_win.cc(568)] CreateDirectory C:\Users\JOE~1.SKE\AppData\Local\Temp\karma-96591541\Crashpad: The system cannot find the path specified. (0x3)
```
benjanderson: I dont think this file should be here

docs/cashmere-import.scss
```diff
@@ -0,0 +1 @@
+@import "../lib/src/colors.scss";
```
benjanderson: It took me a while to realize that this file is just making it easier for other files to import. I like it, but I think a comment would be needed to explain the single import

lib/build.js
```diff
@@ -0,0 +1,85 @@
+// Additional files to be included in the NPM package
+const includeFiles = [
```
benjanderson: when this is merged, could you make an issue that we can add to the read.me that popper.js will need to be added manually to the project

lib/src/button/button.module.ts
```diff
@@ -1,19 +1,19 @@
 import { ButtonComponent } from './button.component';
 import { NgModule } from '@angular/core';
 import { CommonModule } from '@angular/common';
-import { AnchorComponent } from './anchor.component';
+// import { AnchorComponent } from './anchor.component';
```
benjanderson: remove comments

debug.log
corytak: Not sure if we want to include this log file.

docs/test.ts
```diff
@@ -26,8 +26,10 @@ getTestBed().initTestEnvironment(
     platformBrowserDynamicTesting()
 );
 // Then we find all the tests.
-const context = require.context('./', true, /\.spec\.ts$/);
+const docsContext = require.context('./', true, /\.spec\.ts$/);
```
corytak: beautifully done.

lib/public_api.ts
```diff
@@ -0,0 +1,11 @@
+export * from './src/app-switcher/app-switcher.module';
```
corytak: Is there any need to also include other exports besides modules? The scenario I'm thinking of is if somebody uses a cashmere component in their app and needs to declare the cashmere component in their test bed. Perhaps the module itself makes this available but I wasn't sure.

lib/src/button/button.component.spec.ts
```diff
@@ -30,8 +31,8 @@ class TestAppReference {
     constructor() {
         this.fixture = TestBed.createComponent(TestAppComponent);
         this.testAppComponent = this.fixture.debugElement.componentInstance;
-        this.buttonDebugElement = this.fixture.debugElement.query(By.css('button'));
-        this.aDebugElement = this.fixture.debugElement.query(By.css('a'));
+        this.buttonDebugElement = this.fixture.debugElement.query(By.css('button') as any) as any;
```
corytak: I'm guessing these `as any`s have to do with the no implicit any rule?

lib/src/button/button.module.ts
corytak: We probably want to uncomment these. This might be why those unit tests were failing actually. I'd be curious to see if importing the Button Module works with this uncommented.

lib/tsconfig.json
```diff
@@ -25,8 +25,7 @@
         "inlineSources": true
     },
     "files": [
-        "public_api.ts",
-        "node_modules/zone.js/dist/zone.js.d.ts"
+        "src/cashmere.ts"
```
corytak: Does this file exist? It looks like public_api.ts is in this folder now.

tslint.json
```diff
@@ -97,7 +97,6 @@
                 "variable-declaration": "nospace"
             }
         ],
-        "typeof-compare": true,
```
corytak: Any reason in particular why we don't want to "Makes sure result of typeof is compared to correct string values?"

debug.log
joeskeen: yeah, we should delete this for sure

lib/build.js
joeskeen: TODO

lib/tsconfig.json
joeskeen: yup, I'll fix that

tslint.json
joeskeen: I was seeing a deprecation notice in the console when I was linting, saying that the TypeScript compiler checks this now, so the linting rule is redundant.

lib/src/button/button.module.ts
joeskeen: oh.... right
oops, I commented it out to test things one at a time, I didn't mean to keep it commented out.

Good catch

lib/public_api.ts
joeskeen: so, I was just following the advice from the Packaging Angular talk to only export modules.  I did notice that ng-packagr (via ngc) does export individual components, but that shouldn't be something that consumers of the package should normally have to worry about.

lib/src/button/button.component.spec.ts
joeskeen: This was because I was getting this error when running the tests:

>  02 02 2018 16:39:34.737:ERROR [karma]: Error: lib/src/button/button.component.spec.ts(34,67): > error TS2345: Argument of type 'Predicate<DebugElement>' is not assignable to parameter of type 'Predicate<DebugElement>'.

lib/src/button/button.module.ts
joeskeen: done

debug.log
joeskeen: done

joeskeen: done

joeskeen:
@corykon @corytak @benjanderson 
I just pushed to a public branch as this PR is requiring help from multiple sources.  Rather than add you all as contributors on my fork (as much as I like you all!) I decided to push to a branch of the public repo and re-open the PR there. #88

PR #88, opened 2018-02-03
Separate documentation code from library code

joeskeen:
Continuation of #87 , but from a public branch in the Cashmere repository.  Please feel free to contribute if you find ways to fix the existing issues!


corytak:
I know this is probably in progress, but I wanted to mention some things when I tried to install the tarball into the MeasuresManager application:

- `hc-navbar`  is not a part of the library exports right now
- Getting error with `ng serve` 
>>ERROR in >>../../../node_modules/@healthcatalyst/cashmere/src/popover/popoverContent.component.d.ts(2,20): >>error TS2307: Cannot find module 'popper.js'.
- And this error with `ng serve --aot`
>>ERROR in : 'hc-drawer' is not a known element:
>>1. If 'hc-drawer' is an Angular component, then verify that it is part of this module.
>>2. If 'hc-drawer' is a Web Component then add 'CUSTOM_ELEMENTS_SCHEMA' to the '@NgModule.schemas' of this component to suppress this message. ("[ERROR ->]<hc-drawer></hc-drawer>
>>")
>>../../../node_modules/@healthcatalyst/cashmere/src/popover/popoverContent.component.d.ts(2,20): >>error TS2307: Cannot find module 'popper.js'.
- This is interesting because I'm only importing the `DrawerModule` which is not reliant on popper.
- It is more interesting because the error persists after installing popper into the project. 
- I was able to reproduce the same errors by only importing the TabsModule (again not reliant on popper and popper is installed)

joeskeen:
@corytak I'm wondering if the Typescript build error is because you don't have `"skipLibCheck": true` in your `tsconfig.json`.  I always enable that because you shouldn't be responsible for TS build errors in libraries you install.  The reason you are getting it is you don't have popper installed (because you don't need it).  So this really isn't a true error.

joeskeen:
It looks like I was wrong, the `PopoverContentComponent` has a public member `popperInstance`, which is typed as `Popper`, which comes from the `popper.js` library.  We're going to need to add a dependency on the Popper typings to get this to not have build errors.

PR #91, opened 2018-02-04
Safari NavBar Style Fixes

andrew-frueh:
Fixes a few small bits of random weirdness in Safari (webkit).  Apparently it needs dimensions defined for an img with a svg or it displays super tiny:
tinyappswitchericon

...and on the app switcher, apparently Safari takes font weights more literally than Firefox or Chrome, so the app names were showing up extra bold.  A color also needed to be defined for the show more applications anchor link:

safarifontweights

I went back and checked Firefox/Chrome and everything still looks as expected in those two with these tweaks added.


src/app/lib/app-switcher/app-switcher.component.html
```diff
@@ -5,7 +5,7 @@
      [className]="brandBg"
      (popperOnShown)="buttonOn()"
      (popperOnHidden)="buttonOff()">
-    <img src="./assets/TriFlame.svg">
+    <img height="37px" width="auto" src="./assets/TriFlame.svg">
```
benjanderson: There is already sizes defined in the sass file, I would prefer you change them here instead of inline styles that overwrite them.

image

src/app/lib/navbar/navbar.component.html
```diff
@@ -1,7 +1,7 @@
 <nav class="navbar" [ngClass]="{'fixed-top': fixedTop}">
     <div #defaultbrand class="navbar-brand" *ngIf="appswitcherbrand.children.length == 0">
         <a [routerLink]="homeUri" class="brand">
-            <img src="./assets/TriFlame.svg">
+            <img height="37px" width="auto" src="./assets/TriFlame.svg">
```
benjanderson: that sass change will make it so you dont have to duplicate this inline style...

andrew-frueh: Ah, got it.  Perfect, will do.

lib/src/navbar/navbar.component.scss
```diff
@@ -42,8 +42,8 @@ $navbar-fixed-shadow: 0px 2px 6px $shadow;
         }
         .brand {
             & > img {
-                max-width: 70%;
-                max-height: 70%;
+                height: 37px;
+                width: auto;
```
andrew-frueh: Good call on making this a change to the style rather than an inline fix @benjanderson.  This fix look okay?  I think a fixed height value should be fine for desktop.  And if we need to change the size of the header for mobile, we'll do that with media queries anyway.

PR #93, opened 2018-02-05
Fully Responsive Home Page

andrew-frueh:
Updates to the styling of the demo home page content to make it fully responsive - phone to desktop.  Tested on Chrome, Firefox, and Safari.


src/app/home/home.component.html
```diff
@@ -3,36 +3,36 @@
...
     </span>
 </header>
 <main>
     <section>
-        <span><img src="../../assets/CashmereHome_UX.png" height="370px" width="auto" alt="Cashmere UX Illustration"/></span>
-        <span class="sectionCenter"  style="height: 370px;">
-            <div style="padding-left: 100px;">
+        <span class="sectionOneImg"><img src="../../assets/CashmereHome_UX.png" alt="Cashmere UX Illustration"/></span>
+        <span class="sectionCenter"  style="max-height: 370px;">
+            <div class="sectionOneText">
```
corytak: Thanks for taking out these inline styles

andrew-frueh: Thanks Cory, and agreed on the NavBar, I was thinking the same thing.  Also it will be a good enticement for developers to use our navbar over their's if it offers things like built in responsiveness.  I'll add an issue for this and I volunteer myself to get it started.

src/app/home/home.component.html
```diff
@@ -3,36 +3,36 @@
         <img src="../../assets/CashmereLogo.svg" height="190px" width="auto" alt="Fabric.Cashmere Logo"/>
     </span>
     <span class="headerRight">
-        <img src="../../assets/FabricCashmere.svg" height="55px" width="auto" alt="Fabric.Cashmere Title"/>
+        <img src="../../assets/FabricCashmere.svg" height="100%" width="auto" alt="Fabric.Cashmere Title"/>
```
corykon: Looks like most inline styles were ripped out :+1: I'd recommend throwing a class or id on these last few images and divs and pulling out the height/width stuff, too

andrew-frueh: Yep, had to take them out so I could override them in the media queries.  Good suggestion on moving the height/width on all the imgs to classes.  I'll get an updated commit pushed this afternoon.

src/app/home/home.component.html
```diff
@@ -3,36 +3,36 @@
...
         <p class="headerCopy">Health Catalyst's living style guide for Angular</p>
         <button hc-button title="Primary Button" color="primary" routerLink="/demo">Get Started</button>
```
corykon: Super minor, but while you're in there... could we change the title attribute on the buttons on this page to not be "Primary Button"? Was checking out the homepage and noticed the browser tooltips

andrew-frueh: Ah that's a great catch!  I completely missed that.  Will do.

src/app/home/home.component.html
benjanderson: I would love if something like this would be a global style, as responsive images are common.  I am working with jaque on this stuff so I may do that

src/app/home/home.component.html
```diff
@@ -42,7 +42,7 @@ <h2>Grassroots Spirit</h2>
 </main>
 <footer>
     <span class="footerLogo">
-        <img src="../../assets/CashmereLogoWhite.svg" height="40px" width="auto" style="fill: #ffffff;" alt="Fabric.Cashmere Logo"/>
+        <img src="../../assets/CashmereLogoWhite.svg" height="40px" width="auto" alt="Fabric.Cashmere Logo"/>
     </span>
     <span class="footerVersion">
         <p>Current Version: <b>0.1 (pre-release)</b></p>
```
benjanderson: according to npmjs.org the current version is `0.0.2` https://www.npmjs.com/package/@healthcatalyst/cashmere

src/app/home/home.component.html
andrew-frueh: That's an excellent idea.  I know Bootstrap has a responsive image class, and that would be awesome if we could do something similar.  I found that Firefox and Chrome handled % img dimensions a bit differently, so the class might take a bit of tweaking.  But to your point, that's probably outside of the scope of this PR - maybe open an issue for it if you get a minute and assign yourself, me and Jacquie for starters.

src/app/home/home.component.html
andrew-frueh: Hah, okay I just made that number up, so that's great to know.

andrew-frueh:
@corykon - all inline styles removed and button titles updated.  Let me know if everything looks good to you.

PR #95, opened 2018-02-05
Added split-button

### LINKED ISSUE #60: Button Dropdowns
Split Button Dropdowns have a primary action, and then a series of secondary options that can be surfaced by clicking the split arrow.  Below is an example from @corykon in Population Builder.

This button component should closely mirror the main button component, and perhaps we should remove the dropdown option from the main button component?  Then we could call this a "button-dropdown" or "button-select" component and include an option for split button or not.

buttongroup

jtrujill:
Closes #60

### DIFF
--- a/docs/app/demo/button/button-demo.component.html
+++ b/docs/app/demo/button/button-demo.component.html
@@ -88,15 +88,22 @@
     <h3>Angular Component</h3>
     <hc-tab-set direction="horizontal">
         <hc-tab title="HTML">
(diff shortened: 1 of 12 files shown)


jtrujill:
Okay I got the newest updates from master merged in so it should be good to go

jtrujill:
@andrew-frueh I've updated the ui so it has a divider and I've included a some documentation for using the split button

PR #100, opened 2018-02-06
hc-navbar-icon using hc-icon

andrew-frueh:
I got started on the responsive navbar, but realized I needed to get the following corrected first, so I figured it called for it's own pull request:

- Instead of using `<i>` tags for FontAwesome icons in hc-navbar-icon, I adjusted it to use @jtrujill's hc-icon component instead
- Adjusted our demo site's navbar to use a hc-navbar-icon for the GitHub link as @benjanderson suggested
- Added a hover state to hc-navbar-icons
- Updated the hc-navbar-icons on the navbar demo page
- Although it's more of a fringe case in our apps, I made sure that hc-navbar-icons can be wrapped in `<a>` tags if the icon should act as a link off the app (usually the icons will trigger a popover or something like that).  But at least this allows us to do it on the demo page


andrew-frueh:
Alrighty @benjanderson - got that vertical-separator highlight resolved so I think this PR is good to go

PR #102, opened 2018-02-06
Tabs Working With Angular Router

### LINKED ISSUE #76: Enhance tabs to allow deep links/set active tab
User feedback with the tabs module: tried to see if they could get it to work in their app, but need the ability to set the active tab based on the route. Unsure of best implementation/route, but wanted to log the issue as an enhancement.

corytak:
fixes #76 

Does not update docs to use the tabs. Discussed with @andrew-frueh @corykon and @benjanderson about potentially enhancing the interface at a later time to make this possible.


lib/src/tabs/tab-set.component.ts
```diff
@@ -23,16 +38,47 @@ export type TabDirection = 'horizontal' | 'vertical';
...
 
     ngAfterContentInit() {
+        this.defaultToFirstTab();
+        this.checkForRouterUse();
+    }
+
+    private defaultToFirstTab() {
         if (this.tabs.first) {
-            this.tabs.first.active = true;
+            this.setActive(this.tabs.first);
```
benjanderson: Today I learned that QueryList has lots of handy functions

lib/src/tabs/tab-set.component.ts
```diff
@@ -23,16 +38,47 @@ export type TabDirection = 'horizontal' | 'vertical';
...
 
+    private checkForRouterUse() {
+        if (this.tabs.length > 0) {
+            const countUsingRouter = this.tabs.filter(tab => tab.routerLink !== undefined).length;
+            if (countUsingRouter > 0 && countUsingRouter < this.tabs.length) {
+                const tabsMissingRouterLink = this.tabs.filter(tab => tab.routerLink === undefined);
+                throwErrorForMissingRouterLink(tabsMissingRouterLink);
+            }
+
+            if (countUsingRouter === this.tabs.length) {
```
benjanderson: very slick

PR #104, opened 2018-02-07
Sass stuff, open sans, reset, box-sizing: border-box

jacquie:
What fun I have had with this :)
I added open sans as a peer dependency and installed it in demo and then had trouble setting the path to the font files.  Please check over the parts to do with that particularly carefully - I feel like I did some maybe non-standard things to get there.
I added a reset file from [](https://gist.githubusercontent.com/hcatlin/1027867/raw/dc4f0c2ebf8e58026e65edc9fbabe812c3621805/reset.scss)
It looks pretty standard stuff, but it did some weird things to the layout. I have commented it out for now but we can decide what we want to do with it.
I added `*, *::before, *::after {box-sizing: border-box;}` and it was mostly ok. Exceptions were the cashmere logo changed size slightly and the radio button didn't like it at all. That is also commented out  pending a decision on what to do with it.


docs/cashmere-import.scss
```diff
@@ -1,4 +1,2 @@
-//this simplifies importing SASS files from the Cashmere library
-@import "../lib/src/colors.scss";
-@import "../lib/src/functions.scss";
-@import "../lib/src/tables.scss";
+
```
corytak: If we imported the table styles here we might be able to avoid the really long path in table-demo.component.ts. Although I can see why we don't necessarily always want to import everything.

docs/styles.scss
```diff
@@ -1,8 +1,13 @@
...
-@import '../node_modules/font-awesome/scss/font-awesome';
-@import '../lib/src/colors';
-@import '../lib/src/tables.scss';
+@import 'font-awesome';
 
+$FontPathOpenSans: "../node_modules/npm-font-open-sans/fonts";
+@import 'open-sans.scss';
+
+@import './cashmere-import';
+// *, *::before, *::after {
```
corytak: It might require some changes to the doc styles, but uncommenting the _reset could be a good way to ensure that the components appear the same way across all browsers.

.angular-cli.json
```diff
@@ -22,6 +23,13 @@
             "styles": [
                 "styles.scss"
             ],
+            "stylePreprocessorOptions": {
+                "includePaths": [
```
benjanderson: I am reading up on this includePaths, and I don't think this is being used properly.
https://github.com/angular/angular-cli/wiki/stories-global-styles

If you want these to be included in the bundle, I think these go in the "styles" section right above it. 

I could be wrong, we can discuss it

docs/cashmere-import.scss
jacquie: The problem I was having with tables was that it imports the other common .scss files as well and I couldn't find paths that everything was happy with.  In desperation I decided to separate it out. :(

.angular-cli.json
```diff
@@ -22,6 +23,13 @@
             "styles": [
                 "styles.scss"
```
benjanderson: if you include "sass/cashmere" here I don't think you will have to include this in all the other components

docs/styles.scss
jacquie: I'd spent too long on this already, but yes, if we want this reset in we'll have to go through and make changes to accommodate it. Same with the `box-sizing:border-box`.

lib/src/app-switcher/app-switcher.component.scss
```diff
@@ -1,4 +1,4 @@
-@import '../colors.scss';
+@import '../sass/cashmere';
```
benjanderson: there is going to be quite a bit of style duplication if every component imports all the cashmere styles are imported for every component. We can import variables and mixins because they dont actually create any styles, they are just functional. 

We can discuss it, but I would say we keep colors and give instructions on how to include the cashmere styles (or parts of it) in the main readme

lib/src/sass/tables.scss
```diff
@@ -1,6 +1,8 @@
-@import "./colors.scss";
-@import "./functions.scss";
-@import "../../node_modules/font-awesome/scss/variables"; // get rid of this when we roll our own font
+$fa-font-path: '../../../node_modules/font-awesome/fonts';
```
benjanderson: Not a big deal, but I would call this file _tables.scss

docs/styles.scss
```diff
@@ -1,8 +1,13 @@
 $fa-font-path: '../node_modules/font-awesome/fonts';
-@import '../node_modules/font-awesome/scss/font-awesome';
-@import '../lib/src/colors';
-@import '../lib/src/tables.scss';
+@import 'font-awesome';
 
+$FontPathOpenSans: "../node_modules/npm-font-open-sans/fonts";
+@import 'open-sans.scss';
```
benjanderson: maybe you are using the includePaths correct

lib/src/sass/tables.scss
jacquie: I removed the _ because I thought the _ meant that you wouldn't import it directly, just indirectly via cashmere.scss. I separated tables.scss out as it imports the other common .scss files itself. It is probably worth discussing it though

benjanderson: Im ok with that

PR #106, opened 2018-02-07
Responsive Navbar

### LINKED ISSUE #94: Responsive Navbar
Enhance the existing hc-navbar component to be fully responsive, including phone and tablet resolutions.  Should follow the pattern currently defined in the styleguide:

responsivenavbar

andrew-frueh:
Phew, this one pushed me a bit - but many thanks to @benjanderson for helping me think my approach through.  Resolves #94.  The documentation page has also been updated to explain usage.  For the hamburger menu that drops down on smaller screens, we talked it through and it seemed best to give the developer the flexibility of deciding what they do/don't want in that menu.  For example, they might have hc-icons with popover content that needs to be spelled out in the menu, or they want want to leave items off when being viewed on small screens, etc.  For the content of the responsive menu, I'm making use of @jtrujill's hc-list component, which also gives developers plenty of options on formatting their menu rows.

Two open items for discussion beyond this pull request:
- There are two included media queries for tablet (740px) and phone sizes (420px).  These are hard coded, but for apps with lots of navbar links they might want to flip to the hamburger menu sooner.  So we should discuss if/how we allow the developer to specify their responsive steps.
- @benjanderson it wasn't immediately obvious to me how to disable the tri-flame button when we go to tablet size.  As we discussed in the issue thread, once we go mobile, we only want one menu with everything.  But until we get down to mobile size, I still like the tri-flame as a design element, but we just need to disable it as soon as that 740px media query kicks in.


corytak:
It looks like aot is failing with 
image

You can probably repair this by taking the single quotes off of `'true'`.

corytak:
Actually, better yet, just change the ngIf to be `*ngIf="appSwitcher"`

andrew-frueh:
Whoops, I was so eager to be done with this I forgot to run lint.  I'll get that fixed.  @corytak that's exactly how I was doing the ngIf, but for some reason it wasn't working unless I put the ==true in there.  Is that because I made the input a boolean instead of a string or something odd like that?

lib/src/navbar/navbar-menu/navbar-menu.component.html
```diff
@@ -0,0 +1,8 @@
+<div class="menu-dropdown" [@slideInOut]="menuState">
+    <ng-content></ng-content>
+    <div class="bottomSpacer"></div>
+    <a *ngIf="appSwitcher=='true'" class="app-switcher" href={{appSwitcherService.allApplicationsUri}}>
```
corytak: Switch this to be `*ngIf="appSwitcher"` to avoid the aot issues.

lib/src/navbar/navbar-menu/navbar-menu.component.ts
```diff
@@ -0,0 +1,54 @@
...
+    styleUrls: ['./navbar-menu.component.scss'],
+    animations: [
+        trigger('slideInOut', [
+            state('in', style({
+                transform: 'translate3d(0, -100%, 0)'
+            })),
+            state('out', style({
+                transform: 'translate3d(0, 0, 0)'
+            })),
+            transition('in => out', animate('400ms ease-in-out')),
```
corytak: Look at these fancy angular animations. Bravo. :)

lib/src/navbar/navbar.component.ts
```diff
@@ -18,7 +19,38 @@ export class NavbarComponent implements OnInit {
...
+                this.hamburgerClose = 'inline';
+            }
+        }
+    }
+
+    menuClick( event ) {
+        let clickTarget: string = event.target.classList[0];
+
+        //Verify that the click in the mobile menu came from a navigation item
+        if( clickTarget.indexOf("menu-dropdown") == -1 )
```
corytak: Using a `===` instead of `==` will also check the data type and is usually a good idea.

lib/src/navbar/navbar.component.ts
```diff
@@ -18,7 +19,38 @@ export class NavbarComponent implements OnInit {
...
+                this.mobileMenu.first.show();
+                this.menuOpen = true;
+                this.hamburger = 'none';
+                this.hamburgerClose = 'inline';
+            }
+        }
+    }
+
+    menuClick( event ) {
+        let clickTarget: string = event.target.classList[0];
```
corytak: I suppose we know there will only be one class on the target, but this might be fragile if a new class is ever added. We may want to consider looping over the classes to find the right one?

corytak:
@andrew-frueh yeah the error message seemed to indicate that aot didn't like comparing a boolean and a string type. What wasn't working exactly? I'm running it locally and want to see if that broke anything?

lib/src/navbar/navbar.component.ts
andrew-frueh: I think we're probably safe on this one because there really shouldn't even be more than one mobile menu.  Unless I'm not thinking of a situation where it might be useful to have more than one declared?

lib/src/navbar/navbar.component.ts
```diff
@@ -15,10 +16,40 @@ export class NavbarComponent implements OnInit {
 
     @Input() fixedTop: boolean = false;
 
+    @ContentChildren(NavbarMenuComponent) mobileMenu: QueryList <NavbarMenuComponent>;
+
+    menuOpen: boolean = false;
+    hamburger: string = 'inline';
```
corykon: Not important, but just for your consideration, this might be a cleaner approach to swapping between the hamburger and close icon:

Instead of passing around css values, you could get rid of the `hamburger` and `hamburgerClose` properties entirely, and could have a `mobileMenuIcon` property like this:
```
get mobileMenuIcon(): string { return this.menuOpen ? 'fa-times' : 'fa-bars'; }
```

and then pass that into a single hc-icon in navbar.component.html like this:
```
<hc-icon fontSet="fa" [fontIcon]="mobileMenuIcon" icon-md></hc-icon>
```

andrew-frueh: Great suggestion, very clever - and a good use of @jtrujill's hc-icon.  Although it did reveal one small bug in that component.  In the update function, it was checking whether there was a previous icon, and if so removing it.  But the previousIcon string wasn't ever getting set so the old icon wasn't being removed.  Made that change and all seems to be well.  

Thanks for the tip @corykon - this is definitely a much more elegant way to handle the icon swap.

lib/src/navbar/navbar-menu/navbar-menu.component.ts
```diff
@@ -0,0 +1,54 @@
...
+import { IconModule } from '../../icon/icon.module';
+import { IAppSwitcherService, IDiscoveryApplication } from '../../app-switcher/app-switcher-interfaces';
+import { Subscription } from 'rxjs/Subscription';
+import { Subject } from 'rxjs/Subject';
+
+@Component({
+    selector: 'hc-navbar-menu',
+    templateUrl: './navbar-menu.component.html',
+    styleUrls: ['./navbar-menu.component.scss'],
+    animations: [
```
benjanderson: This uses web animations which has limited support in browsers currently
https://caniuse.com/#feat=web-animation
There are three options:
1) no animations for un supported browsers (but they would still be 100% functional)
2) add shim and instructions on library consumers
3) convert it to use css transitions

benjanderson: It looks great and I am impressed. I just ran into these limitations...

andrew-frueh: My fault @benjanderson - you mentioned that a couple days ago and then I promptly ignored it.  Thanks for the reminder, and fortunately the css transition seems to work just as well as the web animation, and uses less code too.  Interestingly the easing is a little bit different between the two, but it's barely enough to notice.

PR #108, opened 2018-02-09
Subnavbar

### LINKED ISSUE #43: Subheader
The subheader is an optional space, but it very commonly used among Catalyst apps. It is meant to be a flexible area that can be used as per the needs of the application. Some may choose to use it for breadcrumbs, others for secondary navigation, filtering, search, etc.  The most common usage pattern for the subheader is breadcrumbs pinned to the left and one or more buttons pinned to the right.

Because content within the subheader is flexible, the component will need to reflect that.  I'd imagine it would allow the user to define whether the components included in it either belong to the left or right side groups.  But I'd very much appreciate input from @corytak @corykon @jacquie @benjanderson and any others who are making use of the subheader in their apps for their suggestions on configuration options for this component.

subheaderspec

andrew-frueh:
Resolves #43.  @corykon already had a great foundation in place for this one from the styling he did on the demo pages.  See the subnav demo page for an overview of the functionality.  Also included in this PR are minor fixes to the secondary button class - the 2px border was throwing off the positioning of the text.  And when I saw it side by side with a primary button in the subnav, it really stood out.  And last, the hc-subnav component has been swapped into the demo pages headers.

### DIFF
--- a/lib/src/navbar/navbar-menu/navbar-menu.component.scss
+++ b/lib/src/navbar/navbar-menu/navbar-menu.component.scss
@@ -1,11 +1,13 @@
 @import '../../colors.scss';
 
+$menu-z-index: 1025 !default;
+
 .menu-dropdown {
     width: 100%;
     background-color: $slate-blue;
     margin-top: 53px;
     position: fixed;
-    z-index: 1025;
+    z-index: $menu-z-index;
     padding-top: 15px;
     display: none;
     transition: top 0.7s ease;
--- a/lib/src/subnav/subnav-right.directive.ts
+++ b/lib/src/subnav/subnav-right.directive.ts
@@ -4,8 +4,5 @@ import { Directive, HostBinding } from '@angular/core';
     selector: '[hcSubnavRight]'
 })
 export class SubnavRightDirective {
-    @HostBinding('class.subnav-right')
-    get hostClass(): boolean {
-        return true;
-    }
+    @HostBinding('class.subnav-right') true;
 }
--- a/lib/src/subnav/subnav-right.directive.ts
+++ b/lib/src/subnav/subnav-right.directive.ts
@@ -1,0 +1,11 @@
+import { Directive, HostBinding } from '@angular/core';
+
+@Directive({
+    selector: '[hcSubnavRight]'
+})
+export class SubnavRightDirective {
+    @HostBinding('class.subnav-right')
+    get hostClass(): boolean {
+        return true;
+    }
+}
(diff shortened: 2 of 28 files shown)


andrew-frueh:
On second thought (and chatting it through with @corytak), I'm not sure having the ability to include the subnav in `hc-navbar` would ever be all that useful.  I'm thinking 9 times of out 10, pages will want to own the subnav, because otherwise there would need to be a service to communicate with the subnav and that's probably more trouble than it's worth.  Not to mention that not every page will need a subnav.  So I'm gonna yank that bit out of this PR and push a new commit.

lib/src/navbar/navbar-menu/navbar-menu.component.scss
```diff
@@ -1,11 +1,13 @@
 @import '../../colors.scss';
 
+$menu-z-index: 1025 !default;
```
corykon: Z-index variables are wise. :+1: Down the road might be good to have a z-index file that helps us track all our z-index's in one place. This can help squash bugs before they happen.

lib/src/subnav/subnav.component.scss
```diff
@@ -0,0 +1,30 @@
...
+    display: flex;
+    align-items: center;
+    z-index: $subnav-z-index;
+
+    p {
+        margin-bottom: 0;
+    }
+}
+
+.subnav::ng-deep .subnav-right {
```
corykon: ::ng-deep is probably ok for now, but I saw @corytak mention elsewhere that it is depreciated, and we may want to find ways to achieve the styles we want without it

lib/src/subnav/subnav-right.directive.ts
```diff
@@ -0,0 +1,11 @@
+import { Directive, HostBinding } from '@angular/core';
+
+@Directive({
+    selector: '[hcSubnavRight]'
+})
+export class SubnavRightDirective {
+    @HostBinding('class.subnav-right')
```
corykon: I've seen this pattern a few times, and it's totally fine, but just a thought - I think these four lines can be shoved into one. No need to scaffold out a getter function that just returns true.
```
     @HostBinding('class.subnav-right') true;
```

andrew-frueh: Brilliant, I had no idea you could do that.  Change made.

lib/src/navbar/navbar-menu/navbar-menu.component.scss
andrew-frueh: Agreed - I saw @benjanderson doing this on `hc-navbar` and it looked like good practice.  There are now three components in the header that have to maintain their correct z-index, the navbar, subnavbar, and responsive menu.  So that's a great idea down the road to collect those all into one place.

PR #110, opened 2018-02-10
Adds Accordion component

### LINKED ISSUE #99: Accordion component
While on Cafe I designed an accordion component

screen shot 2018-02-05 at 4 07 48 pm

I just need to do some cleanup work and it should be good to go. Below is some example html for how to use it.

```html
<hc-accordion #accordion>
	<hc-accordion-header></hc-accordion-header>
	<hc-accordion-content></hc-accordion-content>
</hc-accordion>
```

One thing that I'm thinking about improving upon is that if a header isn't included then I would just hide the entire header, button included. The user of the component could then get access to the method to toggle it by calling the template variable `accordion.toggle()`. 

Maybe it'd be better to have a boolean like hideHeader and if a user includes an `<hc-accordion-header></hc-accordion-header>` just throw an error. That way it's explicit about hiding the header, because what if the user didn't need to put something in the header but they wanted access to that button

jtrujill:
For now I've left the styling very bare. This component allows you to have an accordion which expands/contracts a section in the DOM. There are properties that allow you to change the location of the trigger button and if you wish you could hide the toolbar at the top completely and toggle the accordion with a button anywhere on the DOM styled as you want.

In the future, if others would find it useful, I will add a parent component which controls sibling accordions so that when one accordion is expanded the others are closed, so you only have one accordion in focus at one time

Closes #99

### DIFF
--- a/lib/src/accordion/accordion.component.scss
+++ b/lib/src/accordion/accordion.component.scss
@@ -1,0 +1,88 @@
+$accordion-border: 1px solid #DDDDDD;
+$accordion-background-color: #F0F0F0;
+$accordion-toolbar-padding: 0 10px;
+
+.hc-accordion {
+    display: block;
+    box-sizing: border-box;
+
+    .hc-toolbar-wrapper {
+        display: flex;
+        flex-direction: row;
+        flex-wrap: nowrap;
+
+        &.hc-align-right {
+            .hc-accordion-trigger {
+                order: 1;
+                margin-left: auto;
+            }
+            .hc-accordion-toolbar {
+                order: 0;
+            }
+        }
+    }
+
+    .hc-accordion-trigger {
+        position: relative;
+
+        &::after {
+            display: block;
+            transform: translate(-50%, 2px) rotate(46deg);
+            border-top: 2px solid;
+            border-left: 2px solid;
+            left: 50%;
+            content: "";
+            width: 8px;
+            height: 8px;
+            position: relative;
+        }
+    }
+
+    .hc-accordion-collapse {
+        display: block;
+        overflow: hidden;
+    }
+
+    &.hc-accordion-closed, &.hc-accordion-closing {
+        .hc-accordion-trigger::after {
+            transform: translate(-50%, -2px) rotate(-134deg);
+        }
+    }
+}
+
+.hc-accordion-toolbar {
+    display: flex;
+    flex-direction: row;
+    flex-wrap: nowrap;
+    white-space: nowrap;
+    box-sizing: border-box;
+    align-items: center;
+    width: 100%;
+    padding: $accordion-toolbar-padding;
+}
+
+.hc-accordion-content {
+    background: white;
+    padding: 12px;
+    //border: $accordion-border;
+    //border-top: none;
+    outline: 0;
+}
+
+.hc-accordion-trigger {
+    height: 50px;
+    width: 50px;
+
+    outline: none;
+    background: transparent;
+    border: none;
+    padding: 0;
+
+    color: deepskyblue;
+
+    &::after {
+        transition: transform 500ms cubic-bezier(0.25, 0.8, 0.25, 1);
+    }
+}
+
+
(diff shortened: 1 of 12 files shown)


andrew-frueh:
Functionality looks great @jtrujill - love the animated transitions.  I'll take a closer look at styling once this PR is merged - I'll have to think about what our standard should be because this isn't an element we had thought to put in previous versions of the style guide.

One request: when we did our last round of user testing on Care Management, we discovered that most users try to click anywhere in an accordion row to expand it.  Eventually the all figured out that they had to click on the arrow specifically, but their first click was usually somewhere random in the row.  So would be possible make the whole row clickable and `cursor: pointer` on the whole row?

@jacquie and @ckeates - I'd love to get your thoughts on this component given [the project tiles that you're currently developing in Leading Wisely](https://xd.adobe.com/view/5260b164-4d93-4238-8fef-8bde1575f248/screen/79a56956-4c9f-4556-aac6-0f21658a590a/PM-Table-Intervention) make heavy use of accordion lists.

jtrujill:
I can implement that change with the clicking anywhere on the  toolbar. If you have a look at #99, I posted an image of what Cafe's accordion looked like. I could easily use that styling or wait for some design work. 

I've designed this component so that it can easily be used in many different ways. I've designed this component so that it's css layout and styling are separate. I want to make it difficult to change the layout while making it easy to change the theme. What I would like to do is provide at least one library provided theme, maybe others, and give the users the ability to use this collapsing component in many areas

andrew-frueh:
That's perfect @jtrujill - I really like that approach.  Maybe we package two "themes" with the component: one could be the CAFE style you mentioned which is great for blocks of content, and maybe the second would be a really minimal style suited for lists of text (similar to that Leading Wisely page I linked to).  Which reminds me, if a developer wants to have an accordion hierarchy, would they accomplish that by nesting `<hc-accordion>` elements?

jtrujill:
Yes they could accomplish that. What I want to achieve with this library is to give developers the building blocks to build complex components. To accomplish this, we need to create components which can be composed together and which allows other components to alter their style to fit the needs of that particular component.

lib/src/accordion/accordion.component.scss
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    .hc-accordion-trigger {
+        position: relative;
+
+        &::after {
+            display: block;
+            transform: translate(-50%, 2px) rotate(46deg);
+            border-top: 2px solid;
+            border-left: 2px solid;
```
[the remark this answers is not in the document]
jacquie: Some accordions I've seen have the arrow  as `>` when it's closed.
I can't exactly explain it, but to me, it seems a long way to go through 180 degrees. It's probably just me :)

lib/src/accordion/accordion.component.ts
```diff
@@ -0,0 +1,149 @@
...
+    @Output() closed = new EventEmitter();
+
+    @Output() closeStart = new EventEmitter();
+
+    get isOpen(): boolean {
+        return this._isOpen;
+    }
+
+    @HostBinding('class.hc-accordion') true;
+
```
jacquie: I'm getting a ts error on this line 
`[ts] Member 'true' implicitly has an 'any' type.`

lib/src/accordion/accordion.component.scss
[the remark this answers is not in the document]
jacquie: All picky stuff from me - looking great otherwise :)

lib/src/accordion/accordion.component.scss
```diff
@@ -0,0 +1,88 @@
+$accordion-border: 1px solid #DDDDDD;
+$accordion-background-color: #F0F0F0;
```
corytak: Should these colors be in the `colors.scss` file for reuse? Conversely is there some way these non-standard colors could be achieved with the `tint` and `shade` functions @corykon wrote?

jtrujill: Unless the colors are used in other components I would say no. No reason to pollute the main color.scss

PR #111, opened 2018-02-10
Breadcrumbs

### LINKED ISSUE #78: Breadcrumbs
Breadcrumbs are a graphical control element used as a navigational aid within our app subheader #43.  It allows users to keep track of their location within the current application.  Our breadcrumbs thus far always appear on the left side of the subheader.  See below for overall styling:

breadcrumbexample

andrew-frueh:
Resolves #78.  Learned a lot about Angular Routers on this one.  Thanks @jtrujill and @petermcclanahan for the suggestions.  Made the component responsive as well - at phone resolution it switches to a back button, or the name of the current page if you're at the root of the nav tree.

@petermcclanahan and @benjanderson - let me know if this component would work in Atlas, or if anything else would be needed for it to work in your app.


jtrujill:
@andrew-frueh good job for a first go through. Overall the component looks good. I really like how you didn't make the css super nested and specific.

andrew-frueh:
Thanks @jtrujill!  I realized I forgot to run lint and test before I submitted again - fixing those issues now...

lib/src/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -0,0 +1,105 @@
...
+
+@Component({
+    selector: 'hc-breadcrumbs',
+    templateUrl: './breadcrumbs.component.html',
+    styleUrls: ['./breadcrumbs.component.scss']
+})
+export class BreadcrumbsComponent implements OnInit, OnDestroy {
+
+    public breadcrumbs: IBreadcrumb[];
+    public routerSubscription: any;
```
corytak: I think the appropriate type for routerSubscription would be `Observable<NavigationEnd>`

lib/src/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -0,0 +1,105 @@
...
+
+            // add breadcrumb
+            let parent: any = this.activatedRoute.parent;
+            let fullURL: string = '';
+            if (parent !== null ) {
+                fullURL += '/' + parent.snapshot.url.map(segment => segment.path).join('/');
+            }
+            fullURL += `${url}`;
+            let breadcrumb: IBreadcrumb = {
+                label: child.snapshot.data[ROUTE_DATA_BREADCRUMB],
```
corytak: This is probably something for a second pass, but often times routes would look like `/things/1/edit`. It seems like the data provided to the id portion of the route would need to be generic with this scenario. It might be nice if the breadcrumb could be `Home Page > Thing 1 Summary > Edit Thing 1` rather than `Home Page > Summary > Edit`

lib/src/breadcrumbs/breadcrumbs.component.ts
andrew-frueh: Thanks for the help getting this one resolved @corytak.  Updated in the most recent commit.

lib/src/breadcrumbs/breadcrumbs.component.ts
andrew-frueh: I like this idea, but I'm probably not the best person to implement it.  Might be better for me to get this first pass merged and let folks who've done more with breadcrumbs in their apps (looking at @petermcclanahan and @benjanderson) build out how we might lookup ids and build them into the breadcrumb label.

benjanderson: This would be a requirement for atlas to use these bread crumbs

andrew-frueh: @benjanderson - as you and I discussed, I'll get this PR merged, and then if you can submit an enhancement issues that would be perfect

PR #113, opened 2018-02-11
Loading page option + some broken sass links

jacquie:
Added Leading Wisely's loading html to index.html as a starting point for discussion on how we want it to look generally for our apps.  
Also there were a couple of broken sass imports for things that were added before we merged the new import strategy


andrew-frueh:
Thanks so much for doing this @jacquie!  Looks great, just wondering if we can tighten up the vertical padding between the elements.  Looks like there is about 50px in between the tri-flame, the logo, and loader right now.  If possible, could we switch to 15px between the tri-flame and the logo; and about 25px between the logo and the loader.  So it would look something like this:

cashmereloader

docs/index.html
```diff
@@ -8,7 +9,97 @@
     <link rel="icon" type="image/x-icon" href="favicon.ico">
     <link href="https://fonts.googleapis.com/css?family=Open+Sans:300,400,600" rel="stylesheet">
```
andrew-frueh: Curious if we still need this `<link>` to Open Sans now that you've done all the awesome work in #104?

jacquie:
Also wondering if the full color triflame would look nice?

Issue #58, opened 2018-01-30
Custom Catalyst Icon Font


corykon:
_...recreating this issue. We've got a nifty icon component now , but still have a desire to roll our own icon font at some point..._

Develop an HC-Icons font to remove our dependency on FontAwesome, GlyphIcons, or any other 3rd party glyph set. This document represents the current best understanding of what icons should be included in this font:

https://spark.healthcatalyst.com/docs/DOC-3157

Please look it over and let us know us know if anything is missing that you currently make use of in your apps.

andrew-frueh:
@corykon - is this still something we want to do?  I could probably create the font, and it could potentially solve #412.  But I worry that if we only use a subset of FontAwesome and a few others, that there are going to be apps that need icons we're not including.  In that case do we just let them create their own dependency on FontAwesome?  And if they're going that anyway, does that defeat the point of having a custom font?

corykon:
@andrew-frueh Good question. I still think it would be awesome, but it might be easier to justkeep  pointing people to font awesome for now. Personal opinion, I'd push for using FA 5 over 4. I think alot of those just look better.

Also, I've got a batch of custom, healthcare related stuff, and their are several other icons we use in popbuilder that aren't available in font awesome that others may or may not find useful. Perhaps we stick those SVGs (or a font) in a repo or available for download somewhere. Though I suppose if they've not been asked for by anybody yet, maybe they're not really needed?

![image](https://user-images.githubusercontent.com/12416432/44161334-40e4f080-a07a-11e8-81bb-1297a9dbe279.png)

andrew-frueh:
Ah, that's interesting - I bet people would use them if they knew they were available.  So maybe that would be the best use for a Catalyst font - not to replace FA, but for any icons we use that aren't already in FA?

andrew-frueh:
Suggestion from the steering committee is to create a page in Styles called "Icons" that allows the user to download svgs of the standard iconography.

PR #114, opened 2018-02-12
Button demo page cleanup

andrew-frueh:
Primarily cleaning up the content on our button demo page now that `<hc-split-button>` has been added.  Adjusted all our dropdown indicators to use the down chevron for consistency.  Removed the minimum width on split buttons because it was making them unnecessarily wide.

I noticed after the recent update to our styles that `<a hc-button` isn't working - it's showing up as a primary button.  I pulled it out of the demo page, but I'm also wondering if there is any benefit to keeping anchors as a part of hc-button?  Anchor links can often be within blocks of text, so we really don't want all the padding, etc that we have on our normal buttons - we'd have to unset all that.  So might it be better to include `<a>` styling in our new sass folder?  Either in a new anchors.scss file or in cashmere.scss?  There might be a good reason I'm missing for keeping anchors as a component, but I figured it's worth asking.  If we decide to switch it that can be a separate PR.

### DIFF
--- a/lib/src/button/button.component.scss
+++ b/lib/src/button/button.component.scss
@@ -31,7 +31,7 @@
     min-width: 150px;
     font-size: 15px;
     line-height: 35px;
-    margin-bottom: 0;
+    margin: 0 4px 0 4px;
     font-weight: normal;
     text-align: center;
     white-space: nowrap;
@@ -78,6 +78,13 @@
         @include colorButton($gray-100, 5%);
         color: $gray-600;
         border: 2px solid $gray-300;
+        height: 37px;
+        line-height: 33px;
+        padding-right: 19px;
+
+        &:hover {
+            color: $offblack;
+        }
     }
 
     &.hc-tertiary {
--- a/lib/src/button/button.component.scss
+++ b/lib/src/button/button.component.scss
@@ -123,14 +123,11 @@
         border-top-left-radius: 0;
 
         &::after {
-            display: inline-block;
-            width: 0;
-            height: 0;
-            vertical-align: middle;
-            content: "";
-            border-top: .3em solid;
-            border-right: .3em solid transparent;
-            border-left: .3em solid transparent;
+            display: flex;
+            height: 20px;
+            align-items: center;
+            font: normal normal normal 12px/1 FontAwesome;
+            content: "\f078";
         }
     }
 }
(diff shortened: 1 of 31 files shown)


jtrujill:
`<a hc-button></a>` was designed to style a link like a button. There is reason to keep it because if you want to link to another page using a button, then the appropriate element to use is an anchor. If you want to style anchors that don't have hc-button then that should be placed in a style sheet somewhere.

andrew-frueh:
But why wouldn't you just use a tertiary button in that case?  They would be identical to each other visually, no?

jtrujill:
The difference is that it's no longer a link but a button. If we wanted to link to another page we would need to either wrap that button in an anchor or we'd need to listen for the click event and handle it in our component. This would also mean that we are no longer being semantically correct with our html, since we're using buttons as links.

This also limits the styling that you can use for these anchors to the tertiary color, but what if you have an anchor that navigates somewhere important and you want to show that button as being the primary action to perform on the page.

andrew-frueh:
Ah okay, I get it now.  Thx for the explanation.  I'll try and figure out why it's currently broken and fix it as a part of this PR.

jtrujill:
What's broken? I can do a fix for it. It's best to keep PR focused on one thing

andrew-frueh:
Scratch that - it's working just fine; especially now that I understand better the purpose for having `<button>` and `<a>` versions.  I was thrown off because the anchor version was showing up with the primary style, since color wasn't set.  But now I see that a user might want the `<a>` tag set to any of our color options.  I added the content related to the anchor version back to the demo page and expanded the explanation slightly (for people like me).

I did leave it off of the top section of the page though because visually the anchor and button versions look the same.  So the most important thing in the documentation is in explanation that follows to show developers that they can leverage the styles in either tag.

lib/src/button/button.component.scss
```diff
@@ -123,14 +124,11 @@
...
-            vertical-align: middle;
-            content: "";
-            border-top: .3em solid;
-            border-right: .3em solid transparent;
-            border-left: .3em solid transparent;
+            display: flex;
+            height: 20px;
+            align-items: center;
+            font: normal normal normal 12px/1 FontAwesome;
+            content: "\f078";
```
corytak: Will there ever be scenario where we'll need to use another icon/character for the split button toggle? If not, I'm guessing this could be revised to use the hc font when it's ready rather than FontAwesome.

andrew-frueh: Totally agree, when #58 is ready, this should definitely get switched to the Catalyst font.  I noticed @corykon called out the same in one of his table.scss files.

PR #120, opened 2018-02-12
Fix IE svg issue in bootloader

AdamGMiller:
Update pre-bootloader for document app to handle svg-related issue where svgs don't properly display as a background-image css style in IE.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -37,12 +37,6 @@
                 top: 40%;
             }
 
-            #pre-bootstrap h1 {
-                font-size: 28px;
-                line-height: 33px;
-                margin: 5px;
-            }
-
             #pre-bootstrap .loader {
                 border: 3px solid #f3f3f3;
                 border-radius: 50%;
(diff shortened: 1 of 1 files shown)


jacquie:
The reason I put the tri-flame above was because (and this might just be me!) positioned right next to the logo it looks like it's leaning over to the left slightly.

It does work in IE ok.  Pity it can't deal with svg background images :(

AdamGMiller:
I ended up grabbing what I knew was generally working from Measures Manager, but I have zero investment in where the TriFlame should go.  Happy to defer to Andrew and put it wherever it makes sense.

docs/index.html
```diff
@@ -89,16 +70,14 @@
...
-        <div id="pre-bootstrap">
-            <div class="messaging">
-                <h1>
-                    <img src="./assets/TriFlame.svg" style="width:24px">
-                </h1>
-                <div class="logo-background-brand">
+                <div id="pre-bootstrap">
+                    <div class="messaging">
+                        <h1>
+                            <img src="./assets/TriFlame.svg" width="17" height="50"> <img src="./assets/CashmereAppLogo.svg" width="200" height="50">
```
andrew-frueh: Thanks for doing this @AdamGMiller!  Just a couple adjustments to this line, the tri-flame should be a little bigger, and on a separate line from the logo (or else Paul will get mad that we're messing with the master brand):
`<br>`

AdamGMiller:
Okay - changes updated.  Appreciate the feedback!

PR #123, opened 2018-02-14
Typography and sass fun

### LINKED ISSUE #115: Typography
create typography styles that establish size and font standards with sass variables and mixin helpers

benjanderson:
resolves #115


docs/app/demo/button/button-demo.component.ts
```diff
@@ -3,10 +3,7 @@ import { Component } from '@angular/core';
 @Component({
     selector: 'hc-button-demo',
     templateUrl: './button-demo.component.html',
-    styleUrls: [
-        './button-demo.component.scss',
-        '../shared-demo-styles.scss'
```
benjanderson: I moved the shared styles into the `styles.scss`  doing this was a little redundant

docs/app/demo/shared-demo-styles.scss
```diff
@@ -1,129 +0,0 @@
-@import "colors";
```
benjanderson: most was moved into styles.scss and some was moved into _typography.scss

docs/app/home/home.component.scss
```diff
@@ -136,8 +136,8 @@ section button {
     display: table;
 }
 
-@media screen and (max-width: 740px)
-{
+
+@include media-breakpoint-down(md) {
```
benjanderson: This was a mixin I slipped in to make media queries easier, I got the idea from the bootstrap grid system..

https://getbootstrap.com/docs/4.0/layout/grid/

docs/styles.scss
```diff
@@ -18,3 +17,101 @@ body {
     // navbar spacing
     padding-top: 53px;
 }
+
+@mixin ignore-demo() {
```
benjanderson: these styles were messing up my typography demo, so I made a mixin to ignore them...

lib/src/sass/_mixins.scss
```diff
@@ -1,15 +1,27 @@
-/// Slightly lighten a color
-/// @param {Color} $color - color to tint
-/// @param {Number} $percentage - percentage of `$color` in returned color
-/// @return {Color}
-@function tint($color, $percentage) {
```
benjanderson: moved to _functions.scss

lib/src/sass/_reset.scss
```diff
@@ -10,7 +12,7 @@
     margin: 0;
     padding: 0;
     border: 0;
-    font-size: 100%;
+    font-size: 1rem;
```
benjanderson: I converted all font sizing to rems

lib/src/sass/_variables.scss
```diff
@@ -0,0 +1,11 @@
+$default-font-size: 14px !default;
+
+$default-font-family: 'Open Sans', 'Helvetica Neue', Helvetica, Arial, sans-serif !default;
+
+$grid-breakpoints: (
```
benjanderson: media query break points....

lib/src/subnav/subnav.component.ts
```diff
@@ -1,13 +1,12 @@
-import { Component, Input } from '@angular/core';
+import { Component, Input, HostBinding } from '@angular/core';
 
 @Component({
     selector: 'hc-subnav',
-    template: `<div class="subnav" [ngClass]="{'fixed-top': fixedTop}"><ng-content></ng-content></div>`,
+    template: `<ng-content></ng-content>`,
```
benjanderson: I wanted to add padding to this and couldn't because the .subnav div was inside its parent and angular scopes the styling. Thus I moved everything to the root element

package.json
```diff
@@ -8,6 +8,7 @@
         "start": "ng serve",
         "build": "ng build",
         "test": "ng test",
+        "style": "ng serve --sourcemap --extractCss",
```
benjanderson: This allows you to view the source maps of all the styles, very handy when debugging style issues....

docs/styles.scss
corykon: This for sure would've confused me had I not seen your comment. Could you override the header margins in your typography demo page? 

I think you could just do this in `typography-demo.component.scss`:
```
.ignore-demo {
    margin: 0;
}
```

Then totally wipe out the ignore-demo mixin thing in the main styles.scss file.

package.json
corykon: :+1::+1:

docs/app/demo/typography/typography-demo.component.html
```diff
@@ -0,0 +1,422 @@
...
+            <span class="type-label">Line Height</span>
+            inherit
+          </li>
+        </ul>
+      </td>
+      <td>
+        <ul>
+          <li>
+            <span class="type-label">Weight</span>
+            700
```
andrew-frueh: I've found that font-weight: 700 is really heavy.  Can you change this to 600 (and in the typography scss too)?  I'm probably going against the style guide in places here, but that's the beauty of building this library - to actually see what these values look like in context.

docs/app/demo/typography/typography-demo.component.html
```diff
@@ -0,0 +1,422 @@
...
+          </li>
+        </ul>
+      </td>
+    </tr>
+    <tr>
+      <td>
+        <strong>strong. Strong importance</strong>
+      </td>
+      <td>
+        <code>&lt;code&gt;&lt;/code&gt;</code>
```
andrew-frueh: I think you want this to be `<strong></strong>`

docs/app/demo/typography/typography-demo.component.html
```diff
@@ -0,0 +1,422 @@
...
+  </p>
+  <h3>Fonts</h3>
+  <h4>Open Sans</h4>
+  <em>Our primary font for all things web.</em>
+  <p>
+    <a href="https://fonts.google.com/specimen/Open+Sans">https://fonts.google.com/specimen/Open+Sans</a>
+  </p>
+  <h4>Flexo</h4>
+  <em>The Health Catalyst corporate font. Only to be used for product names displayed in the header</em>
+  <img alt="cashmere app logo" src="./assets/CashmereAppLogo.svg" style="background-color:#384655; width:500px">
```
andrew-frueh: Could you add `margin-top: 15px` if we're doing inline styles on the logo?  Just so it doesn't run right up against the paragraph above.

corykon:
A couple other thought on the fonts, worth discussing here or elsewhere  

1. I believe these font styles are pulled from the style guide, and look pretty good, but IMO some of the font weights feel too light for a web interface
2. Do we encourage everyone to use REMs for font sizes? I understand the benefits, but I've never been totally convinced they're worth the pain to work with. If we do in fact prefer everyone uses REMs for fonts, worth mentioning that other UI components (anything besides fonts) really shouldn't be measured out using REMs. (You can but its totally not worth it). 
3. Again, if we do suggest REMs for all font sizing, a couple approaches that make it a little smoother to use. 
- Setting the body font size to 10px to make REM calculating easier. With 10px for the body, 1.6rem = 16px, 1.4rem = 14px, etc. (Setting body to 10px can cause some silly small font sizes if you aren't being explicit about it in your css; but its obviously easy to catch this.)
- Creating a function that calculate precise REM size based on a given pixel size, so you don't constantly have to go look up or recalculate your REM sizes

andrew-frueh:
Completely agree on the first point @corykon - I think our best bet is to have Ben merge this first pass, then you and I take a look at the font definitions and make some adjustments.  Good suggestion on number three if we're gonna use rems - which we probably should.  I agree px feel a lot easier to work with, but probably at least from an accessibility standpoint, we should be doing our text in rems.

docs/styles.scss
benjanderson: I like that....

docs/app/demo/typography/typography-demo.component.html
benjanderson: I am moving this to the style sheet...

Issue #117, opened 2018-02-12
Checkbox / Radio Button labels are no longer vertically centered after CSS reset


andrew-frueh:
I noticed after the css reset was added, there are a couple styling issues that have cropped up.  Two that I noticed today are in the checkboxes and radio buttons.  Our text labels are no longer centered vertically with their box/button to the left.

@nickrflynn - do you have a few minutes to take a look at these?  If not, I can probably work on it later today.

andrew-frueh:
Also on this PR - take a look at the hc-icon demo page, the labels to the right of the example icons aren't centered anymore either.

nickrflynn:
I will take a look at this @andrew-frueh

andrew-frueh:
Thanks @nickrflynn!  One last thing so we can lump a bunch of small tweaks into one PR: I noticed on the Checkbox and Radio button demo pages that you're using unordered lists for your "usage" section content.  Can we switch that to `<p>` instead so that it matches the other pages and takes advantage of the global demo styles.

PR #124, opened 2018-02-14
Fix vertical alignment for checkbox and radio button components

nickrflynn:
Fix for #117


nickrflynn:
Just saw your comments regarding using `<p>` on the demo pages @andrew-frueh.  I'll push changes for that shortly.

lib/package-lock.json
```diff
@@ -480,7 +480,6 @@
             "requires": {
                 "anymatch": "1.3.2",
                 "async-each": "1.0.1",
-                "fsevents": "1.1.3",
```
andrew-frueh: Out of curiosity @nickrflynn - are the three changes to package-lock.json, package.json, etc intentional?

benjanderson: sometimes when you run `npm install` it changes this file, especially if the collaborator are in two different enviroments ie. mac and windows. npm tries its best to keep it consistent but fails alot

PR #130, opened 2018-02-16
Fixes Button Hover

### LINKED ISSUE #129: Buttons lost hover state
Looks like on one of our recent merges, we introduced a style that is conflicting with the hover state on our buttons - either that or they aren't finding the shade() function.

jtrujill:
Fixes button hover state #129

### DIFF
--- a/lib/src/button/button.component.scss
+++ b/lib/src/button/button.component.scss
@@ -1,4 +1,5 @@
-@import 'colors';
+@import '../sass/colors';
+@import '../sass/functions.scss';
 
 /// Sets up colors for a button
 /// @param {Color} $color - base color of the button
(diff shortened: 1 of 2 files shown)


lib/src/button/button.component.scss
```diff
@@ -1,4 +1,4 @@
-@import 'colors';
+@import '../sass/cashmere';
```
benjanderson: if you include the whole cashmere bundle in components you duplicate all the styles. I would suggest to only include what you need, ie `functions`

also with `includePaths` you no longer have to remember the relative path to the common sass files, you can just do `@import 'functions';`

At least it appears that way when I debug
image

jtrujill:
@benjanderson I fixed that issue, but noticed all components import from the cashmere file, so another issue should be created for that

PR #131, opened 2018-02-16
Demo Site Layout Overhaul

### LINKED ISSUE #126: Updated Demo Site Layout
We've had some conversations about updating the layout and navigation of our demo site, and this is what I'm proposing as an upgrade:

https://xd.adobe.com/view/7a081799-3095-46c2-b22b-3b9f6393eb1d/

It borrows from some patterns on the Angular Material site and other documentation sites I like.  Rather than separating the component demo and documentation into two separate navbar tabs, @corykon had the great suggest to include horizontal tabs on a single component section.  That way the developer info is still separate, but it keeps all the info about a single component in one place.

I also gave a little thought to how the page might work on a small screen:
https://xd.adobe.com/view/7a081799-3095-46c2-b22b-3b9f6393eb1d/screen/b2eb28d7-8e5b-40b4-8f1c-dbbcf479239f/Small-Screen

Let me know what you think about the proposed look and feel revision.  Thanks!

andrew-frueh:
Resolves #126.  Breaks our content out into "Styles" and "Components" in the navbar.  We can add "Guides" as soon as you have the "Getting Started" page ready, @benjanderson.  Makes use of @corytak's vertical and horizontal tabs for our navigation on those pages.  Also added a `hc-tile` component which is heavily used throughout the demo site (and also in many Catalyst apps).

@corykon - totally agree with your comments, and let's open an additional issue to do an audit of our documentation and decide on a template for all those pages.  Also, the only reason I went with "Documentation" over "API" was because it made for a bigger button - API was so small that it felt like it was getting lost.

Still need to make the content of these pages responsive, but I'll make that a separate issue - this PR is already gigantic.


lib/src/tabs/tab-set.component.scss
```diff
@@ -1,17 +1,22 @@
...
+}
+
 .tab-bar-vertical {
     padding-top: 15px;
     background-color: $white;
     width: 20%;
-    float: left;
-    height: calc(100% - 15px);
+    display: table-cell;
+    vertical-align: top;
```
andrew-frueh: @corytak - the issue I was running into was that if the tab bar was taller than the content, the gray background did not extend to the entire height of the tab bar.  Basically it's the issue of trying to match two floating divs with variable heights.  I researched and tried a number of approaches, but using display: table seemed to be the best cross browser solution.  Works well on the demo site, but let me know if you think this is going to cause problems down the road.

corytak: Works for me. 👍

lib/src/tile/tile.component.ts
```diff
@@ -0,0 +1,13 @@
+import { Component, HostBinding } from '@angular/core';
+
+@Component({
+    selector: 'hc-tile',
```
corytak: This is a cool new component!

PR #134, opened 2018-02-19
Getting started

### LINKED ISSUE #97: Standard Template for Documentation
I would like to change the way that we do documentation so that we're using markdown instead of straight html. With css and html, there is too much freedom in the way to style the documentation, which makes writing it and maintaining it difficult.  Markdown is simpler to write and would allow us to have more uniform docs.

The other issue is that for a user to read about how to use our library they must have it running locally or go to http://cashmere.azurewebsites.net/. It would allow people to read on docs on github or read it locally without running a server.

It's still great that we have demo usage of the app but I'd like to separate the two.

benjanderson:
I am not finished with this pull request, I just want to get feedback on the direction I am going. I am not offended if you do not like what I did, so please give honest feedback.

Here are things I did:

- added guides section / getting started
- import markdown through webpack `require('raw-loader!./index.md')`. This wouldn't work with the rollup build but it doesn't matter as these imports aren't a part of the library
- added directive that converts markdown to html and supports `highlight js` highlighting
  - this was added to the demo section not the library
  - added dependency to demo on marked, a markdown parsing library. Not added to library

Questions I have:
- importing markdown does help resolve #97 What do you think of this solution?
  - would we convert all the other docs to this?
- Should the markdown directive / marked dependency be added to the library?
  - this would add support to highlight SQL which atlas uses
- I am struggling with some styling and will get with you @andrew-frueh 

I based my branch off of @andrew-frueh recent reorg pull request to avoid too much hassle with conflicts.


docs/app/guides/getting-started/getting-started.component.html
```diff
@@ -0,0 +1 @@
+<div [markdown]="doc" [lineNumbers]="false"></div>
```
andrew-frueh: Would be happy to chat about how best to format the markdown.  I've seen in the past that markdown docs usually have a "template" html file that the content is being inserted into, does the same apply here?  I didn't see a template but admittedly I didn't look all that hard.  One thing that will help the formatting out on this page is to wrap this div in `<div class="demo-content">` - either on here on in the template.  And then we should talk about how best to put this content into tiles as well.

jtrujill:
I like think the splitting of components is good, but for now I'd rather keep our documentation as-is. Later I'd like to have a whole directory devoted to documentation written in markdown, which can be run through a markdown => html converter to be linked with the demo page.

As part of the markdown PR we should standardize the layout of our app with other projects and place both our lib/ and demo/ under src/. Then we could put our markdown documentation under the docs/.

In issue #97 I gave a brief overview of what we could do with markdown.

I can take on the markdown issue after I clear pushing the demo app to Azure and the lib to NPM

benjanderson:
@jtrujill Wow, I spent a good half hour making a reply that I totally deleted. 

I wasn't confident that we could generate the html files and make a smooth spa experience. However looking into how material does it with their style guide, it seems they do something similar to what you proposed in #97 . 

Here is their guide list: https://github.com/angular/material.angular.io/blob/master/src/app/shared/guide-items/guide-items.ts
Here is how they bind that Guide object to html: https://github.com/angular/material.angular.io/blob/master/src/app/pages/guide-viewer/guide-viewer.html
Here is the doc viewer directive that ultimately serves everything: https://github.com/angular/material.angular.io/blob/master/src/app/shared/doc-viewer/doc-viewer.ts

I am ok if we want to go that direction, but I still think my solution is much simpler. We could still have a `docs` directory in the root of the repo and just reference it from each guide component eg. `require('raw-loader!../../../docs/index.md');`

It would be good to loop one of the Corys into the conversation to get a second opinion @corykon  or @corytak I will invite them to the 2pm discussion

joeskeen:
Sorry, a little late to the party.  A couple of things:
* I looked at the code, it looks good to me.
* I would like to see it in action, though.  I think I'll need to pull it down to inspect it further.
* I'm wondering which code highlighting library would be best to go forward with if we are going to add it to the library (this might be better as a separate issue).  I've used highlight.js in the past and it's always worked for me, but I recently discovered that Atlas uses PrismJS.  I'm not sure which of these (or something else) would work best, but we should have that discussion before bringing it in.

Issue #63, opened 2018-01-30
Filter Drawer


andrew-frueh:
@jtrujill got us started on this by creating a drawer component #42.  I'm wondering if we can create an additional component that leverages Jeremy's drawer specifically for filtering.  The reason I think a filter drawer component is worth including is because of how common this use pattern appears in our apps.

In my mind, the way this would work would be something like:
`<hc-filter-drawer visible='false' filterSets='true'>
     <hc-filter target='agePicker'>Patient Age</hc-filter>
     <hc filter target='payerPicker'>Payor</hc-filter>
</hc-filter-drawer>`

_Some additional functionality info below:_
Within the drawer users can save Filter Sets to easily flip between previously saved sets of filter selections. Clicking on a filter row typically triggers a modal with a picklist where specific selections can be made.

The filter drawer integrates with filter chiclets that are visible in header row of the page content next to the button used to open/close the filter drawer. The chiclets tell the user what filters are currently active and allow the user to clear specific filters without opening the drawer.

A clickable prototype to demonstrate the functionality described above is available here:
https://xd.adobe.com/view/5260b164-4d93-4238-8fef-8bde1575f248/screen/9180b57f-a454-46eb-bf46-fe1bbc8a005a/Dashboard-Filters

![35475134-4ac49752-0356-11e8-97d7-031f4a047147](https://user-images.githubusercontent.com/22795893/35576384-12dee732-059c-11e8-9404-24b3cfef970a.png)

jtrujill:
I think we should try to keep this library more general, as a tool for a developer. What I was going to do was create a list component that can be used to generate a list in the filter as you see in the image above.

If we make the library too specific it'll restrict its use and make it more difficult to use. What we can do in the design of our components is style other library components within our library component without introducing new attributes/properties to our components.

andrew-frueh:
You make a fair point @jtrujill - and I completely hear you about general components.  Which is why I think making a drawer component first as you did was the right thing to do.  And having those general components as tools for developers as you said makes perfect sense.  So I'm not advocating that we overwrite the general component with a filters only component.  I'm suggesting we have both.  

[...]

Issue #61, opened 2018-01-30
Picklist


andrew-frueh:
Much credit to @corykon and the Population Builder team for developing a pattern that has been followed by a number of apps.  The picklist is called upon when a user needs to select a number of items from a potentially length list of options.  Although the picklist component would most commonly be used within a #40 modal dialog, it can stand alone as well.  Cory, I know I've already dropped the massive "table" component on your plate, but since you set the standard here, I think you'd be the most qualified to componentize this.

![picklist](https://user-images.githubusercontent.com/22795893/35548996-5a17da0c-0540-11e8-8255-62624a7f2575.png)

jtrujill:
This is similar to a the list component that I've created. I'm going to be expanding it's capabilities further so that it can have a header and divider between items. 

I'm thinking for this component that it'd be best to provide developers with the tools to accomplish 80% of this and have them implement the rest. There are just so many ways that this could be different between applications, that it'd be very difficult to provide them a component that would be useful for their particular needs.

I think the most difficult part about this is creating a selectable list, which is something I plan to write. It will build off of the list component and will provide an easy to use interface that would allow you to do all of the above. Once the selectable list is implemented the implementation of this design would be very easy.

If you can just assign this to me, I can look at extending the list component

corykon:
Sounds good, Jeremy. For reference, here's where you can play with what we've got in builder:
- [https://hcs-clinicalppr.hqcatalyst.local/test/ruleset/new](https://hcs-clinicalppr.hqcatalyst.local/test/ruleset/new)
- New Ruleset > Drag out a Diagnosis Fitler > open "Diagnosis Codes" from drawer

Would be good to make note of the functionality that is there- not that you'd implement it all, but would be cool if you could try to match the feel of the picklist that's there, and architect things such we'd be able to build on top of the cashmere component and achieve the functionality that's there.

andrew-frueh:
One bit of commentary, @jtrujill - I agree that this type of picker _could_ be very custom from app to app, but we do in fact have three or four Catalyst apps that use it exactly as @corykon has it built out in PopBuilder.  I'm all for customization, but my request would be that the PopBuilder style be the default so the multiple apps using the same functionality can plug and play.

jtrujill:
Maybe in addition to the selection list I can look at making a component which allows you to transfer objects from one list to the other. I don't think that would be too difficult and would reduce the amount of work required of a user to implement the design.

jtrujill:
[...]

PR #135, opened 2018-02-20
Added list subheader and divider component and updated list styles

jtrujill:
This is a precursor to implementing #63 and #61


lib/src/divider/divider.component.scss
```diff
@@ -0,0 +1,11 @@
+$hc-divider-width: 1px;
+
+.hc-divider {
+    display: block;
+    margin: 0;
```
andrew-frueh: Recommend changing this to `margin: 10px 0;` so we have a little more whitespace between sections

lib/src/list/list.component.scss
```diff
@@ -1,3 +1,96 @@
...
+}
+
+.hc-list-subheader {
+    display: flex;
+    box-sizing: border-box;
+    align-items: center;
+    padding: $hc-list-side-padding;
+
+    .hc-list & {
+        margin: 0;
```
andrew-frueh: Recommend adding `padding-bottom: 3px;` here to tighten up the space between the header and the first list item

lib/src/divider/divider.component.scss
jtrujill: I'm using the padding between lists to have whitespace between elements. By setting the margin to 0, I ensure that any element that has a margin defined will be set to 0. This makes the elements more uniform

lib/src/list/list.component.scss
jtrujill: Same as above padding is being used for "whitespace" between elements. Margin: 0 ensures that elements that have the directive hcListSubheader defined don't interfere with the "whitespace" of list elements

jtrujill:
@andrew-frueh I rebased the latest changes from master so that this will now merge without issue

PR #143, opened 2018-02-22
Chip Component

### LINKED ISSUE #118: Chip Component
Primarily used in conjunction with a filter drawer, but there may be other uses as well.  I think the [Angular material version](https://material.angular.io/components/chips/overview) gives us a pretty good model to follow.  As in that component, having a standard chip and an "input chip" directive that can be closed is probably a good idea.

chipexample

andrew-frueh:
Resolves #118.  Includes a few different flavors of chips, as well as the ability to define whether or not it should be removable.  Begging forgiveness from @jtrujill - in addition to the base component, I've also included two optional components that a dev can use if it saves them time, or ignore if it makes more sense to use their own.  There is a chip-row component that has the ability to constrain the set of chips to one line with overflow ellipses or display multi-line.  And there is also a filter button that can be used in the chip row which includes our non-FA icon.

Thanks to @corykon for his suggestion on how to approach the single line overflow functionality.


andrew-frueh:
Shoot - the text-overflow is working great in Firefox (see below) but in Chrome it's just clipping the overflow.  Any ideas @corykon?

textoverflow

lib/src/chip/chip-row/chip-row.component.ts
```diff
@@ -0,0 +1,23 @@
...
+import { ChipComponent } from '../chip.component';
+import { FilterButtonComponent } from '../filter-button/filter-button.component';
+
+@Component({
+    selector: 'hc-chip-row',
+    templateUrl: './chip-row.component.html',
+    styleUrls: ['../chip.component.scss']
+})
+export class ChipRowComponent implements OnInit {
+    @HostBinding('class') hostClass = 'chip-row';
```
jtrujill: I believe that binding to the class attribute directly will whip out any classes the user may have put on there. I would verify and if so change it to
`@HostBinding('class.chip-row') hostClass = true;`

lib/src/chip/chip-row/chip-row.component.ts
```diff
@@ -0,0 +1,23 @@
...
+    @Input() singleRow: boolean = false;
```
jtrujill: just being picky about the name but maybe change to wrap.

You'll also want to create a getter and setter for the input because you want the user to be able to assign a string to this input and you'll need to convert that to a boolean. You can check out button.component.ts's disabled input

lib/src/chip/chip-row/chip-row.component.ts
```diff
@@ -0,0 +1,23 @@
...
+    contentClass: string = 'chip-row-contents';
+
+    constructor( ) {
+    }
+
+    ngOnInit() {
+        if ( this.singleRow ) {
```
jtrujill: You'll need to monitor the changing of singleRow so that you add/remove this single-row class. You can do this through the setter method for singleRow

lib/src/chip/chip.component.html
```diff
@@ -0,0 +1,3 @@
+<div [ngClass]="chipType" [style.cursor]="chipCursor" [style.display]="chipDisplay">
```
jtrujill: I wouldn't use a style attribute since that takes styling away from the user of the component. This should be defined in the chip.component.scss

lib/src/chip/chip-row/chip-row.component.ts
```diff
@@ -0,0 +1,23 @@
+import { Component, HostBinding, OnInit, Input } from '@angular/core';
+import { ChipComponent } from '../chip.component';
+import { FilterButtonComponent } from '../filter-button/filter-button.component';
+
+@Component({
+    selector: 'hc-chip-row',
+    templateUrl: './chip-row.component.html',
+    styleUrls: ['../chip.component.scss']
```
jtrujill: You'll want to add encapsulation: ViewEncapsulation.None. We'll want to give user's the power to alter these

lib/src/chip/chip.component.scss
```diff
@@ -0,0 +1,77 @@
...
+
+.chip-row-button { display: inline-block; }
+
+.chip-row-contents { display: inline-block; }
+
+.single-row {
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    overflow: hidden;
+}
```
jtrujill: I just want to congratulate you. These styles are nicely defined. Too often people make css definitions extremely specific

lib/src/chip/chip.component.ts
```diff
@@ -0,0 +1,27 @@
+import { Component, Input, OnInit } from '@angular/core';
+
+@Component({
+    selector: 'hc-chip',
+    templateUrl: './chip.component.html',
+    styleUrls: ['./chip.component.scss']
+})
+export class ChipComponent implements OnInit {
+
+    @Input() action: boolean = false;
```
jtrujill: Same as the ChipRowComponent, you'll want to create a getter and setter to handle the user setting a string value to this input

lib/src/chip/chip.component.ts
```diff
@@ -0,0 +1,27 @@
...
+    @Input() color: string = 'neutral';
+    chipCursor: string = 'auto';
+    chipDisplay: string = 'inline-block';
```
jtrujill: I've already mentioned this but this should be moved to the scss file

lib/src/chip/chip.component.ts
```diff
@@ -0,0 +1,27 @@
...
+    chipType: string = 'chip';
+
+    constructor() {
+    }
+
+    ngOnInit() {
+        if (this.action) {
+            this.chipCursor = 'pointer';
+            this.chipType += ' close';
```
jtrujill: This action could change so you'll want to move this logic so that it gets re-evaluated. Otherwise, win the user changes the action variable nothing will happen

lib/src/chip/chip.component.ts
```diff
@@ -0,0 +1,27 @@
...
+        }
+
+        this.chipType += ' ' + this.color;
```
jtrujill: The += will cause an issue when you change this to re-evaluate when the action or color change so make sure you change the way you're handling this

lib/src/chip/filter-button/filter-button.component.ts
```diff
@@ -0,0 +1,22 @@
...
+    @Input() labelOff: string = 'Filters off';
+    @Input() labelOn: string = 'Filters on';
+    @Input() buttonState: boolean = false;
+    buttonLabel: string = this.labelOff;
+
+    constructor() {
+    }
+
+    ngOnChanges() {
+        this.buttonLabel = (this.buttonState) ? this.labelOn : this.labelOff;
```
jtrujill: It's easier to handle this in the setter function for the buttonState so move this logic there

lib/src/chip/filter-button/filter-button.component.ts
```diff
@@ -0,0 +1,22 @@
...
+    selector: 'hc-filter-button',
+    template: `<img src="./assets/FilterIcon.svg" class="filter-img">{{buttonLabel}}`,
+    styleUrls: ['../chip.component.scss']
+})
+export class FilterButtonComponent implements OnChanges  {
+    @HostBinding('class') hostClass = 'filter-button';
+
+    @Input() labelOff: string = 'Filters off';
+    @Input() labelOn: string = 'Filters on';
+    @Input() buttonState: boolean = false;
```
jtrujill: getter/setter to handle string boolean case

lib/src/chip/filter-button/filter-button.component.ts
```diff
@@ -0,0 +1,22 @@
+import { Component, HostBinding, Input, OnChanges } from '@angular/core';
+
+@Component({
+    selector: 'hc-filter-button',
+    template: `<img src="./assets/FilterIcon.svg" class="filter-img">{{buttonLabel}}`,
+    styleUrls: ['../chip.component.scss']
+})
+export class FilterButtonComponent implements OnChanges  {
+    @HostBinding('class') hostClass = 'filter-button';
```
jtrujill: This could wipe out a user's class that they define on the element. Go with
`@HostBinding('class.filter-button') hostClass = true;`

lib/src/chip/filter-button/filter-button.component.ts
```diff
@@ -0,0 +1,22 @@
+import { Component, HostBinding, Input, OnChanges } from '@angular/core';
```
jtrujill: After seeing what the filter-button is exactly. I'm thinking this would be better left as a template style guide. The reason is that I don't think it's difficult to implement and I feel it **steps out-of-bounds for what I'd expect this component to do**

jtrujill:
@andrew-frueh Sorry if that seems a bit harsh. I'm going in for Lasik tomorrow and will be out for a bit. Once you get those things fixed I think your component is good to go

andrew-frueh:
Awesome review @jtrujill - thanks for all the input.  Let me sort through all this over the next couple days and I'll let you know if I have ahy follow up questions.  

And for the record, I'm not opposed to moving the filters button out of this module.  On second thought today, I think you make a fair point that it doesn't really fit with the goal of this component.  So I can move that into that component to just be part of the demo - not the library.  And then folks will still have a pattern they can follow if they want, and they can grab the custom icon there as well.

Good luck with the surgery today!

andrew-frueh:
Thanks again @jtrujill - the feedback was very helpful (and educational).  Component feels much more streamlined now.  I yanked the filters button out of the component, and it's stashed in the component demo for now (for display purposes only).  As soon as we get the Guide for "advanced filters" written, I'll pull it out of the chip demo entirely and move it over there.

lib/src/chip/chip.component.ts
```diff
@@ -0,0 +1,38 @@
...
+    selector: 'hc-chip',
+    template: `<div [ngClass]="chipType"><ng-content></ng-content></div>`,
+    styleUrls: ['./chip.component.scss'],
+    encapsulation: ViewEncapsulation.None
+})
+
+export class ChipComponent implements OnInit {
+
+    _action: boolean = false;
+    @Input() color: string = 'neutral';
```
corykon: I see that `neutral`, `red`, `yellow`, and `green` are the color options, so we could enforce that a little more, rather than taking in any string.
```
@Input() color: 'neutral' | 'yellow' | 'green' | 'red' = 'neutral';
```
Done this way, the implementer of the component should be able to get helpful tooltips in their IDE about what is available, and errors if they use an invalid color.

lib/src/chip/chip.component.ts
```diff
@@ -0,0 +1,38 @@
...
+    chipType: string = 'chip';
+
+    constructor() {}
+
+    ngOnInit() {
+        this.chipType += ' ' + this.color;
+    }
+
+    @Input()
+    get action(): boolean {
```
corykon: - If the color property were changed, that change wouldn't be reflected unless you also update the action property. 
- I think it would be possible to toggle the 'close' class based on action without having to use getters/setters

lib/src/chip/chip.component.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    set action(isAction) {
+        this._action = anyToBoolean(isAction);
+        if ( this._action ) {
+            this.chipType = 'chip close';
+        } else {
+            this.chipType = 'chip';
+        }
+
+        this.chipType += ' ' + this.color;
```
corykon: Sorry I'm throwing alot at you here, with the above comments put together, I think you could do something like this:
```
@Component({
    selector: 'hc-chip',
    template: `<div [class]="color" [ngClass]="{'close': action, 'chip': true}"><ng-content></ng-content></div>`,
    styleUrls: ['./chip.component.scss'],
})

export class ChipComponent {
    @Input() action: boolean = false;
    @Input() color: 'neutral' | 'yellow' | 'green' | 'red' = 'neutral';
}
```

lib/src/chip/chip-row/chip-row.component.ts
```diff
@@ -0,0 +1,29 @@
+import { Component, Input, ViewEncapsulation } from '@angular/core';
+import { ChipComponent } from '../chip.component';
+import { anyToBoolean } from '../../util';
+
+@Component({
+    selector: 'hc-chip-row',
+    template: `<div [ngClass]="contentClass"><div><div class="row-buffer"><ng-content></ng-content></div></div></div>`,
```
corykon: I think you could remove the getter/setter, and also get rid of `contentClass` and do this:
```
class="chip-row-contents" [ngClass]="{'single-row': !wrap}"
```

andrew-frueh:
Great stuff @corykon - thank you!  That's some new syntax I learned thanks to this around string literals and all you can do with ngClass.  I left in a very simplified version of my getter and setters on my boolean variables so I could follow  @jtrujill's suggestion to use anyToBoolean and allow users to enter strings or straight values.  But otherwise I've got all your suggestions incorporated.  Thanks again!

PR #146, opened 2018-02-24
Fully responsive demo site content

### LINKED ISSUE #132: Responsive Demo Site Content
Take the new layout for the demo site content and make it responsive.  This is what I'm proposing for the mobile version:

https://xd.adobe.com/view/7a081799-3095-46c2-b22b-3b9f6393eb1d/screen/b2eb28d7-8e5b-40b4-8f1c-dbbcf479239f/Small-Screen

The mobile version would hide the vertical tabs and display a subheader with a select component containing the navigation options.

andrew-frueh:
Resolves #132.  Makes the entire demo site responsive.  It looks great at tablet resolution, it's view-able but not exactly optimized for phone resolution.  When we hit our first breakpoint, the vertical tabs disappear and a select box appears in a subnavbar at the top of the page for navigation within the current section.  The navbar's responsive menu is used for navigating between sections.  Making use of @benjanderson's function to trigger the media breakpoints based on library variables rather than hardcoded.

Please take a look at the way I'm doing the routing in components.component.ts and styles.component.ts.  I suspect there is likely a more elegant way to do it then all those if else blocks.

And @corytak - the I had to go back on the display:table stuff in the tabs component...again.  Table display does some funny things with responsive layouts.  It's now using flex for display instead, and I added one bit of css to the demo styles (not the component) to make it full page if the content is shorter than the page - I also added a paragraph in the documentation about how to do the same if a developer wants to recreate that behavior.


docs/app/components/components.component.ts
```diff
@@ -1,15 +1,85 @@
...
+                    this.thisPage = 'Tabs';
+                } else if ( event.url === '/components/tile') {
+                    this.thisPage = 'Tile';
+                }
+            }
+        });
     }
 
+    // Handle changes to the select component and navigate
+    selectUpdate ( event: any ) {
```
corytak: For consideration: 

You could turn this piece into a function like so:
```
    navigateToRoute(path: string) {
        this.router.navigate([path]);
    }
```
Then in the template you can call that function with the appropriate path?

docs/app/components/components.component.ts
```diff
@@ -1,15 +1,85 @@
...
 
-    constructor() {
-    }
+    thisPage = 'Accordion';
 
-    ngOnInit() {
+    constructor( private router: Router ) {
+        // Listen for vertical tab bar navigation and update the select component
+        router.events.subscribe((event) => {
+            if (event instanceof NavigationEnd) {
```
corytak: Am I correct in understanding that this is listening for the current route and setting the current page? I'm wondering if injecting `private route: ActivatedRoute` and subscribing to the url would be simpler than listening to all of the router events.

docs/app/styles/typography/typography-demo.component.html
```diff
@@ -2,7 +2,7 @@
     <h1>Typography Guidelines</h1>
     <h6>Last updated {{lastModified | date:'longDate'}}</h6>
 
-    <hc-tile>
+    <hc-tile class="width-overflow">
```
corytak: It might be nice to have these responsive styles baked into the components themselves rather than requiring the developer to create a class like `width-overflow` to make them responsive.

docs/styles.scss
```diff
@@ -37,13 +35,56 @@ body {
...
     .tab-bar-horizontal { margin-bottom: 25px; }
 
+    .width-overflow{ overflow-x: auto; }
+
     code {
         max-width: 870px;
     }
 }
 
+.full-height > .vertical-container { min-height: calc( 100vh - 53px ); }
```
corytak: I've had mixed results with using `vh`. What was the difficulty presented with using `%` instead?

lib/src/tabs/tab-set.component.scss
```diff
@@ -6,16 +6,15 @@
 }
 
 .vertical-container {
-    display: table;
     height: 100%;
     width: 100%;
+    display: flex;
```
corytak: I have a feeling we're going to go back and forth on this until the end of time. ;)

docs/app/components/components.component.ts
benjanderson: agreed, this is a strange way to do this. Another option is you can add "data" to routes, so their is extra metadata when you navigate to any page https://angular.io/guide/router
image

docs/app/components/drawer/drawer-demo.component.ts
```diff
@@ -2,7 +2,8 @@ import { Component } from '@angular/core';
 
 @Component({
     selector: 'hc-sidenav-demo',
-    templateUrl: './drawer-demo.component.html'
+    templateUrl: './drawer-demo.component.html',
+    styles: [ '[hc-button] { margin-top: 10px; margin-right: 10px;}' ]
```
benjanderson: this is repeated quite a few places, would it make sense to add it to the `styles.scss` global demo styles or even make a reusable css class?

docs/app/styles/styles.component.ts
```diff
@@ -1,15 +1,39 @@
...
 })
-export class StylesComponent implements OnInit {
+export class StylesComponent {
 
-    constructor() {
+    thisPage = 'Colors';
+
+    constructor( private router: Router ) {
+        // Listen for vertical tab bar navigation and update the select component
+        router.events.subscribe((event) => {
```
benjanderson: same changes to `component.component.ts` should be applied here

docs/app/styles/typography/typography-demo.component.html
benjanderson: you could do something like the following
```scss
@include media-breakpoint-down(sm) {
     hc-tile {
     overflow-x: auto;
}}```

File: lib/src/breadcrumbs/breadcrumbs.component.scss [hunk 8]
```diff
@@ -35,7 +35,7 @@
 }
 
 .responsive-container { display: none; }
-@media screen and (max-width: 420px) {
+@include media-breakpoint-down(sm) {
```
benjanderson: :+1:

docs/app/components/drawer/drawer-demo.component.ts
andrew-frueh: Yup, agreed - I realized this after I submitted the PM.  I'll move this into the global demo styles.

docs/app/components/components.component.ts
andrew-frueh: @corytak - it's actually listening to the current route and then setting the responsive Select component value.  The idea is that if I pick "Breadcrumbs" in the vertical tabs, and then I shrink my page down - the responsive subheader with the select component navigation should say "Breadcrumbs" not "Accordion".  I like the idea of adding data to the route to handle this.  I'll probably reach out to either you or Ben to double check my approach before I build this in.

andrew-frueh:
Thanks @benjanderson and @corytak for the suggestions on the responsive Select routing.  Looks waaay better now than having a bunch of hard-coded if/else statements.

docs/app/components/components.component.ts
```diff
@@ -9,77 +9,27 @@ export class ComponentsComponent {
...
-                    this.thisPage = 'Select';
-                } else if ( event.url === '/components/subnav') {
-                    this.thisPage = 'Subnavbar';
-                } else if ( event.url === '/components/tabs') {
-                    this.thisPage = 'Tabs';
-                } else if ( event.url === '/components/tile') {
-                    this.thisPage = 'Tile';
+            if ( event instanceof NavigationEnd ) {
+                if ( activatedRoute.firstChild ) {
+                    this.thisPage = activatedRoute.firstChild.snapshot.data['title'];
```
benjanderson: very nice!

docs/app/components/components.component.ts
```diff
@@ -9,77 +9,27 @@ export class ComponentsComponent {
...
-        } else if ( event === 'Select') {
-            this.router.navigate(['/components/select']);
-        } else if ( event === 'Subnavbar') {
-            this.router.navigate(['/components/subnav']);
-        } else if ( event === 'Tabs') {
-            this.router.navigate(['/components/tabs']);
-        } else if ( event === 'Tile') {
-            this.router.navigate(['/components/tile']);
+        let root = this.activatedRoute.routeConfig;
+        if ( root && root.children ) {
```
benjanderson: I like it! ❤️

PR #150, opened 2018-02-26
Bugs/140 tooltip

### LINKED ISSUE #140: Horizontal Tabs showing title tooltip anywhere within content
@corytak - this is a minor issue but probably one we should fix.  When using horizontal tabs, if you hold your mouse anywhere within the content area, you'll get a tooltip with the "title" of that tab.  I'm guessing it's because we're using the title property to set the name.

tooltip

corytak:
Closes #140 

Rename `title` which is a standard attribute to `tabTitle` to avoid the omnipresent tooltip. 
I also noticed that after my last PR you could no longer directly route to a tab and that it would always route to the first tab and have included a fix for that.

### DIFF
--- a/lib/src/tabs/tab-set.component.ts
+++ b/lib/src/tabs/tab-set.component.ts
@@ -86,7 +86,7 @@ export class TabSetComponent implements AfterContentInit {
             this.tabs
                 .map(tab => tab.routerLink)
                 .map(routerLink => this.mapRouterLinkToString(routerLink))
-                .find(routerLink => routerLink === this.router.url);
+                .find(routerLink => this.routerMatchFound(routerLink));
 
         if (foundRoute) {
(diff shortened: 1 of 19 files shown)


lib/src/tabs/tab-set.component.ts
```diff
@@ -102,4 +102,20 @@ export class TabSetComponent implements AfterContentInit {
         }
         return routerLink;
     }
+
+    private routerMatchFound(routerLink: string): boolean {
```
corykon: Just a thought, could probably make this func a little shorter, or even keep it inline (up on ln 89):
```
const currentRoute = this.router.url;
return currentRoute === routerLink || currentRoute.indexOf(`${routerLink}/`) > -1;
```

PR #162, opened 2018-03-02
create modal components and services

### LINKED ISSUE #40: Modal Dialog
A modal dialog is a window that forces the user to interact with it before they can go back to using the parent application.  Our Catalyst modals include a header section with a title, a content area, and a footer section with Save and Cancel buttons.  The height of the header and footer should be fixed, but height and width of the overall modal should be configurable.

modaldialog

corytak:
closes #40


lib/src/modal/modal-header.component.ts
```diff
@@ -0,0 +1,21 @@
...
+@Component({
+    selector: 'hc-modal-header',
+    template: `<div class="hc-modal-header">
+                   <ng-content></ng-content>
+                   <button *ngIf="showCloseButton" class="hc-modal-close-button" (click)="dismissModal()">&times;</button>
+               </div>`
+})
+export class ModalHeaderComponent {
+
+    @Input() showCloseButton: boolean = true;
```
andrew-frueh: I know we had a close button up in the header of some modals back in older prototypes I did - but I've stopped including that.  I think it's confusing to have two methods for dismissing a modal; best if the user always knows they have their save/cancel buttons at the bottom, and they have to pick one of those.  So I'd recommend we remove the showCloseButton option from this component.

lib/src/sass/_modals.scss
```diff
@@ -0,0 +1,59 @@
...
+$header-footer-height: 75px;
+$modal-border: 1px solid $gray-200;
+$body-height: 100%;
+
+.hc-modal-header {
+    padding: $modal-padding;
+    height: $header-footer-height;
+    border-bottom: $modal-border;
+    font-size: calculateRem(18px);
+    color: $offblack;
```
andrew-frueh: I'd suggest using the `$text` value for color instead (it points to `$offblack`) - here and on any other styles.  Just in case we ever want to change our text color, it'll be easier if we just have to change it in one place.

lib/src/sass/_modals.scss
```diff
@@ -0,0 +1,59 @@
+@import './colors';
+@import './mixins';
+
+$modal-padding: 20px;
+$header-footer-height: 75px;
+$modal-border: 1px solid $gray-200;
+$body-height: 100%;
+
+.hc-modal-header {
```
andrew-frueh: Can you add `font-weight: 600` to your .hc-modal-header style.  We want that header text to be semi-bold.

docs/app/components/modal/modal-demo.component.html
```diff
@@ -0,0 +1,252 @@
...
+            <hc-tile>
+                <h5>About</h5>
+                <p>A modal dialog is a window that forces the user to interact with it before they can go back to using the
+                    parent application.</p>
+            </hc-tile>
+
+            <hc-tile>
+                <h5>Example</h5>
+                <button hc-button color="primary" (click)="open()"> Open Modal from Component </button>
+                <button hc-button color="secondary" (click)="openTemplate(content)"> Open Modal from Template </button>
```
andrew-frueh: Can you add `class="demo-button"` to both of these buttons?  That just makes sure that if the buttons end up getting stacked on top of each other on a smaller screen, there is some vertical space between them.

lib/src/modal/modal-window.component.scss
```diff
@@ -0,0 +1,37 @@
...
+    display: block;
+}
+
+.hc-modal {
+    overflow: auto;
+}
+
+.hc-modal-sm {
+    width: 300px;
+    height: 200px;
```
jtrujill: I would remove the height from this all of the hc-modal sizes.

lib/src/modal/modal-header.component.ts
corytak: The close button and its options have been removed from ModalHeaderComponent

lib/src/sass/_modals.scss
corytak: Done

lib/src/sass/_modals.scss
corytak: Done

docs/app/components/modal/modal-demo.component.html
corytak: Done

lib/src/modal/modal-window.component.scss
corytak: This is done. Now the modal body will be at minimum `150px` and will max out 75px less than the view's height. Otherwise, it will expand to fill the content provided.

PR #163, opened 2018-03-03
Input Component & Select Component Update

### LINKED ISSUE #14: Input component
I'd like to volunteer for creating an input component. I'd like to use Angular Material's input as a basis, which can be found here https://material.angular.io/components/input/examples

My experience with using bootstrap for forms was that it was cumbersome with all of it's required DOM and classes, and I'd like to simplify that process to have a more consistent look. I really like how Material's input takes child components which represent the various error states that an input could be in and shows that state right below the input. This is an idea I'd like to pull into our component.

Another thing I like about Material's input is that the placeholder acts like a label. When the input doesn't have focus and no value, the placeholder acts like a regular placeholder; however, if it has focus or a value has been entered, that placeholder animates above the input. Now whether we do that animation or not I don't really care, but I would like for the component to handle having a label.

I don't know a whole lot about what the company style guide says or what everyone thinks about this idea, so lets get some communication going to see where we should go.

andrew-frueh:
Resolves #14. This was originally meant to only be a PR for a new Input text component, but as I got into it I realized that a significant number of our forms use both input text and select boxes.  So it's important our select component and our input components look and act very similarly.  Also, the select component was one of the first added to Cashmere, so it was missing some of our newer practices (ViewEncapsulation to none, not setting style elements directly, etc).

The markup for the Select component was a little odd too in hindsight.  An array of options was being passed to the component in the tag.  But that's pretty limiting because a dev can't set the select option values themselves, they were automatically being set to the array name.  So I've adjusted the markup to let the user set the `<option>` tags themselves.  And to make the input component, I've added the ability to add a label.

Last, the trickiest part of the input component was deciding how to do validation and warning messages.  HTML has a lot of ways to validate text fields directly, but I didn't really want to lock the component into any of them - because different apps will likely want to validate for different things.  It seems that most of our apps validate after a submit click, so I included a `valid` parameter on both the input and select components.  If set to false, it highlights the component and displays any appropriate warning messages.

### DIFF
--- a/docs/app/components/input/input-demo.component.html
+++ b/docs/app/components/input/input-demo.component.html
@@ -1,0 +1,90 @@
+<div class="demo-content">
+    <h1>Input Text Field</h1>
+    <h6>Last updated {{lastModified | date:'longDate'}}</h6>
+
+    <hc-tab-set direction="horizontal">
+    <hc-tab tabTitle="Overview">
+
+    <hc-tile>
+        <h5>About</h5>
+        <p>Text fields allow users to input, edit, and select text.
+            They typically reside inside forms, but can also be used in modal dialogs and subnavigation bars.  Input text fields are also used for
+            search and filtering of content.</p>
+    </hc-tile>
+
+    <hc-tile>
+        <h5>Example</h5>
+        <h4>Input with Validation</h4>
+        <form (ngSubmit)="onSubmit()">
+            <hc-input [valid]="validCheck">
+                <label hcInputRequired for="inputEmail">Validate an email address:</label>
+                <input required name="inputEmail" id="inputEmail" [(ngModel)]="emailText" #inputEmail="ngModel" width="250px"  />
+                <div hcInputErrors>
+                    <div *ngIf="errorVal===1">
+                        Email address is required
+                    </div>
+                    <div *ngIf="errorVal===2">
+                        Please enter a valid email address
+                    </div>
+                </div>
+            </hc-input>
+            <br>
+            <button hc-button title="Validate Input Text" color="primary" type="submit">Validate Input</button>
+        </form>
+
+        <h4>Input with Icon</h4>
+        <hc-input>
+            <input name="search" placeholder="search" width="250px" />
+            <hc-icon fontSet="fa" fontIcon="fa-search" class="search-icon"></hc-icon>
+        </hc-input>
+
+        <h4>Disabled Input</h4>
+        <hc-input>
+            <input disabled name="disabled" width="250px" />
+        </hc-input>
+
+    </hc-tile>
+
+    </hc-tab>
+    <hc-tab tabTitle="Documentation">
+
+    <hc-tile>
+            <h5>Angular Component</h5>
+            <pre>
+                <code>
+    &#x3C;hc-input valid=&#x22;true&#x22;&#x3E;
+        &#x3C;label hcInputRequired for=&#x22;inputEmail&#x22;&#x3E;Validate an email address:&#x3C;/label&#x3E;
+        &#x3C;input required &#x3E;
+        &#x3C;div hcInputErrors&#x3E;
+            Error messages are added here
+        &#x3C;/div&#x3E;
+    &#x3C;/hc-input&#x3E;
+                </code>
+            </pre>
+        </hc-tile>
+
+        <hc-tile>
+            <h5>Properties</h5>
+            <table class="api-table">
+                <tr>
+                    <th>valid</th>
+                    <td><span class="type-label">Type:</span> Boolean</td>
+                    <td>When true, highlights the border red and makes any items with the <code>hcInputErrors</code> directive visible</td>
+                </tr>
+            </table>
+        </hc-tile>
+
+        <hc-tile>
+            <h5>Directives</h5>
+            <p><strong>hcInputErrors</strong><br>
+                Should be added to an element or group of elements that contain error messages to be displayed when the component's <code>valid</code> flag is set to true.
+            </p>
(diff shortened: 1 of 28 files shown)


docs/app/components/input/input-demo.component.html
```diff
@@ -0,0 +1,53 @@
...
+        <p>Text fields allow users to input, edit, and select text.
+            They typically reside inside forms, but can also be used in modal dialogs and subnavigation bars.  Input text fields are also used for
+            search and filtering of content.</p>
+    </hc-tile>
+
+    <hc-tile>
+        <h5>Example</h5>
+        <h4>Input with Validation</h4>
+        <form (ngSubmit)="onSubmit()">
+            <hc-input [valid]="validCheck" required="true">
```
corytak: Since `hc-input` is tied into angular forms, we probably don't need to add our own valid input as it's given to us for free. https://angular.io/guide/form-validation

docs/app/components/select/select-demo.component.ts
```diff
@@ -3,15 +3,12 @@ import { Component, ViewEncapsulation } from '@angular/core';
...
     templateUrl: './select-demo.component.html',
-    styleUrls: [
-        './select-demo.component.scss'
-    ]
+    styleUrls: ['./select-demo.component.scss']
 })
+
 export class SelectDemoComponent {
-    showTemplate: boolean = true;
+    validCheck: boolean = true;
```
corytak: I would remove `validCheck` and `toggleValidate()` and rely on angular forms functionality instead.

lib/src/input/input.component.ts
```diff
@@ -0,0 +1,25 @@
...
+    _valid: boolean = true;
+    _required: boolean = false;
+
+    constructor() {}
+
+    @Input() get valid(): boolean { return this._valid; }
+
+    set valid(validVal) { this._valid = anyToBoolean(validVal); }
+
+    @Input() get required(): boolean { return this._required; }
```
corytak: Just like with `valid`, `required` is a Validator available by default from angular forms. it might be nice to hook into that native functionality.

lib/src/select/select.component.ts
```diff
@@ -13,43 +15,45 @@ import { ControlValueAccessor, NG_VALUE_ACCESSOR } from '@angular/forms';
...
+
+    @Input() get valid(): boolean { return this._valid; }
+
+    @Input() get disabled(): boolean { return this._disabled; }
+
+    @Input() get required(): boolean { return this._required; }
+
+    set required(isRequired) { this._required = anyToBoolean(isRequired); }
+
+    set disabled(isDisabled) { this._disabled = anyToBoolean(isDisabled); }
```
corytak: You may also want to look into reactive/template based forms for setting disabled.

docs/app/components/input/input-demo.component.html
andrew-frueh: Good suggestion Cory - just recapping what we discussed on slack here.  Instead of hc-input being a separate component, I'll adjust so that it actually extends the base input component to `input [hc-input]` - to take advantage of all the built in form validation.  Then in addition I'll create `label [hc-label]` - which is better because it can be used by both select and input that way.  And probably keep hcErrors as a directive that can be used by any form component.  I'll also take a look and see if it would be possible to be consistent with our select component and switch it to `select [hc-select]`.

andrew-frueh:
@corytak - I've updated the `hc-input` component to be an attribute selector on `<input>` as you suggested, and did the same with `hc-label` on the `<label>` element.  I didn't add another vertical tab for label because it's more of a utility component.  But I did call it out in the documentation on both the Input and Select pages and put it into the example code.

I kept a `required` property on `hc-label` because as best as I can tell, the native label element doesn't have a required value.  So that flag is there to put a red asterisk on the label if set to true.

Also, I kept a `highlight` property on the label so that the dev can decide when they want to put a red highlight around the element.  I could set it to automatically come on with `:invalid` on the css, but it seems like different instances might require validation at different times, so I figured this would be more flexible.  But let me know if I'm overthinking that.

Also updated the Select component's documentation to be markdown on this push as well.

docs/app/components/select/select-demo.component.html
```diff
@@ -15,68 +15,41 @@ <h5>About</h5>
         <h5>Examples</h5>
         <h4>Standard Select Component</h4>
         <div class="select-sample">
-            <hc-select placeholder="Select a city:"
-                    [options]="['Philadelphia', 'Atlanta', 'Salt Lake City', 'Chicago', 'Orlando']"></hc-select>
+            <label hc-label>Facility Location:</label>
+            <hc-select placeholder="Select a city:">
+                <option>Philadelphia</option>
```
corykon: This is a great API change 👍

lib/src/input/input.component.scss
```diff
@@ -0,0 +1,30 @@
+@import '../sass/colors';
+@import '../sass/functions.scss';
+@import '../sass/mixins.scss';
+
+.hc-input {
+    color: $text;
```
corykon: Future PR maybe, just noting it while it's on my mind... would be sweet if there were a "form-controls" sass file containing things that are common across controls... like the styles for borders (default vs error state), font size, padding, etc. This could be a set of variables that each component pulls in, or perhaps standalone sass file that people could use even if they for some reason can't use the angular components but won't style guide approved form controls.

andrew-frueh: That's a great idea.  I had to add the error messages css to the typography css file, but it didn't really seem like it belonged there.  So a forms.scss would be really useful.  If you remember, can you add that as an issue once this PR gets merged?

PR #164, opened 2018-03-06
Table Component Scaffolding

corytak:
This PR includes the beginnings of a table component. I thought since this was a larger feature we may be able to work independently and collaborate on a table branch in the main fork of the repository. 

At this point I have the following working (although needing refinement):

- Sorting
- Paging Client side (toggle between more results button & numbered pagination)

What still needs work:

 - Input Table Options (total number of pages (server side pagination) rows per page, server/client side pagination, checkboxes/selectable rows)
 - Table Output events (current page, selected rows, sortEvents for server side pagination)
 - Bake in checkboxes (input row type extend some `IsCheckable` class that has an `isChecked` property?)
 - Documentation
 - Tests
 - Toggle to choose page size


benjanderson:
Great job @corytak this is a great starting point to something we all need.

Quick question, Why not call it a `grid` or `data table` component?

corytak:
@benjanderson Either of those work for me. I can change it to `hc-grid`?

benjanderson:
I would also ping @kpaul to see if this works for his needs as well

joeskeen:
Just saw this and had one thought @corytak @andrew-frueh @benjanderson : 
Would it make sense to have the pagination component be separate from the grid component?  I can see some use cases (such as search results page) where the results would be paged, but not necessarily be in a data grid.  This component could be completed out-of-band (ahead of) with the grid changes as well.

PR #172, opened 2018-03-09
Grid Updates

corytak:
This includes the ability to select rows (either the entire row using existing styling or with checkboxes). Still to do: emit serve- side pagination items. 

Unfortunately, this includes the latest from master so it's difficult to discern between master changes and grid changes.


lib/src/checkbox/checkbox.component.scss
```diff
@@ -33,6 +33,36 @@
         }
     }
 
+    .hc-checkbox-indeterminate {
```
corytak: Added indeterminate overlay here

docs/app/components/grid/grid-demo.component.html
```diff
@@ -12,16 +12,17 @@ <h5>About</h5>
...
                         <tr>
+                            <th hc-master-checkbox></th>
                             <th hc-sortable="heading1">Heading 1</th>
                             <th hc-sortable="heading2">Heading 2</th>
                             <th hc-sortable="heading3">Heading 3</th>
                         </tr>
                     </thead>
                     <tbody>
-                        <tr *ngFor="let row of tableExample.rows">
+                        <tr *ngFor="let row of tableExample.rows" [hc-selectable]="row" [showCheckbox]="true">
```
corytak: I like that this interface is flexible, but it also feels a bit complex. I would love feedback about this.

docs/app/components/grid/grid-demo.component.html
```diff
@@ -12,16 +12,17 @@ <h5>About</h5>
 
             <hc-tile>
                 <h5>Example </h5>
-                <table [hc-grid]="rows" [gridOptions]="gridOptions" #tableExample>
+                <table [hc-grid]="rows" [gridOptions]="gridOptions" class="hc-table-borders" #tableExample>
                     <thead>
                         <tr>
+                            <th hc-master-checkbox></th>
```
corytak: Started out with wanting to include whether the rows should be selectable in the gridOptions, but wasn't sure of the best way to add the checkboxes in that instance or have access to which rows were selected.

benjanderson:
I noticed you are merging into `HealthCatalyst:table` did you mean to merge into master?

benjanderson:
I noticed there is a bug with the select all

lib/src/checkbox/checkbox.component.scss
benjanderson: I really like it! Took me a google to figure out what indeterminate meant

lib/src/grid/master-checkbox.component.ts
```diff
@@ -0,0 +1,45 @@
+/* tslint:disable:component-selector */
+/* tslint:disable:no-input-rename */
+/* tslint:disable:use-host-property-decorator */
+// https://github.com/mgechev/codelyzer/issues/178#issuecomment-265154480
+import { Component, HostBinding, HostListener, Input, Output, EventEmitter } from '@angular/core';
+
+@Component({
+    selector: 'th[hc-master-checkbox]',
```
benjanderson: I don't love the name `hc-master-checkbox` it could possibly be `hc-grid-checkbox`?

benjanderson: also, what is the value add to having a "grid-checkbox" over just using `hc-checkbox` inside the thead>tr>th? I feel removing this component simplifies things quite a bit

lib/src/checkbox/checkbox.component.scss
benjanderson: Wow, this looks very nice! Great job

docs/app/components/grid/grid-demo.component.html
benjanderson: Just a styling correction, on page 53 of the style guide it has an example of a 'selected' row of a table. That is not happening what a row is "selected". https://spark.healthcatalyst.com/docs/DOC-3141

docs/app/components/grid/grid-demo.component.html
benjanderson: This is a tough problem as I don't think their is an easy answer. We could get together about some of the options...

lib/src/grid/grid.component.ts
```diff
@@ -70,13 +81,42 @@ export class GridComponent implements OnChanges, AfterContentInit {
...
+        if (this.selectedRows.length > 0 && this.selectedRows.length < this.fullDataSet.length) {
+            this.masterCheckbox.setIndeterminateState();
+        } else if (this.selectedRows.length === 0) {
+            this.masterCheckbox.setUncheckedState();
+        } else {
+            this.masterCheckbox.setCheckedState();
+        }
+    }
+
     private sort(sortEvent: SortEvent) {
```
benjanderson: many of these data manipulation/retrieving functions should be separated into a service. So they could be replaced with a call to the server. I would also consider making them work properly with observables.

docs/app/components/grid/grid-demo.component.html
corytak: Good catch on style. My thought process was you can have a row be selectable and if no checkboxes were present it would look like the selected style on the styles tab now (https://cashmere.azurewebsites.net/styles/table), but if checkboxes were present it would just update the checkbox. But perhaps we do the lighter blue highlight when checkboxes are present and use the darker blue one when no checkbox is present?

lib/src/grid/master-checkbox.component.ts
corytak: Agreed about the name.

corytak: let's talk options for `hc-grid-checkbox`. We could create a guide, we could do that work in a service instead.

lib/src/grid/selectable.component.ts
```diff
@@ -0,0 +1,49 @@
...
+    @HostBinding('class.hc-row-selected') public selected = false;
+    @Input('hc-selectable') row: any;
+    @Input() showCheckbox: boolean = false;
+    @Output() selectEvent: EventEmitter<SelectEvent> = new EventEmitter<SelectEvent>();
+
+    ngOnChanges() {
+        this.updateStyles();
+    }
+
+    @HostListener('click', ['$event'])
```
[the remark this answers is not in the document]
corytak: This is for when the checkbox is turned off. When it's turned off you can get the same select row abilities by clicking anywhere on the `<tr>`

lib/src/grid/grid.component.ts
corytak: @benjanderson please do! That's why I created the separate table branch on the main fork so that other people could jump in.

waldnercharles: @benjanderson I think I'd prefer a sort output and a paging output to a service. I think that calls to  the server are client specific and wouldn't want to be presumptuous about how data should be retrieved. Thoughts?

lib/src/grid/selectable.component.ts
```diff
@@ -0,0 +1,49 @@
...
+    onTableRowClick(event: any) {
+        if (!this.showCheckbox) {
+            this.selectRow();
+        }
+    }
+
+    public selectRow() {
```
waldnercharles: This is minor, but I think I'd prefer the name `toggleRow` or `toggleRowSelection`. `selectRow` to me means that the row will always end up highlighted.

corytak:
After talking with @benjanderson We're merging this in with the knowledge that there are still some tasks to be done:

1.  Change label to `display: none` for checkbox when there is no text ( you can currently click to the side of a checkbox to check it)
2. Rename hc-master-checkbox to hc-grid-checkbox
3. Evaluate GridService v EventEmitters decision
4. Add option for client side or server side pagination in table
5. rename `selectRow` to `toggleRowSelection`
6. evaluate feasibility of omitting hc-grid-checkbox and including a grid checkbox based on gridOptions passed in

PR #177, opened 2018-03-13
Error Pages Styles

### LINKED ISSUE #176: 404 & Other Error Pages
We've had a lot of requests from our various app teams on how to handle 404 or other error pages.  I'm thinking this would probably be a component because it has some specific formatting and would simplify displaying the error for developers.  But if a guide makes more sense that's fine too.  The image below shows the basic structure that can be extended to whatever the particular error is.

andrew-frueh:
Resolves #176.  Originally I was planning on this being a guide, but since there is a whole style sheet associated with it, it seemed more appropriate to put it into the Styles section.  I'm not sure why GitHub is saying the whole markdown.directive.ts file changed, the only change made was to line 18 that @benjanderson and I discussed.


guides/styles/error-pages.md
```diff
@@ -0,0 +1,70 @@
...
+    <div class="content-column">
+        <div class="error-header-code">404</div>
+        <div class="error-message">
+            Sorry, we can't find the page you're looking for!
+        </div>
+    </div>
+    <hr>
+    <div style="display: flex"><div class="error-button">Go to Homepage</div></div>
+</div>
+<br><br><br>
```
benjanderson: This is an indication that we may need to add some padding to the global style instead of having so many line breaks `<br>`

docs/app/styles/styles-routes.ts
```diff
@@ -36,6 +37,11 @@ export const routes: Routes = [
                 component: CodeDemoComponent,
                 data: { title: 'Code' }
             },
+            {
```
benjanderson: The components are alphabetized. Should the styles be too?

guides/styles/error-pages.md
andrew-frueh: I think we're probably okay for now because the padding between the normal `hc-tiles` are fine - it's just when I start inserting raw html that it became an issue.  And I didn't want to add extra padding to the error-tile class because that might make for formatting issues when it's used in apps.

docs/app/styles/styles-routes.ts
andrew-frueh: At first I was thinking we'd manually order them, but now that you mention it, we have enough in there that we probably should alphabetize them.  I'll make that change.

andrew-frueh:
That's an excellent idea @corykon.  The only issue I see with embedding the image in the stylesheet is that it would be more of a pain to swap it out.  @benjanderson and I had talked about included an `assets` folder in lib for something like this.  I would have done that, but I was struggling to figure out what the `src` path would need to be for an svg in lib/assets from the error-pages.md markdown file.  Maybe if I put it in the css as `content:url` instead that will be easier to reference?  

What do you think about an assets folder in lib for the svg?  And what would be the path to get to it?

corykon:
Assets folder could work. The trick is that the implementer will need to know to include the library assets as part of their build process. For us, that would mean going into angular-cli.json and adding to the "assets" array, so that webpack knows to include it:
```
"assets": [
                "assets",
                {"glob": "**/*", "input": "../lib/assets", "output": "./assets/"},
                "favicon.ico",
                "web.config"
            ],
```

After that, our path would be something like ``.

All of the above could look different for each implementer, depending on what type of build system they use, but the devs should hopefully know what to do, as long as we mention that they need to manually include the image assets to make use of them.

The encoding technique does add an extra step, but it's really quick... makes life a little easier for the implementer. We may run into other places where an asset folder makes more sense though, so maybe just go that route now?

andrew-frueh:
Great point @corykon about some of the additional challenges with including an assets folder in lib.  I think your suggestion to encode the image in this case is right way to go.  I just had to give a little more thought to how to handle the responsive sizing with a `background-image`, but I think it's all good.  Let me know how the updated version looks to you.

corykon:
@andrew-frueh I'm seeing a border around the image now (Chrome on windows):
image

If you use div tag instead of an img tag, that should fix it.

Issue #144, opened 2018-02-22
Add bullets to typography guidelines


cedar-ave:
This would be helpful for web documentation. This is an example of a use case for three levels of bullets. In current state it doesn't look great.

![2018-02-22 16_31_25-set up the workspace _ articles](https://user-images.githubusercontent.com/15255009/36570264-7cdf03d8-17ee-11e8-8ba5-e3591b387ace.png)

andrew-frueh:
Great suggestion @cedar-ave, I hadn't even thought of this before.  For something like this, I'd trust your recommendation more than I would my own.  Do you have a suggestion for the best way to present nested bullets?  Particularly in a way that would work best for the documentation you're doing.  And if so, maybe add it as an additional tile on [this page](https://github.com/HealthCatalyst/Fabric.Cashmere/blob/master/docs/app/styles/typography/typography-demo.component.html). 

Then point me to the branch of your fork with the change.  Once I see what you've got, I can help you out translating your recommendation into matching styles for our _typography.scss file.  And then we submit all that as one pull request.  Sound good?

cedar-ave:
Sounds good, thanks. I'll give it a shot. After looking around briefly, not many living style guides have typographical list components, especially with nested lists. But this one seems like a potential one for us  to model after: [http://carbondesignsystem.com/components/list/code](http://carbondesignsystem.com/components/list/code). I'll pursue something like that unless anyone feels differently.
Ultimately I like this look (with potentially another `–` or a `>` for third level)
![2018-02-28 09_51_38-cap install guide pdf](https://user-images.githubusercontent.com/15255009/36800629-3320c5fa-1c6d-11e8-8f37-494ebe666f7d.png)
(this is from my PDF documentation)
Also, like you did on the help modal thread, I'm going to tip off Knowledge Dev on this discussion - their two cents on decisions like this that affect web documentation would be valuable.

cedar-ave:
FYI @aflanders

aflanders:
@cedar-ave The HC Word template uses a smaller dash for the third level of bullet.
<img width="171" alt="bullets" src="https://user-images.githubusercontent.com/36929945/36802889-5a5aa676-1c73-11e8-8d68-2f4bf15ca1b1.PNG">
https://spark.healthcatalyst.com/docs/DOC-4201

SherriVance:
I believe the first level of secondary bullet is em dash, second is en dash

cedar-ave:
[...]

PR #179, opened 2018-03-13
add lists to typography guidelines

cedar-ave:
To close [#144](https://github.com/HealthCatalyst/Fabric.Cashmere/issues/144).
I formatted one of the files in VS Code without realizing it would show up as a bunch of changes - sorry.


cedar-ave:
Good point - that would work. The thinking behind the separation is that the font size and padding in tables is smaller and tighter than in body text. At least, I think it looks a lot better and is more readable that way. We have a ton of tables in the docs and some of the cells can get hefty. I noticed Microsoft does the same thing (see body text at the top, then the table beneath)

2018-03-13 16_50_27-understanding azure virtual machine usage _ microsoft docs

In the DOS docs site, these are our regular bullets

2018-03-13 16_46_33-protect entities and bindings from unauthorized schema changes _ articles - mic

These are our table bullets

2018-03-13 16_46_19-protect entities and bindings from unauthorized schema changes _ articles - mic

But I see your point and don't want to clutter up the style guide for no reason.

andrew-frueh:
Makes sense @cedar-ave - and I definitely like the different styles for article lists in tables as opposed to paragraphs.  Maybe for the sake of the demo tile, we just show an example of the tile version.  But we add a sentence to the top of the tile that describes the differences in font-size and padding for lists in tables vs paragraphs.

cedar-ave:
@andrew-frueh I love this solution. I did this and also explained the `<article>` tag.

cedar-ave:
I forgot to add the comments about tables being more compact - will do that now

PR #181, opened 2018-03-16
Documentation cleanup

benjanderson:
added quite a few guides and cleaned up the existing ones


guides/getting-started.md
```diff
@@ -1,5 +1,5 @@
 # Getting Started
-###### Last updated February 20, 2018
+###### Last updated March 10, 2018
```
joeskeen: Do we really need a "last updated"?  The Git commit history for the file should be sufficient, right?
I have seen many times in the past that people tend to forget to change the "last updated" later on down the road, making it less useful.

joeskeen:
Thanks @benjanderson for doing this.  This PR shows a lot of work on your part and I think it will be a great improvement for both consumers and contributors.

guides/getting-started.md
```diff
@@ -9,22 +9,22 @@ The library uses Angular 5, if you have a previous version of angular see [the u
...
+##### Setting up project
 #### Step 1. Install npm module
 
-```BASH
+``` BASH
 npm i --save @healthcatalyst/cashmere
 ```
 
-#### Install peer dependencies (optional)
+#### Install peer dependencies on [popper.js](https://popper.js.org/), [font awesome](https://fontawesome.com), and [open sans](https://fonts.google.com/specimen/Open+Sans) (optional)
```
cedar-ave: Not sure if "optional" applies to the whole step or just to Open Sans
If if applies to the whole step, maybe put `(Optional)` at the beginning before `Install peer dependencies...`. This is how I changed it on Step 4 (commit coming soon).

guides/contribution-guide.md
```diff
@@ -1,13 +1,25 @@
 # Guidelines for Contribution
-###### Last updated February 20, 2018
+###### Last updated March 10, 2018
 
 :::
-In order for a contribution to be considered, the following conditions must be met:
+##### All may contribute
+We would love for you to contribute to Cashmere and help make it ever better! As a contributor, you will be part of a community making this project better. To get started, just follow the [getting started guide](http://cashmere.healthcatalyst.net/guides/getting-started)
```
cedar-ave: Not sure what is happening, but when I click the Getting Started guide link in the ng serve build, it goes to the "right" link but the content is off.
2018-03-17 09_40_07-cashmere
 Maybe that is something wrong on my end.

guides/submit-an-issue.md
```diff
@@ -0,0 +1,30 @@
...
+###### Last updated March 15, 2018
+
+:::
+#####  Found a Bug?
+
+Before you [submit an issue](https://github.com/HealthCatalyst/Fabric.Cashmere/issues), search the archive, maybe your question was already answered.
+
+If your issue appears to be a bug, and hasn't been reported, open a new issue. Help us to maximize the effort we can spend fixing issues and adding new features by not reporting duplicate issues. Providing the following information will increase the chances of your issue being dealt with quickly:
+
+- **Overview of the Issue** - if an error is being thrown a non-minified stack trace helps
```
cedar-ave: It might be nice visually to have the 2nd-line indents match the 1st-line indents in ul > li. I tried but my CSS is not up to par.

guides/getting-started.md
```diff
@@ -41,9 +41,35 @@ export class SomeModule { }
...
+Add a reference to font awesome ([official instructions](https://github.com/angular/angular-cli/blob/master/docs/documentation/stories/include-font-awesome.md)) and open sans by adding a reference in `src/styles.scss`
+
+``` scss
+$fa-font-path: '../node_modules/font-awesome/fonts';
+@import '../node_modules/font-awesome/scss/font-awesome';
+
+$FontPathOpenSans: "../node_modules/npm-font-open-sans/fonts";
+@import '../node_modules/npm-font-open-sans/open-sans';
+```
+:::
```
cedar-ave: When I first set up to build Cashmere locally I had to do these additional steps for it to work -
Install angular-devkit/core
npm install d3 --save
Not sure if these are of value to other users or if this is even the right place for them but FYI.

guides/contribution-guide.md
```diff
@@ -1,13 +1,25 @@
 # Guidelines for Contribution
```
cedar-ave: You will see in my commit that I tried to standardize this title in a couple places. FYI as to why I made those changes.

guides/getting-started.md
benjanderson: @andrew-frueh introduced it, so I would create an issue and discuss it with him

guides/contribution-guide.md
benjanderson: awesome thank you so much 👍

guides/getting-started.md
benjanderson: This needs to be added to the demo package.json, I will do that here

benjanderson: Its not a requirement to use the library (as opposed to the demo app), so I think the getting started is fine the way it is

guides/getting-started.md
joeskeen: @andrew-frueh just for future reference, we could always add this to the MD file during the build before it is published using a command like this:

```bash
git log -1 --format="%ad" --date=short -- guides/getting-started.md
```

```
2018-03-03
```

This way we could not have to worry about updating the date or making sure the date isn't out of date.  We should probably make a separate issue of this if you are open to the idea.

PR #182, opened 2018-03-17
Include all SASS stylesheets in the package so they can be used in consuming applications

joeskeen:
I was trying to add Fabric.Cashmere to the Fabric.Authorization.AccessControl demo project.  When trying to import the Cashmere SASS styles, it caused an error.  Looking into it, I realized that the cashmere.sass file was just importing other SASS files, which were not included in the package when built for NPM.  This fixes the issue, copying all SASS styles into the NPM package directory when the package is built.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 /dist
 /tmp
 /out-tsc
+.ng_pkg_build
 
 # dependencies
 /node_modules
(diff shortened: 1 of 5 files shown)


benjanderson:
I am working on getting the npm package to work as well and this is something I need, so if you don't mind squashing and merging it

PR #185, opened 2018-03-20
Library package cleanup

benjanderson:
- removed `index.ts` barrel files (many of them were wrong or out of date)
- include many components in library and fix linting errors
- add brand icon as @input of navbar
- clarify in getting started


lib/src/navbar/navbar.component.html
```diff
@@ -1,7 +1,7 @@
 <nav class="navbar" [ngClass]="{'fixed-top': fixedTop}">
     <div #defaultbrand class="navbar-brand" *ngIf="appswitcherbrand.children.length == 0">
         <a [routerLink]="homeUri" class="brand no-switcher">
-            <img src="./assets/TriFlame.svg">
+            <img src="{{brandIcon}}">
```
joeskeen: 👍

lib/src/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -33,17 +34,19 @@ export class BreadcrumbsComponent implements OnInit, OnDestroy {
...
-            this.breadcrumbs = this.getBreadcrumbs(root);
-            if ( this.breadcrumbs.length > 1 ) {
-                this.backURL = this.breadcrumbs[ this.breadcrumbs.length - 2 ].url;
-                this.backShow = 'inline';
-                this.locationLabel = '';
-            } else {
-                this.backShow = 'none';
-                this.locationLabel = this.breadcrumbs[ this.breadcrumbs.length - 1 ].label;
+        this.routerSubscription = this.router.events.subscribe((event: Event) => {
+            if (event instanceof NavigationEnd) {
```
joeskeen: Just curious the advantage of putting an `if` statement around the entire function body as opposed to using `filter`.

benjanderson: This was being a pain. WIth rollup I wasn't able to get `filter` to be recognized and the build would fail. This is just the hacky way I could get the build to succeed.

You may ask why was it passing before? Because breadcrumbs weren't included in the `public_api.ts` export and rollup wouldn't transpile them

PR #186, opened 2018-03-20
Set `z-index` of popover so it pops... over.

joeskeen:
When integrating the current Cashmere as an NPM package into Fabric.Auth.AccessControl, I found that the popover likes to pop under:

image

This solved the issue.

### DIFF
--- a/lib/src/drawer/drawer.component.scss
+++ b/lib/src/drawer/drawer.component.scss
@@ -1,3 +1,5 @@
+@import '../sass/_variables.scss';
+
 $drawer-background-color: #6E6F71;
 $drawer-font-color: #b9babb;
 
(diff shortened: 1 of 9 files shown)


benjanderson:
First off, this is awesome that you found a bug and you fixed it with a pull request. Thank you for being an owner. 

Looking at the search results here 
https://github.com/HealthCatalyst/Fabric.Cashmere/search?utf8=%E2%9C%93&q=z-index&type=

It looks like we are getting z-indexes spread across our project. Its just a little more work but I would love to store all of them in the `_variables.scss` file so we can better have a hierarchy of what should show first. 

Here is an example of what bootstrap did, I wouldn't mind if we used it as a template: https://getbootstrap.com/docs/4.0/layout/overview/#z-index

joeskeen:
Love the idea.  I'm on it.

joeskeen:
@benjanderson I've put all the z-indexes into the _variables.scss file.  I based the ideas off of Bootstrap as suggested, but tailored it to our needs.  For example, since the modal service increments the z-index for subsequent modals, I reserved a range (2000-2999) for that and set other things higher or lower than that.

PR #190, opened 2018-04-04
Fix Prod/AOT builds by making exports explicit; upgrade libraries

joeskeen:
We were having some awful issues building Fabric.Authorization.AccessControl with the `--prod` flag after integrating Cashmere into it.  We would get this error message:

```
Unexpected value 'undefined' exported by the module 'ButtonModule'
```

Eventually, stumbling upon [the advice of @deebloo](https://github.com/angular/angular-cli/issues/1831#issuecomment-242953578): 

> barrels can be tricky. all I know is that exporting using a wildcard sometimes causes issues.
> ¯_(ツ)_/¯

I decided to replace all `export *` occurrences with explicit exports.  This fixed the issue.

It turns out that nested `export *`s don't get resolved properly, as `export * from './public-api'` in healthcatalyst-cashmere.d.ts (a generated file) tried to resolve `export * from './src/button/button.module'` but got `undefined`, thus causing the error.  Explicit exports do not have this problem.

Along the way, I ended up upgrading some packages, which led me to better error messages to resolve the issue.


tslint.json
```diff
@@ -128,9 +128,6 @@
         "use-life-cycle-interface": true,
         "use-pipe-transform-interface": true,
         "component-class-suffix": true,
-        "directive-class-suffix": true,
-        "no-access-missing-member": true,
```
joeskeen: The newest version of TSLint/Codelyzer removed these three rules.  My guess is they are obsolete, as the TS compiler has improved its checks.

lib/public_api.ts
```diff
@@ -1,22 +1,22 @@
...
-export * from './src/progress-indicators/progress-indicators.module';
-export * from './src/radio-button/radio-button.module';
-export * from './src/select/select.module';
-export * from './src/subnav/subnav.module';
-export * from './src/tabs/tabs.module';
-export * from './src/tile/tile.module';
-export * from './src/typeform-survey/typeform-survey.module';
+export { AccordionModule } from './src/accordion/accordion.module';
+export { BreadcrumbsModule } from './src/breadcrumbs/breadcrumbs.module';
+export { AppSwitcherModule } from './src/app-switcher/app-switcher.module';
```
benjanderson: This exported the `app-switcher.service.ts` and the `app-switcher-interfaces.ts` which wasn't included in the module.  I am not sure they will be exported, can you verify this or include them to be excluded somehow?

joeskeen: Good point.  I'll adjust this.

drogars:
So just to be clear if we need to fix this before the next release, we need to remove `export *` from any TS file in Cashmere that is causing an issue and replace it explicit exports?

drogars:
I am not sure, but we are getting this error and we tried to get the latest version of Cashmere and it's still happening.  I was hoping you all might know what is going on.

PR #196, opened 2018-04-08
Collapse Navbar Links

### LINKED ISSUE #169: Collapse Navbar Links
Didn't notice this until we added the Guides section, but it is possible for a navbar icon pinned to the right to cover the Guides link before the window hits the first responsive breakpoint:

navbarlinks

I think this is an issue that many apps will face, especially those with long section names or a lot of sections.   But I'm not entirely sure what the best way to handle this is.  Should the horizontal padding condense as they run out of space?  Should we introduce an "Other" link at the end if we run out of space that collects the overflow items into a popover?

@corykon @corytak @benjanderson @iamboo  - have you seen this done well anywhere that we could model our navbar after?

andrew-frueh:
Resolves #169.  Before the navbar reaches it's first responsive breakpoint, it now looks to see if all the links can fit on the screen.  If not, it first reduces the padding on the logo, and then it starts collecting the links into a "More" popover.  Based on the link @corykon provided in the issue, this seems to be a best practice for what to do with navbars with more elements than can fit on the screen.


lib/src/navbar/navbar.component.ts
```diff
@@ -1,55 +1,126 @@
...
+    menuOpen: boolean = false;
+    linkWidths: Array<number> = [];
+    linksMax: number = 0;
+    collapse: boolean = false;
+    logoCondense: boolean = false;
+    logoWidth: number = 0;
+    moreList: Array<Object> = [];
+
+    @HostListener('window:resize') navResize() {
+        this.popper.hide();
```
corykon: This is some awesome & complex logic. Couple thoughts on it that don't necessarily need to be addressed now:

- Most components don't have a whole lot of real unit tests, but this type of code is prime for testing. Maybe not now, but at some point it would be good to have some solid tests here.

- This function might be a little more digestible if broken up into smaller functions. Func names nicely document/explain what's going on, and are more easily unit testable. I could see there being at least a couple private functions that navResize() calls... maybe `condenseLogoIfNeeded()` and `collapseNavLinksIfNeeded()`. Also looks like lines 54-56 and 60-62 are almost identical, could be a simple private function there, too.

lib/src/navbar/navbar-link/navbar-link.component.ts
```diff
@@ -1,20 +1,22 @@
...
     @Input() exact: boolean = false;
+    hidden: boolean = false;
 
-    constructor() {
+    constructor( private el: ElementRef, private ref: ChangeDetectorRef ) {
     }
 
-    ngOnInit() {
-    }
+    hide() { this.hidden = true; this.ref.detectChanges(); }
```
corykon: Just curious why detectChanges() is needed here. Does it force the hide/show to happen faster so that the calculations in navResize() are accurate?

docs/app/components/navbar/navbar-demo.component.html
```diff
@@ -12,10 +12,15 @@ <h5>About</h5>
 
     <hc-tile>
         <h5>Example</h5>
-        <hc-navbar appIcon="./assets/CashmereAppLogo.svg" brandIcon="./assets/TriFlame.svg" user="Christine K." [homeUri]="undefined" [fixedTop]="false">
+        <hc-navbar appIcon="./assets/CashmereAppLogo.svg" brandIcon="./assets/TriFlame.svg" [homeUri]="undefined" [fixedTop]="false">
             <hc-app-switcher></hc-app-switcher>
-            <hc-navbar-link [active]="true" [uri]="undefined">Home</hc-navbar-link>
-            <hc-navbar-link [active]="false" [uri]="undefined">Buttons</hc-navbar-link>
+            <hc-navbar-link [active]="true" uri="undefined" linkText="Home"></hc-navbar-link>
```
benjanderson: It is nice when the example in the markdown matches the demo. Could you update it with these changes?

docs/app/components/navbar/navbar-demo.component.html
```diff
@@ -40,6 +45,17 @@ <h5>Example</h5>
         </hc-popover-content>
     </hc-tile>
 
+    <hc-tile>
+        <h5>Responsive Behavior</h5>
```
benjanderson: I am a little confused on what belongs in the overview tab and what belongs in the documentation. Just out of curiosity, why would this be here instead of documentation?

docs/app/components/navbar/navbar-demo.component.html
```diff
@@ -12,10 +12,15 @@ <h5>About</h5>
 
     <hc-tile>
         <h5>Example</h5>
-        <hc-navbar appIcon="./assets/CashmereAppLogo.svg" brandIcon="./assets/TriFlame.svg" user="Christine K." [homeUri]="undefined" [fixedTop]="false">
+        <hc-navbar appIcon="./assets/CashmereAppLogo.svg" brandIcon="./assets/TriFlame.svg" [homeUri]="undefined" [fixedTop]="false">
```
benjanderson: I like the way we show off the responsive navbar but the "user" was removed, what if we had a second example navbar with the user?

lib/src/navbar/navbar.component.scss
```diff
@@ -63,6 +63,8 @@ $navbar-fixed-shadow: 0px 2px 6px $shadow;
         display: flex;
         justify-content: center;
         align-items: center;
+
+        &.logo-condense { padding-right: 25px !important; }
```
benjanderson: Can this be removed? I didn't see any difference when removing the `!important`. I try to avoid it when possible

lib/src/navbar/navbar.component.scss
```diff
@@ -176,3 +221,10 @@ $navbar-fixed-shadow: 0px 2px 6px $shadow;
         .navbar-app { padding: 0 0 0 5%; }
     }
 }
+
+.more-popover a {
+    &.active {
+        color: $text;
```
benjanderson: The active link being set to black looks strange, did you mean to do this?

image

lib/src/navbar/navbar.component.ts
benjanderson: Wow, this is awesome. This was a lot of work

docs/app/components/navbar/navbar-demo.component.html
andrew-frueh: Thanks for the review @benjanderson and sorry it's taken me so long to get back to it.  As @jtrujill has already brought up in a recent issue, this question comes back to a fundamental question about who the demo site is for.  The reason I included it over here was because I was thinking product managers who might not want to click into the documentation, but would still have an interest in understand its responsive behavior.  So I was thinking of it more as an overview.  I think coming to a consensus on our target audience will be one of the first jobs of the steering committee, and that should give us some guidance around what content to include and where.  So I'll probably leave this tile as is until we reach that point, and then we can revisit this and others.

docs/app/components/navbar/navbar-demo.component.html
andrew-frueh: I actually only removed the user because there was so little space in the demo tile and everything was getting squeezed.  But I agree it's good to have it in there.  So I'll just put it back in.

andrew-frueh:
Thanks again for the feedback @benjanderson.  I made the following changes:
- Updated the documentation code to match the demo component
- Added the user back into the navbar demo (which was good because I realized there were some styles that needed adjusting around that)
- Removed the important tag from the logo-collapse style

It sounds like we need to do some more testing on the responsive behavior, but if we could do those fixes as a separate PR that would be great so this one doesn't sit around too long.  Let me know if that's okay with you.

benjanderson:
Sounds good, can you create those issues after you complete the pr. Make sure to test it in firefox

andrew-frueh:
Will do @benjanderson - I do all my development on Firefox on the Mac, so I can verify it works fine there.  Must be something to do with the Windows version of Firefox.  I'll take a look at that.

PR #209, opened 2018-04-16
Picklist

### LINKED ISSUE #61: Picklist

corykon:
Huge PR. Let me know if anyone wants to sit down and do a walkthrough with me.

Constructed it so that it'll work for PopBuilder, and hopefully for others, too. Supports cases as simple as just passing an array of strings to choose from, to passing in callback functions that handle paged server requests across values and value sets.

Closes #61


joeskeen:
@corykon could we set up a time tomorrow to go over this?  I'll take a look at it tonight.

andrew-frueh:
Great idea @joeskeen - just sent out an invite for tomorrow.

joeskeen:
I'm going to take some time in the next few days to give this a close look.  I should have a review by the end of the week.

jtrujill:
You'll need to add an index file that follows all the other components

Issue #211, opened 2018-04-17
hc-select acts as block element


jtrujill:
hc-select acts like a block element which isn't expected. The hc-select element should have `display: inline-block` applied to the host element. Instead of having the `<span class="chevron"></span>` wrap the native select element, a padding should be applied to the right side so that the chevron can be absolutely positioned there.

Related to #203

PR #212, opened 2018-04-17
Fix select box chevron overlap bug (fixes #203)

### LINKED ISSUE #203: Select chevron overlaps content inside of an element of dynamic width
If you set the parent div to `display: inline-block` and don't set the width, the select chevron overlaps the option content:

image

This can be fixed by adding `padding-right: 50px` to the `hc-select select` selector in `select.component.scss`.

isaaclyman:
Add padding to the select box so the chevron doesn't overlap the contents.

`ng serve` isn't working for me so I was unable to test this on the demo site, but the fix works on another project.


jtrujill:
There's a related issue to this #211. Would you mind taking a look at it, otherwise I could do a PR once this gets merged

isaaclyman:
@jtrujill I feel like you have a much better idea of what you're looking for there than I do. So I'll leave that one to you if that's okay.

isaaclyman:
Fixes #203.

PR #214, opened 2018-04-17
Fixes #211 - select acts as a block element

### LINKED ISSUE #211: hc-select acts as block element

jtrujill:
This has the same fix as PR #212 so if this gets accepted I'll close that

### DIFF
--- a/docs/app/components/components.module.ts
+++ b/docs/app/components/components.module.ts
@@ -9,7 +9,7 @@ import { TabsModule } from '../../../lib/src/tabs/tabs.module';
 import { NgModule } from '@angular/core';
 import { RouterModule } from '@angular/router';
 import { BrowserModule } from '@angular/platform-browser';
-import { FormsModule } from '@angular/forms';
+import { FormsModule, ReactiveFormsModule } from '@angular/forms';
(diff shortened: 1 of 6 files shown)


jtrujill:
I removed an input called highlight since it seemed to want to do the same thing that all inputs that implement ControlValueAccessor. When you use angular forms and an input implements ControlValueAccessor it'll add the appropriate classes ng-invalid, ng-dirty, ng-pristine... based on the state of the control.

This would be the more appropriate way to implement the functionality that hightlight provided. I would actually want to change the styling in the future so that we can create a simple forms.scss which styles all ng-invalid control appropriately

joeskeen:
@jtrujill #212 has been merged, do you need to revisit this?

jtrujill:
I'll update this PR so that it has those changes

jtrujill:
Updated

PR #219, opened 2018-04-17
Fixes #218 - parseBooleanAttribute works more as expected with standards

### LINKED ISSUE #218: anyToBoolean is incorrect
I created a helper function called anyToBoolean a while ago, but one of the issues with it is that it expects the value to be a boolean or to have a string value that is 'true' or 'false'. Here's the correct html 5 boolean attribute def:

```
2.5.2 Boolean attributes

A number of attributes are boolean attributes. The presence of a boolean attribute on an element represents the true value, and the absence of the attribute represents the false value.

If the attribute is present, its value must either be the empty string or a value that is an ASCII case-insensitive match for the attribute's canonical name, with no leading or trailing whitespace.

The values "true" and "false" are not allowed on boolean attributes. To represent a false value, the attribute has to be omitted altogether.
```

**_With that definition an empty required attribute would be marked as false which is incorrect._**

To fix the current implementation it should be checked to see if the value is empty or any value other than false. To try and match the name of the element to the required attribute wouldn't be worth the trouble.

### The following would pass:
- required=""
- required="true"
- required="something"
- [required]="isRequired" *isRequired=true

### The following would fail:
- required="false"
- [required]="isRequired" *isRequired=false

jtrujill:


jtrujill:
This change looks larger than it is but it's just a rename of the function

PR #221, opened 2018-04-17
Closes #215 - adds *.tgz to gitignore and removes dist.tgz from repo

### LINKED ISSUE #215: dist.tgz
I'm not 100% but I imagine this file is generated during the npm publish, correct? This shouldn't be committed to the repo. Every time this file changes it'll add another 1mb to the size of this repo and anything that is generated shouldn't be committed to a repo.

Is there any issue with me adding a gitingore rule for this file? I can't remember if npm uses the .gitignore file for it's publish feature

jtrujill:


jtrujill:
In gitignore a leading slash will ignore a file/directory relative to the root. An ignore rule with a slash only in the tail position then it will ignore any directories that match within the directory tree

Ex.
`/node_modules`

will only ignore a node modules found at the root

`node_modules/`
Will ignore all node_module directories in the repo

PR #224, opened 2018-04-18
Remove hc-subnav margin

### LINKED ISSUE #223: hc-subnav should not have a margin
`<hc-subnav>` has a bottom margin of 22px. This makes it harder to implement e.g. a sub-sub-navbar, such as the one in this design spec:

image

IMO a subnav should have no margin, and if a margin is desired, the element beneath should implement margin or padding.

Any dissenting opinions? If not, I'll make a pull request.

isaaclyman:
Fixes #223.


isaaclyman:
Fixes #223.

PR #225, opened 2018-04-18
Fixes #220 - Add index for all components

### LINKED ISSUE #220: Cleanup imports
I'm going to cleanup the imports in the project so that they are easier to read. An example of what the imports look like can be found in the components.module.ts

```
import { NavbarModule } from '../../../lib/src/navbar/navbar.module';
import { PopoverModule } from '../../../lib/src/popover/popover.module';
```

I'm going to use typescript pathMapping to help clean them up so that we can import the components needed by doing the following

```
import { NavbarModule } from '@healthcatalyst/cashmere/navbar/navbar.module';
import { PopoverModule } from '@healthcatalyst/cashmere/popover/popover.module';
```

@joeskeen what was the issue with using the index.ts file for every component? It's standard in all libs to have an index file for individual modules which can be imported. It is also how node module resolution works by looking for an index.js file. 

The imports above could be further shortened if we went with the standard:

```
import { NavbarModule } from '@healthcatalyst/cashmere/navbar';
import { PopoverModule } from '@healthcatalyst/cashmere/popover';
```

If you are able to tell me what the issue was I can also introduce index files for all components and make sure everything works

### LINKED ISSUE #230: Cashmere does not export ActiveModal
If I try to inject an ActiveModal object in my modal component, VS code offers to import it from Cashmere. But if I do, I get this error:

image

I think this is because ActiveModal is in the TypeScript declaration file for the module but not exported by the module itself.

~As a workaround, I can do this:~ This makes the IDE error go away but doesn't build:

```typescript
import { ActiveModal } from '@healthcatalyst/cashmere/src/modal/active-modal';
```

jtrujill:
Add index for all components, Fixes demo duplicate imports of BrowserModule and BrowserAnimationsModule, Fixes RouterModule.ForRoot uses that are outside of the root component

Fixes #230


jtrujill:
This seems to work. I fixed some issues with the demo app and can do an AOT build for the demo app and I can successfully run a build of the library. I was also able to use the generated library in another app without issue.

Issue #243, opened 2018-05-11
Navbar app icon as html


jtrujill:
I was messing around with the NavbarComponent today and couldn't get it to let me put a custom html element in the navbar. I just wanted to style some text and have that get projected into the app icon location but nothing was happening.

Right now it's limiting in that you need to pass a link to an asset instead of being able to define it as html. This issue is also present for the app brand location(tri-flame).

My proposal is to alter Navbar so that you instead have directives that will project the content into the correct location. For example:

### Using images:
```
<hc-navbar>
  <img src="./assets/tri-flame.svg" hcNavbarBrandIcon/>
  <img src="./assets/cqe.svg" hcNavbarAppIcon/>
<hc-navbar>
```

### Using html: 
```
<hc-navbar>
  <img src="./assets/tri-flame.svg" hcNavbarBrandIcon/>
  <div class="app-icon" hcNavbarAppIcon>
    CQE 
  </div>
<hc-navbar>
```

The end result should look about the same:
<img width="122" alt="screen shot 2018-05-10 at 7 11 34 pm" src="https://user-images.githubusercontent.com/5596515/39901497-1b80405a-5486-11e8-9ff7-838ce614f28d.png">

benjanderson:
The idea behind making the component opinionated as I did was to have the navbar consistent through each of the applications. The navbar is one of the most important elements that make are web apps look like a "brand". 

However the `<hc-navbar-icon>` are as you described and allow you to put any html you desire. I do agree with you on the username section, looking back it shouldn't be so restrictive. 

To quote the guiding principles I want to make things "flexible enough to be customized" but firstly a "unified, consistent user experience". An angular component gives developers an interface that they can plug into and things just work. If we want it more customizable I suggest we abstract the css so people can just use pure html and leave the navbar opinionated. If you look at your suggestions, it doesn't have much angular in it. The idea of extracting css from our components give us the best of both worlds while allowing the majority to subscribe to supported components.

andrew-frueh:
[...]

Issue #236, opened 2018-04-25
Breadcrumbs component doesn't support some common use cases


isaaclyman:
Suppose I have two pages in my app, with the following breadcrumb trails:

1. `Job History`
2. `Job History > Create Job`

The first one is easy to implement:

```typescript
const routes: Route[] = [
  {
    path: 'jobs',
    component: JobsComponent
    data: { breadcrumb: 'Job History' }
  }
]
```

If "Job History" and "Create Job" each have content they need to display, and it doesn't make sense to have the "Create Job" page inside of the "Job History" page, at the moment Cashmere does not offer a viable solution. This is a common use case because one page often links to another page from which the user may want to return.

### Attempt 1:

```typescript
const routes: Route[] = [
  {
    path: 'jobs',
    component: JobsComponent,
    data: { breadcrumb: 'Job History' },
    children: [
      {
        path: 'create',
        component: CreateJobComponent,
        data: { breadcrumb: 'Create New Job' }
      }
    ]
  }
]
```

With this configuration, the "Create Job" page will never appear because the "Job History" page does not have a `<router-outlet>` element.

### Attempt 2:

```typescript
const routes: Route[] = [
  {
    path: 'jobs/create',
    component: CreateJobComponent,
    data: { breadcrumb: 'Create New Job' }
  },
  {
    path: 'jobs',
    component: JobsComponent,
    data: { breadcrumb: 'Job History' }
  }
]
```

With this configuration, the "Create Job" page will appear but it will not have the "Job History" breadcrumb.

### Attempt 3:

```typescript
const routes: Route[] = [
  {
    path: 'jobs',
    data: { breadcrumb: 'Job History' },
    children: [
      {
        path: '',
        component: JobsComponent,
        data: { breadcrumb: null }
      },
      {
        path: 'create',
        component: CreateJobComponent,
        data: { breadcrumb: 'Create Job' }
      }
    ]
  }
]
```

With this configuration, the "Job History" page will have the breadcrumb trail `Job History >  `. Also, if you click the `Job History` breadcrumb from the "Create Job" page, it will redirect you to `jobs/jobs`, which is not a valid route.

### Proposed feature:

[...]

PR #249, opened 2018-05-15
extract breadcrumbs styles and clean things up a little

### LINKED ISSUE #125: enhance breadcrumbs to support dynamic creation
Currently the bread crumbs component reads the routes to build itself. This works for a simple site, but most sites including Atlas will need the ability to have the breadcrumbs built dynamically. 

As an example, atlas uses metadata to build its routes/breadcrumbs. This is something we need supported
image

benjanderson:
- extract out breadcrumbs styles
- extract out subnav styles
- fix issue with breadcrumbs not setting initial state
- breadcrumbs style page + guide
closes #125


jtrujill:
Why the need to extract the subnav and breadcrumb styles?

benjanderson:
This was something we had talked a long time ago with issue #125, Atlas had some unique needs and the solution we came up with is to extract and reuse the styles, so at least we had some reuse.

I think the long term solution would be a custom breadcrumb resolver proposed in #236. But this is the first step

jtrujill:
In the future it'd be best to have a PR for each issue so that things can be reviewed separately. What about the subnav?

benjanderson:
Good point, I will separate pull requests in the future. The reason I extracted the styles from the subnav as well was because in the breadcrumb demo I was only using markdown, so I couldn't use the `<hc-subnav>` component. 

I also agreed with your comment where fixed-top could be easier solved in css. https://github.com/HealthCatalyst/Fabric.Cashmere/issues/243#issuecomment-388231621 This takes us one step closer.

jtrujill:
We can worry about the fixed-top stuff in issue #243. I'll approve but I think this should be temporary like you said until we can get a better solution

PR #258, opened 2018-05-21
Fix setting tab-set active tab when using dynamic content

### LINKED ISSUE #250: Tabs *ngFor doesn't set active tab
When using iterating over an object to create a set of tabs using *ngFor the active tab never gets set

The following code causes the issue:

```
<hc-tab-set #tabs direction="horizontal">
    <hc-tab [tabTitle]="table" *ngFor="let table of tables"> <-------Issue
      <div class="table-container">
        <table class="hc-table">
          <thead>
            <tr>
              <th class="hc-col-sortable" [ngClass]="columnSortClasses('source')" (click)="triggerSort('source')">Source Field</th>I'll 
              <th class="hc-col-sortable" [ngClass]="columnSortClasses('target')" (click)="triggerSort('target')">Target Field</th>
              <th class="hc-col-sortable" [ngClass]="columnSortClasses('conversion')" (click)="triggerSort('conversion')">Convert</th>
            </tr>
          </thead>
          <tbody>
            <tr *ngFor="let mapping of sort(getMappings(table))"> 
              <td>{{getSourceColumn(mapping)}}</td>
              <td>{{getTargetColumn(mapping)}}</td>
              <td>{{mapping.requiredDataTypeConversion}}</td>
            </tr>
          </tbody>
        </table>
      </div>
    </hc-tab>
  </hc-tab-set>
``` 

This can be fixed by using a @ContentChildren and watching for any changes and setting an active tab.

To fix this whenever the content changes I'll set the first hc-tab to be the active one

jtrujill:
Fixes #250

### DIFF
--- a/lib/src/tabs/tab-set.component.html
+++ b/lib/src/tabs/tab-set.component.html
@@ -1,0 +1,22 @@
+<div class="hc-{{direction}}-tab-container">
+    <div class="hc-tab-bar-{{direction}}">
+        <div *ngFor="let tab of tabs">
+            <a *ngIf="routerEnabled" class="hc-tab-{{direction}}"
+               [routerLink]="tab.routerLink"
+               routerLinkActive="active"
+               (click)="setActive(tab)">
+                {{tab.tabTitle}}
+            </a>
+            <a *ngIf="!routerEnabled" class="hc-tab-{{direction}}"
+               [class.active]="tab.active"
+               (click)="setActive(tab)">
+                {{tab.tabTitle}}
+            </a>
+
+        </div>
+    </div>
+    <div class="hc-tab-content-{{direction}}">
+        <router-outlet *ngIf="routerEnabled"></router-outlet>
+        <ng-content *ngIf="!routerEnabled"></ng-content>
+    </div>
+</div>
--- a/lib/src/tabs/tab-set.component.ts
+++ b/lib/src/tabs/tab-set.component.ts
@@ -1,6 +1,6 @@
-import { Component, Input, ContentChildren, AfterContentInit, QueryList } from '@angular/core';
+import { AfterContentInit, Component, ContentChildren, Input, QueryList } from '@angular/core';
 import { TabComponent } from './tab.component';
-import { Router, ActivatedRoute, UrlSegment } from '@angular/router';
+import { ActivatedRoute, Router } from '@angular/router';
 
 export type TabDirection = 'horizontal' | 'vertical';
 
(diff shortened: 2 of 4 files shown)


lib/src/tabs/tab-set.component.html
```diff
@@ -0,0 +1,22 @@
+<div class="hc-{{direction}}-tab-container">
```
corytak: Thanks for moving this out to its own template. It's good to be consistent.

lib/src/tabs/tab-set.component.ts
```diff
@@ -10,52 +10,36 @@ export function throwErrorForMissingRouterLink(tabsWithoutRouterLink: TabCompone
...
+    templateUrl: './tab-set.component.html',
     styleUrls: ['../sass/_tabs.scss']
 })
 export class TabSetComponent implements AfterContentInit {
-    @ContentChildren(TabComponent) tabs: QueryList<TabComponent>;
-    @Input() direction: TabDirection = 'vertical';
     routerEnabled: boolean = false;
 
-    constructor(private router: Router, private route: ActivatedRoute) {}
+    @ContentChildren(TabComponent)
```
corytak: For decorators do we always want them to be above the property they decorate or to the left. I'm fine with either but prefer consistency.

lib/src/tabs/tab-set.component.ts
```diff
@@ -10,52 +10,36 @@ export function throwErrorForMissingRouterLink(tabsWithoutRouterLink: TabCompone
...
+    tabs: QueryList<TabComponent>;
+
+    @Input() direction: TabDirection = 'vertical';
+
+    constructor(private router: Router,
+                private route: ActivatedRoute) {
+    }
+
+    ngAfterContentInit(): void {
+        this.tabs.changes.subscribe(_ => {
```
corytak: It seems to be convention to use `() => ` instead of `_ => `

lib/src/tabs/tab-set.component.ts
jtrujill: I was looking at using a library called [Prettier](https://github.com/prettier/prettier) to handle this stuff so that we have a consistent code base

lib/src/tabs/tab-set.component.ts
jtrujill: I was working with more functional libraries and so this convention slipped into this code. However, if you have a function that requires that you define an input but you don't care about using that input it's standard to use an underscore signifying that you don't care about that value.

The reason it works for subscribe to use empty parenthesis is because the next argument is defined to be optional. This won't work in all situations.

PR #265, opened 2018-05-22
Added prettier to CI process

jtrujill:


jtrujill:
This will enforce that all files comply with prettier

guides/contribution-guide.md
```diff
@@ -27,4 +27,4 @@ Before you submit your pull request (PR), consider the following guidelines:
     *   Unit tests demonstrating that it functions as intended
 *   A new component should adhere to the [Health Catalyst style](http://cashmere.healthcatalyst.net).
 *   "Squash and merge" a PR to complete it.
-:::
+    :::
```
andrew-frueh: I noticed the other day that having the indent at the end causes the markdown interpreter to try and add another `hc-tile` inside the current tile.  But removing the indent fixes the issue.

jtrujill: This is a format change by prettier and can't be changed. What exactly is being done with these colons, since they aren't standard markdown

andrew-frueh: @benjanderson can explain better, but as you said, he wrote that custom into the interpreter to define an `hc-tile` around blocks of content.  So there is a ::: to start the tile and another ::: to close it.  It's strange that prettier changes the closing ::: but not the opening.  If we insert a newline in between would that make prettier keep from inserting the indent?

jtrujill: The correct thing to do, for now, is to figure out why it's interpreting that incorrectly. I need to figure out why the CI is failing though when I add a check for scss.

andrew-frueh: Well it might be simpler than that, because it's not like Prettier is doing it to every instance of ::: - it looks like it's only in the case of a line with a bullet list item followed immediately by the ::: on the next line.  Otherwise it seems to leave them alone.  That's why I'm thinking if we just insert a newline in those two files after the last bullet list item, prettier will hopefully not insert the indent.

jtrujill: I'll take a look. I was able to fix the problem, which had to do with the package.json file for whatever reason, so I'll ignore that file as a workaround. If I can fix the issue with the indentation I'll get that pushed out

andrew-frueh: Thanks Jeremy

jtrujill:
@andrew-frueh Fixed the issue with the ::: so this should be good to go

PR #268, opened 2018-05-23
* Fixes several issues with checkbox

jtrujill:
* Added unit test for checkbox
* Added normalize.css to _reset.scss

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -17,7 +17,7 @@
                         "polyfills": "src/polyfills.ts",
                         "tsConfig": "src/tsconfig.app.json",
                         "stylePreprocessorOptions": {
-                            "includePaths": ["projects/cashmere/src/lib"]
+                            "includePaths": ["dist/cashmere"]
                         },
                         "assets": ["src/assets", "src/favicon.ico", "src/web.config"],
                         "styles": ["src/styles.scss"],
@@ -69,7 +69,10 @@
                         "karmaConfig": "src/karma.conf.js",
                         "scripts": ["node_modules/popper.js/dist/umd/popper.min.js"],
                         "styles": ["src/styles.scss"],
-                        "assets": ["src/assets", "src/favicon.ico", "src/web.config"]
+                        "assets": ["src/assets", "src/favicon.ico", "src/web.config"],
+                        "stylePreprocessorOptions": {
+                            "includePaths": ["dist/cashmere"]
+                        }
                     }
                 },
                 "lint": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -5170,13 +5170,15 @@
             "dependencies": {
                 "abbrev": {
                     "version": "1.1.0",
-                    "bundled": true,
+                    "resolved": false,
+                    "integrity": "sha1-0FVMIlZjbi9W58LlrRg/hZQo2B8=",
                     "dev": true,
                     "optional": true
                 },
                 "ajv": {
                     "version": "4.11.8",
-                    "bundled": true,
+                    "resolved": false,
+                    "integrity": "sha1-gv+wKynmYq5TvcIK8VlHcGc5xTY=",
                     "dev": true,
                     "optional": true,
                     "requires": {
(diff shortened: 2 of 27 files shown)


jtrujill:
I downgraded typescript to 2.6.2 because Angular doesn't support anything above it and my IDE was giving me a bunch of errors about unassigned variables

jtrujill:
Disabled AOT for dev app because of issues with ngc throwing "Cannot determine the module for class". Here's a related issue https://github.com/angular/angular/issues/13590

The library itself compiles fine with AOT, but for whatever reason the demo app tries to compile a spec file which defines several components that are used for testing purposes only.

When we get the sites separated there shouldn't be an issue. It could also be some kind of misconfiguration with the build, but I honestly don't know why it's trying to compile *.spec.ts when there is an ignore in the tsconfig.app.json

lib/package.json
```diff
@@ -52,8 +52,8 @@
         "popper.js": "^1.12.9",
         "rxjs": "5.5.8",
         "shelljs": "^0.7.8",
-        "tslint": "^5.8.0",
-        "typescript": "^2.8.1",
+        "tslint": "~5.8.0",
+        "typescript": "~2.6.2",
```
benjanderson: thank you, this warning has been bothering me

docs/styles.scss
```diff
@@ -4,7 +4,7 @@ $fa-font-path: '../node_modules/font-awesome/fonts';
 $FontPathOpenSans: '../node_modules/npm-font-open-sans/fonts';
 @import '../node_modules/npm-font-open-sans/open-sans';
 
-@import 'cashmere';
+@import '../lib/src/sass/cashmere';
```
benjanderson: I am ok with this change but the reason it was working is the `includePath` in the config
https://github.com/angular/angular-cli/wiki/stories-global-styles

lib/src/checkbox/checkbox.component.spec.ts
```diff
@@ -1,26 +1,284 @@
...
+            inputElement.click();
+            fixture.detectChanges();
+
+            expect(checkboxInstance.checked).toBe(false);
+            expect(ngModel.valid).toBe(false);
+        });
     });
 });
+
+@Component({
```
benjanderson: If this is moved to the top, you could remove the tslint ignore

lib/src/sass/_reset.scss
```diff
@@ -1,152 +1,343 @@
...
-p,
-blockquote,
-pre,
-a,
-abbr,
-acronym,
-address,
-big,
-cite,
+/*! normalize.css v8.0.0 | MIT License | github.com/necolas/normalize.css */
```
benjanderson: We made quite a few changes to fix issues that were introduced when we added the reset style. It looks like we are replacing the entire file. A few things with this big of change:
1. Why are we switching our reset/normalize file?
2. We will probably need a style review by @andrew-frueh to make sure this doesn't introduce any new bugs.

docs/styles.scss
jtrujill: This will be changed in the CLI update

lib/src/checkbox/checkbox.component.spec.ts
jtrujill: I want to keep those at the bottom so it's easier to read the unit tests. The tests are the main focal point not those components

lib/src/sass/_reset.scss
jtrujill: The reason I replaced it was that I was experiencing a weird issue where these styles for the header elements were overriding the styles defined in the demo app. I can take another look but I couldn't figure out why this was taking precedence. This normalize file is very common and is also used by bootstrap

benjanderson: I have had a few issues with the current reset as well, so I am ok with this change. I would like to point out this is just more risky. Lets make sure we validate styles are the same when this gets checked in

lib/src/checkbox/checkbox.component.spec.ts
benjanderson: If thats the case, maybe we should remove this from the tslint rules.

PR #270, opened 2018-05-24
Converted over to angular/cli@6.0.3 with library support

jtrujill:
This is working and you can get the demo app working. I need to still do a few things like modify the npm scripts to work with the new system and I need to figure out how to copy over static content such as sass into the library dist directory. Otherwise this is working great, but it did require a lot of work


jtrujill:
To get it working you'll need to use the following command: `ng build cashmere && npm serve user-guide`

jtrujill:
Looks like I'll also need to fix the apps assets. Please leave comments if you see any other issues

jtrujill:
Okay, I believe that I've fixed all issues and the only thing I need to do is copy all the sass files into a sass directory within the dist for the library. Should be an easy fix and I'll have that out today

jtrujill:
Just pushed a node script which copies over the necessary stuff into the dist folder. This is all good to go. I'd like to get this merged as soon as possible because of how large it is.

angular.json
```diff
@@ -0,0 +1,151 @@
+{
```
benjanderson: woot, woot! angular 6!

projects/cashmere/ng-package.json
```diff
@@ -0,0 +1,11 @@
+{
```
benjanderson: nice new project structure :)

scripts/postbuild.js
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env node
```
benjanderson: very nice!

src/browserslist
```diff
@@ -0,0 +1,9 @@
+# This file is currently used by autoprefixer to adjust CSS to support the below specified browsers
```
benjanderson: whats this?

src/polyfills.ts
```diff
@@ -1,68 +1,75 @@
...
-/**
- * Date, currency, decimal and percent pipes.
- * Needed for: All but Chrome, Firefox, Edge, IE11 and Safari 10
- */
-// import 'intl';  // Run `npm install --save intl`.
-/**
- * Need to import at least one locale-data with intl.
- */
-// import 'intl/locale-data/jsonp/en';
+/**
```
benjanderson: why re-comment many of these?

src/browserslist
jtrujill: Autoprefixer is a tool that handles css prefixes for features that require the -webkit or -moz. When this is configured it will correctly add those prefixes to the browsers we have defined as our min

src/polyfills.ts
jtrujill: I'll take a look at this

src/browserslist
benjanderson: cool!

src/polyfills.ts
benjanderson: ok, I am going to block this p: until this is fixed. These fixed several bugs

PR #277, opened 2018-05-29
Upgraded to Angular 6

### LINKED ISSUE #276: Upgrade to Angular 6
When do we want to upgrade to Angular 6? I could do it now without much hassle. If you read the changes that have been done for 6, they all seem to be about dev tooling. I don't think it'd be difficult for developers to upgrade to 6, but before doing this I need to know from those using it if they're in a position to make that upgrade soon

jtrujill:
Closes #276


jtrujill:
This PR includes rxjs-compat which allows us to use deprecated rxjs methods until we get those fixed. I want to leave that to another PR and when that is taken care of we can remove rxjs-compat

PR #279, opened 2018-05-29
Updated guides and readme to reflect upgrade to angular-cli@6

### LINKED ISSUE #278: Update getting-started guide for the new angular-cli 6 changes
Since upgrading to angular-cli@6 the documentation we have is no longer valid. This needs to be updated with new instructions on how to run the user-guide app

jtrujill:
Closes #278

### DIFF
--- a/guides/packaging-library.md
+++ b/guides/packaging-library.md
@@ -9,13 +9,10 @@
 In order to publish an npm package, you must be a part of the [@healthcatalyst npm members](https://www.npmjs.com/org/healthcatalyst/members).
 
 1.  Verify all guidelines pass from the [Guidelines for Contribution](http://cashmere.healthcatalyst.net/guides/contribution-guide).
-1.  Increment the version in `package.json` and `lib/package.json`. Use [SemVer](http://semver.org/) for versioning.
-1.  Navigate to lib directory `cd ./lib/`.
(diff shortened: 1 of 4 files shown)


guides/packaging-library.md
```diff
@@ -1,21 +1,18 @@
...
+
+:::
+
+##### Building npm package and publishing
+
+In order to publish an npm package, you must be a part of the [@healthcatalyst npm members](https://www.npmjs.com/org/healthcatalyst/members).
+
+1.  Verify all guidelines pass from the [Guidelines for Contribution](http://cashmere.healthcatalyst.net/guides/contribution-guide).
+1.  Increment the version in `package.json` and `projects/cashmere/package.json`. Use [SemVer](http://semver.org/) for versioning.
+1.  Run `npm run build:lib` to bundle the library.
```
benjanderson: didnt this change to `npm run build`?

guides/packaging-library.md
```diff
@@ -1,21 +1,18 @@
...
+1.  Navigate to dist directory `cd ./dist/cashmere`.
```
benjanderson: also I think this should just be `cd ./dist`

guides/packaging-library.md
jtrujill: `npm run build` is a composition of `npm run build:lib && npm run build:user` so if you just need the library built you just use `npm run build:lib` however if you want to build user-guide you'll need to run npm run build because user-guide is dependent on the library

PR #288, opened 2018-06-05
Finished new hc-input component

jtrujill:


projects/cashmere/src/lib/tile/tile.component.spec.ts
```diff
@@ -1,24 +1,25 @@
...
-        fixture = TestBed.createComponent(TileComponent);
-        component = fixture.componentInstance;
-        fixture.detectChanges();
-    });
-
-    it('should create', () => {
-        expect(component).toBeTruthy();
-    });
-});
+// import {async, ComponentFixture, TestBed} from '@angular/core/testing';
```
corytak: Any reason why we don't want to delete this file or use `xdescribe`?

projects/cashmere/src/lib/input/hc-form-field.component.html
```diff
@@ -0,0 +1,27 @@
+<div class="hc-form-field-wrapper">
+    <span class="hc-form-field-label-wrapper">
+                <label class="hc-form-field-label"
+                       [attr.for]="_control.id">
+                    <ng-content select="hc-label"></ng-content>
+                    <span class="hc-required-marker" aria-hidden="true"
+                          *ngIf="(_control.required )&& !_control.disabled">&nbsp;*</span>
```
corytak: in this scenario `_control` is public in the sense that it is visible to the template, but it's marked with an underscore meaning it's private in the sense that it's not part of the `hc-form-field`'s public api. Is this a correct understanding?

jtrujill: That's correct. We'll have to adopt some kind of convention to be able to distinguish public API vs private and underscore is very common

projects/cashmere/src/lib/tile/tile.component.spec.ts
jtrujill: Forgot to uncomment this. I was having weird issues with this file

jtrujill:
@corykon I removed the clear button because I didn't like the implementation and it's not important for initial functionality. Rather than require a whole new element within the dom to represent the clear button, I'd rather have a flag that can be set.

I'd like to give that some thought because I want to hc-form-field to be used with all input types: input, select, date... It's easier to add new features than to remove features, so it's best to flesh ideas out before piling on too much

I also think we need to go over the design of the button in more depth. I'll be doing a follow up PR that adds more interactivity with different events: focus, blur, error, disabled. I'll need to have a conversation with @andrew-frueh about the aesthetic design of the input

corykon:
Gotcha! Well I won't block the PR here, but it'd be cool to not lose the functionality (even if not the perfect implementation) until we have a replacement

projects/cashmere/src/lib/input/hc-form-field.component.scss
```diff
@@ -0,0 +1,121 @@
...
+    display: inline-flex;
+    align-items: baseline;
+    box-sizing: border-box;
+    width: 100%;
+
+    padding: 0 7px 0 0;
+    border: 1.5px solid $gray-300;
+}
+
+.hc-form-field-prefix,
```
benjanderson: Here they are! That doesn't match the css class listed in the other files...

projects/cashmere/src/lib/input/hc-form-field.component.spec.ts
```diff
@@ -0,0 +1,165 @@
+/* tslint:disable:no-use-before-declare component-class-suffix */
```
benjanderson: nice tests!

benjanderson: I ssee this tslint exception enough I think it could be disabled from the .tslint file

projects/cashmere/src/lib/input/hc-prefix.directive.ts
```diff
@@ -0,0 +1,8 @@
+import {Directive, HostBinding} from '@angular/core';
+
+@Directive({
+    selector: '[hcPrefix]'
+})
+export class HcPrefixDirective {
+    @HostBinding('class.hc-prefix') hostHcPrefixClass = true;
```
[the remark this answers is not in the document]
jtrujill: It's not a float and it's applying the name of the directive to the component it's applied to

projects/cashmere/src/lib/input/hc-suffix.directive.ts
```diff
@@ -0,0 +1,8 @@
+import {Directive, HostBinding} from '@angular/core';
+
+@Directive({
+    selector: '[hcSuffix]'
+})
+export class HcSuffixDirective {
+    @HostBinding('class.hc-suffix') hostHcSuffixClass = true;
```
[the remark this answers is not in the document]
jtrujill: Same as above

projects/cashmere/src/lib/input/hc-form-field.component.spec.ts
jtrujill: Which are you referring to? For tests it's alright but I'd like to have that for the rest of the app

projects/cashmere/src/lib/input/hc-form-field.component.scss
jtrujill: This refers to something else

projects/cashmere/src/lib/input/hc-form-field.component.scss
```diff
@@ -0,0 +1,121 @@
...
+}
+
+.hc-form-field-infix {
+    display: block;
+    position: relative;
+    flex: auto;
+    min-width: 0;
+    width: $mat-form-field-default-infix-width;
+
+    padding: $infix-padding 5px $infix-padding 5px;
```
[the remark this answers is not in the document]
jtrujill: Done

jtrujill:
@andrew-frueh I made the changes that you requested. At sometime we'll need to get together to discuss some styling that will need to happen. I'll hold off on the red asterisk until we can go over those other stylings

PR #314, opened 2018-06-11
Added dgeni for doc generation

### LINKED ISSUE #242: Documentation Generation
Right now we don't have great documentation for Cashmere which requires that users of the library look at the source code to know what to use. Our current way of creating a demo component involves a lot of work and ends up not being very informative. As our components evolve, the documentation that we currently have will get our-of-sync with the code because we aren't generating our documentation from the code. 

I've been taking a look at several documentation generators and I think the one I like the most is Dgeni(Jenny). You use JSDoc style comments to document your code and you're able to process your documents with Processors to manipulate, filter or add meta-data about your documents. Dgeni also supports templates, and using the generated meta-data you're able to pick the template used for each document.

A good starting point would be to have each component author go through and document their code using jsdoc style comments and getting Dgeni setup with templates and processors. This would allow us to get some documentation out the door for developers to start using. To simplify the documentation and creating examples for a component, it would be best to replace http://cashmere.healthcatalyst.net/ with the generated docs from Dgeni. 

Initially we should keep the site very basic just to get documentation out the door for 1.0 release. Dgeni allows for you to use your own annotations and using that fact we could create an annotation which we'd attach to the share link of a stackblitz app. In the processor we could look for that annotation and generate an iframe with the source pointed to that stackblitz app. This would allow us to embed demo's of our components within our generated documents.

Documenting the library will be a top goal for me and I think that Dgeni is the tool to use. I would like to take on the task of getting Dgeni configured, but I'll need help with getting things documented. What are people's thoughts?

jtrujill:
Closes #242


jtrujill:
This will generate html for our code but it'll need to be styled still

PR #321, opened 2018-06-14
style: added login page styles

benjanderson:
I am helping Marla add a login page for https://docs.healthcatalyst.com/ and needed to implement a login page. I decided to do a little extra work and add it to cashmere. My project doesn't really need angular so I made it so it can handle both angular and non angular setup


jtrujill:
Can this PR be altered to merge into branch `dev`. `master` is going to be our release branch.

benjanderson:
will do, thanks @jtrujill

PR #327, opened 2018-06-14
style: added login page styles

benjanderson:
I am helping Marla add a login page for https://docs.healthcatalyst.com/ and needed to implement a login page. I decided to do a little extra work and add it to cashmere. My project doesn't really need angular so I made it so it can handle both angular and non angular setup


andrew-frueh:
Hey @benjanderson , thanks for doing this - great addition to the library.  I think this whole login flow might become (or already is) a Fabric service, but that's why doing this as a style was a great idea.  They can easily leverage what you've done here.  Just two small changes and a Firefox bug I noticed.

1) On .hc-login .privacy - can you add `line-height: 1rem;` to that?  The text was pretty crammed together otherwise.

2) Can you add `margin-top: 5px;` to .hc-login .login-input?  Just do give a little more space between the label and the input.

And then last, take a look at what's going on with Firefox.  Probably something `display` related, but I didn't have a chance to troubleshoot it.

screen shot 2018-06-20 at 11 21 06 am

andrew-frueh:
Sorry one other thing: on @media (min-width: 576px) .hc-login - can you make the `width: 400px;` instead of 300?  It gives the box a little more space and doesn't feel so cramped, especially with the legal disclaimer.

PR #337, opened 2018-06-20
fix(hc-form-field.component.scss): set background to white

### LINKED ISSUE #336: Input box background color should be white
**Describe the bug**
The input box takes the same color as the background, but I _think_ all our mockups have it with a white background, regardless of the container, similar to what the select box has in the demo below.

**To Reproduce**
https://stackblitz.com/edit/cashmere-demo-trmhwv

**Expected behavior**
I'd expect the background of the input box to be white, regardless of what it's contained within, with the option to change the color (not sure how exactly we want to accomplish that).

**Screenshots**
image

jonest:
Set the backgronud of the hc-form-field-flex to white by default

fix #336


jonest:
I was thinking there was more to it than just this since mine usage seemed to have messed up the placeholder text, but I didn't see the same here :thinking: Did we want to add in a way to configure this more easily? I didn't have any thoughts off hand myself.

jtrujill:
If you're going to set the background-color to an explicit color then you should do the same for the text color

jonest:
@jtrujill I'm on it! I just noticed that there's an `input.scss` file. I was thinking I'd have to use the `ng-deep` stuff and didn't want to do that. I was thinking I'd go with `$slate-gray-300`, does that sound good @andrew-frueh ?

PR #349, opened 2018-06-24
docs(button component): adds jsdoc comments for button component

### LINKED ISSUE #295: Button - Documentation Update
Update component code to work with Dgeni doc generation.  See the [project overview](https://github.com/HealthCatalyst/Fabric.Cashmere/projects) for more details.

jtrujill:
documents button component's public api while hiding fields/methods that are private

close #295

### DIFF
--- a/projects/cashmere/src/lib/button/split-button/split-button.component.ts
+++ b/projects/cashmere/src/lib/button/split-button/split-button.component.ts
@@ -2,10 +2,12 @@ import {ChangeDetectionStrategy, Component, ElementRef, EventEmitter, HostBindin
 import {parseBooleanAttribute} from '../../util';
 import {validateButtonColor} from '../button.component';
 
+/* SplitButton click event */
 export class SplitButtonClickEvent {
     constructor(public source: SplitButtonComponent) {}
 }
 
+/* Split button provides a primary action button along with a secondary menu of actions */
 @Component({
     selector: 'hc-split-button',
     templateUrl: './split-button.component.html',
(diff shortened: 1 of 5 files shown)


projects/cashmere/src/lib/button/split-button/split-button.component.ts
```diff
@@ -18,13 +20,22 @@ export class SplitButtonComponent {
     private _disabled: boolean = false;
     private _color: string = 'primary';
 
+    /* primary button's click event */
     @Output() click = new EventEmitter<SplitButtonClickEvent>();
 
+    /* Additional information shown as tooltip */
     @Input() title: string;
+
+    /* Type of button */
```
corykon: I'm not really sure what this means. Can we add more details to this comment?

projects/cashmere/src/lib/button/split-button/split-button.component.ts
```diff
@@ -18,13 +20,22 @@ export class SplitButtonComponent {
...
     @Input() type = 'button';
+
+    /* Used as a reference in JavaScript, or to reference form data after a form is submitted */
     @Input() name: string;
+
+    /* Value of button */
```
corykon: Might be a little clearer to explain that this is the text that will appear on the button.

projects/cashmere/src/lib/button/split-button/split-button.component.ts
```diff
@@ -34,6 +45,7 @@ export class SplitButtonComponent {
         this._tabIndex = value == null ? 0 : value;
     }
 
+    /* Theme color */
```
corykon: Can we add the available options here? If possible, it might be preferable to use a unioned string literal of available options as the type so documents itself.

projects/cashmere/src/lib/button/split-button/split-button.component.ts
jtrujill: Sure this is actually the native type attribute, nothing special

PR #352, opened 2018-06-25
Icon docs

jtrujill:


jtrujill:
Good catch @corykon. I'm going to move those to directives instead of having them be css classes. The added benefit is that you get auto-completion when they are directives. I'll create a new issue for this

PR #355, opened 2018-06-25
docs: adds demos to docs

jtrujill:
adds demos for cashmere components along with source code viewing for those demos.


jtrujill:
Easiest way to test this app is to run `npm run build` and then do `ng serve`. Build will compile cashmere, cashmere-examples, user-guide, create docs and move those docs into user-guide's assets/.

Be careful when switching between this branch and others since it will add files into the user-guides assets/. There is a .gitignore rule in this branch but not in the other branches so you could easily commit those generated files if not careful

jtrujill:
This is also a first iteration on this design. The styling still needs some work and I need to add some additional functionality like stackblitz and also a collapsable demo source block. 

The demos also need to be simplified so that they only show one aspect of a component rather than everything in one demo. The demos also need headers, wrappers and any other superfluous markup which goes beyond the component being demoed.

Each component should have multiple demos and be focused on one aspect. To do this you navigate to `projects/cashmere-examples/src/lib` create a directory for the demo. Use a descriptive name for the directory. Then add the files `{directory-name}-example.{ts,css,html}`. For example if you have the directory name `checkbox-forms`, you would have the following files `checkbox-forms-example.{ts,css,html}`. You must follow this naming convention.

In the demo ts file you must use the following annotation on the demo component

```
/**
 * @title {Title}
 */
```
This will indicate that this is the primary demo component and will give it a title. If that is missing the demo won't work. You can have multiple classes defined for a demo but they must be defined in the same ts file. These demos are meant to be simple references for users and not full apps, so don't make things difficult.

Next in the document-items.service there is a list of DocItems which is config that drives the entire component section of user-guide. For each component it can have an examples property which should be an array and this is where you place the directory name of the demo component. So going back to the checkbox-forms example you'd have the following config for checkbox:

`{id: 'checkbox', name: 'Checkbox', examples: ['checkbox-forms']},`

Now you just need to build the examples with `npm run build:examples && npm run copy-assets` and `ng serve`

jtrujill:
There's also a src/app/old directory that has the old demo component files. I've commented out the ts but all the code is there. The only demos that I don't have working are breadcrumbs and picklist. Picklist needs to be simplified because there is just too much data there for a user to see as an example.

projects/cashmere-examples/src/lib/progress-indicator-overview/progress-indicator-overview-example.css
```diff
@@ -0,0 +1,40 @@
+/*.progress-example {*/
```
corykon: Any way we could've left these as .scss, and avoided commenting them out? Left as is probably would work just fine in a stackblitz, which is our intended target anyways.

jtrujill: I probably could, but I want to keep the styling agnostic in-case people chose to go with Less or something other than Sass. I can take a look at that in another PR. I really just want to get this thing merged so I don't have to deal with complicated merges.

src/app/core/document-items.service.ts
```diff
@@ -0,0 +1,42 @@
...
+    {id: 'drawer', name: 'Drawer', examples: ['drawer-overview']},
+    {id: 'icon', name: 'Icon', examples: ['icon-overview']},
+    {id: 'input', name: 'Input', examples: ['input-overview']},
+    {id: 'list', name: 'List', examples: ['list-overview']},
+    {id: 'modal', name: 'Modal', examples: ['modal-overview']},
+    {id: 'navbar', name: 'Navbar', examples: ['navbar-overview']},
+    {id: 'pagination', name: 'Pagination', examples: ['pagination-overview']},
+    {id: 'picklist', name: 'Picklist'},
+    {id: 'popover', name: 'Popover', examples: ['popover-overview']},
+    {id: 'progress-indicator', name: 'Progress Indicator', examples: ['progress-indicator-overview']},
```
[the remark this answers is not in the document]
jtrujill: Thanks

jtrujill:
@corykon Everything is supposed to go into the dev branch from now on. Only when someone does a merge into master will it trigger a publish. So this can be merged without it being pushed out. However, I believe that even in it's current state, it'd be more use to people. I was hoping to get these new docs pushed out by the end of the week.

@corykon @corytak @andrew-frueh If you guys are up for a review I'll get one scheduled as soon as possible. I was told that @benjanderson will be out of the country

jtrujill:
@corykon are you good with the PR or do you need anything explained?

corykon:
Nope, merge 'er in

PR #359, opened 2018-06-27
docs: add jsdocs for chips

### LINKED ISSUE #297: Chips - Documentation Update
Update component code to work with Dgeni doc generation.  See the [project overview](https://github.com/HealthCatalyst/Fabric.Cashmere/projects) for more details.

andrew-frueh:
document chip's public api while hiding fields/methods that are private

closes #297


projects/cashmere/src/lib/chip/chip-row/chip-row.component.ts
```diff
@@ -10,10 +11,11 @@ import {parseBooleanAttribute} from '../../util';
     encapsulation: ViewEncapsulation.None
 })
 export class ChipRowComponent {
-    _wrap: boolean = true;
+    private _wrap: boolean = true;
```
jtrujill: this needs to be public since it's used in the template

jtrujill: Leave the private off. We should be following the standard of leaving the public modifier off since that is the default access level

projects/cashmere/src/lib/chip/chip.component.ts
```diff
@@ -1,18 +1,21 @@
...
+/** Chips represent complex entities in small blocks, such as filters, contacts, or system information */
 @Component({
     selector: 'hc-chip',
     template: `<div [class]="color" [ngClass]="{'close': _action, 'chip': true}"><ng-content></ng-content></div>`,
     styleUrls: ['./chip.component.scss'],
     encapsulation: ViewEncapsulation.None
 })
 export class ChipComponent {
-    _action: boolean = false;
+    private _action: boolean = false;
```
jtrujill: Private needs to be left off since it's used in the template

jtrujill: or _action in the template needs to be changed to use the action property

andrew-frueh: Good catch @jtrujill - and thanks for talking this through with me.  Completely missed this before, so the docs update is making our code better!  I adjusted the templates in chips and chip-row, checked a build and everything still works fine.

PR #364, opened 2018-07-25
docs: doc styling

andrew-frueh:
Updates to the style of the dgeni templates following [the prototype](https://xd.adobe.com/view/7a081799-3095-46c2-b22b-3b9f6393eb1d/screen/9f562696-4bae-4b0f-9b89-fd07a33fe5b4/Components-API) that was vetted with the community.

Styling the dgeni docs was actually really nice, they give unique classes to just about everything, so I was happy how smoothly that went.  @corykon - I'd love to have your eye on my changes and see if there's anything you'd do differently from a design perspective.  

A few next steps for us to look at if you guys are good with this update:

1. **Style Resets** - There is something weird since we did this dgeni update going on with the css when you go between "Styles" or "Guides" and "Components".  It's almost like the css gets reset or overridden now when you go back to "Styles" after visiting "Components".  Could use some help debugging that.
2. **Examples** - I've got the Examples section roughed in, but there is still some work to be done there.  For one, the tabs at the top of the tile don't display an active state.  And I need to do some cleanup on all the demo html now that they are in that format.  But @jtrujill mentioned that we might split them up, so I figured we should talk that through first.
3.  **Usage** - we had also talked about adding a third tab to the components where we could store markdown, and I believe that was on your radar, Jeremy.  So let me know what you think about how we can attack that.

### DIFF
--- a/src/app/components/component-list/components-list-router.module.ts
+++ b/src/app/components/component-list/components-list-router.module.ts
@@ -14,9 +14,9 @@ const routes: Route[] = [
                 path: ':id',
                 component: ComponentViewerComponent,
                 children: [
-                    {path: 'API', component: ComponentApiComponent, pathMatch: 'full'},
-                    {path: 'Examples', component: ComponentExamplesComponent, pathMatch: 'full'},
-                    {path: '**', redirectTo: 'API'}
+                    {path: 'api', component: ComponentApiComponent, pathMatch: 'full'},
+                    {path: 'examples', component: ComponentExamplesComponent, pathMatch: 'full'},
+                    {path: '**', redirectTo: 'api'}
                 ]
             }
         ]
--- a/src/app/components/component-list/components-list-router.module.ts
+++ b/src/app/components/component-list/components-list-router.module.ts
@@ -14,9 +14,9 @@ const routes: Route[] = [
                 path: ':id',
                 component: ComponentViewerComponent,
                 children: [
-                    {path: 'api', component: ComponentApiComponent, pathMatch: 'full'},
-                    {path: 'examples', component: ComponentExamplesComponent, pathMatch: 'full'},
-                    {path: '**', redirectTo: 'api'}
+                    {path: 'API', component: ComponentApiComponent, pathMatch: 'full'},
+                    {path: 'Examples', component: ComponentExamplesComponent, pathMatch: 'full'},
+                    {path: '**', redirectTo: 'API'}
                 ]
             }
         ]
--- a/src/app/components/component-viewer/component-api/document-viewer/document-viewer.component.scss
+++ b/src/app/components/component-viewer/component-api/document-viewer/document-viewer.component.scss
@@ -1,10 +1,96 @@
+@import 'scss/cashmere';
+
 .hc-doc-viewer {
     display: block;
 }
 
-.docs-api {
-    background-color: #fff;
-    border-radius: 5px;
-    margin-top: 10px;
-    padding: 10px 25px;
+.hc-tile:first-of-type {
+    margin-top: 20px;
+}
+
+.docs-api-h3 {
+    font-size: 1.07142857rem;
+    text-transform: uppercase;
+    color: #006d9a;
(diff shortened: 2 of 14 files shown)


src/app/components/component-viewer/component-api/document-viewer/document-viewer.component.scss
```diff
@@ -1,10 +1,96 @@
...
-    background-color: #fff;
-    border-radius: 5px;
-    margin-top: 10px;
-    padding: 10px 25px;
+.hc-tile:first-of-type {
+    margin-top: 20px;
+}
+
+.docs-api-h3 {
+    font-size: 1.07142857rem;
```
corykon: Can we use the fontSize() helper function?

src/app/components/component-viewer/component-examples/example-viewer/example-viewer.component.html
```diff
@@ -1,10 +1,14 @@
-<hc-tab-set direction="horizontal">
-    <hc-tab [tabTitle]="ext" *ngFor="let ext of ['html', 'ts']">
-        <pre>
-            <hc-doc-viewer [documentUrl]="exampleFileUrl(ext)">
-            </hc-doc-viewer>
-        </pre>
-    </hc-tab>
-</hc-tab-set>
+<div class="hc-tile">
+    <h3 class="docs-api-h3" >Example Name Header</h3>
```
corykon: Do plan to stick the actual component name here?

andrew-frueh:
Thanks Cory!  I actually added the import of cashmere.scss to Document Viewer for convenience, but it's not really necessary.  I'd realize that calls the css reset, so I'll remove that and see if that resolves the issue.

src/app/components/component-viewer/component-examples/example-viewer/example-viewer.component.html
andrew-frueh: Ah yes, thanks for calling this out.  I left this in because we talked about having multiple examples on the page.  Breaking up our big pages of examples into smaller chunks so the code is easier to consume.  So these titles at the top of the tile would be the name of the specific example - I just wasn't sure how @jtrujill was planning on dynamically pulling multiple examples in.  But hopefully we can pull a name in as well on those examples that can go in this place.

jtrujill: Yeah this information is available for every example. If you look at the examples they have a title annotation and that's how that gets setup

jtrujill:
I'll get cracking on that third tab that will pull in the markdown "overview" or whatever it's being called.

src/app/components/component-viewer/component-examples/example-viewer/example-viewer.component.html
andrew-frueh: Ah perfect, okay I see it now.  Awesome.  Way to think ahead on how you set those up.

src/app/components/component-list/components-list-router.module.ts
```diff
@@ -14,9 +14,9 @@ const routes: Route[] = [
                 path: ':id',
                 component: ComponentViewerComponent,
                 children: [
-                    {path: 'api', component: ComponentApiComponent, pathMatch: 'full'},
-                    {path: 'examples', component: ComponentExamplesComponent, pathMatch: 'full'},
-                    {path: '**', redirectTo: 'api'}
+                    {path: 'API', component: ComponentApiComponent, pathMatch: 'full'},
```
jtrujill: This is a bit inconsistent with our urls, so if we want to change the case on the tabs we could handle it by having a "display" version or by using pipe

jtrujill: If you want, could you revert this change to the urls and I'll make the change so the tabs show up with the correct case. I'll approve

andrew-frueh: Okay - cool, will do.  So yeah the display versions will be "API", "Examples", and "Usage"

PR #383, opened 2018-08-02
docs(all): examples cleanup

andrew-frueh:
For the release of our new docs, this PR goes through all the component examples and breaks them into smaller pieces (hence the massive number of files changed); also adds some general styling cleanup for the release


jtrujill:
@corykon When it comes to the breadcrumb example we might need to hold off for a bit until I can find a clean way to demo the component. Also as-is the component needs quit a bit of work before it's useful

projects/cashmere-examples/src/lib/chip-action/chip-action-example.ts
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * @title Action Chips
+ */
+@Component({
+    selector: 'chip-action-example',
+    templateUrl: 'chip-action-example.html'
+})
+export class ChipActionExample {
+    hideChip(event: any) {
+        event.target.style.display = 'none';
```
jtrujill: Doesn't an action chip already hide itself? I believe that's what the API docs say

projects/cashmere-examples/src/lib/drawer-double/drawer-double-example.html
```diff
@@ -0,0 +1,21 @@
+<div class="drawer-container">
```
jtrujill: Could this be changed to drawer-side-example, since it's main purpose is to show that a drawer can be stationed to the side of the window

projects/cashmere-examples/src/lib/chip-action/chip-action-example.ts
andrew-frueh: nah, it just puts the 'X' and adds the pointer cursor to the chip.  I figured the hiding was something the developer would want to do themselves so they could remove it from filters at the same time.  If you think it would be useful to change that - let's add an issue

projects/cashmere-examples/src/lib/drawer-double/drawer-double-example.html
andrew-frueh: sure thing, done.

jtrujill:
Looks good! Thanks for taking that on I was kinda dreading having to do that @andrew-frueh

projects/cashmere-examples/src/lib/chip-action/chip-action-example.ts
jtrujill: makes sense

PR #387, opened 2018-08-06
docs(cashmere-examples): moved styles for cashmere examples to css files

### LINKED ISSUE #372: Add ability to view css for the component examples
Need to add a css tab to the example-viewer.component.html and need to add a css file for each example component.

jtrujill:
users can now view css for a cashmere example on a separate page rather than it being embedded in
the typescript

close #372

### DIFF
--- a/projects/cashmere-examples/src/lib/accordion-overview/accordion-overview-example.css
+++ b/projects/cashmere-examples/src/lib/accordion-overview/accordion-overview-example.css
@@ -1,0 +1,1 @@
+/* No CSS rules defined for this example */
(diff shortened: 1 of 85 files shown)


jtrujill:
@andrew-frueh The issue with that is that it'd be more config to put into DocItem[], which probably isn't worth it

andrew-frueh:
...another possibility @jtrujill would be to include a comment in all the empty css files that just says:
`/*No CSS rules defined for this example*/`
That way at least it's clear that nothing is broken

jtrujill:
Simple and solves the issue

jtrujill:
I'll quickly do that

PR #392, opened 2018-08-06
390 typography additions

### LINKED ISSUE #390: Typography Additions
Started switching over to Cashmere styles, and found a few handy typography styles that I use often that probably could/should be in Cashmere.

The following new classes
- `.hc-text-link`: styled just like `<a>`, but in cases where I don't actually want to use an anchor tag
- Sizes: 
     - `.hc-type-xsm`: fontSize(12px)
     - `.hc-type-sm`: fontSize(14px)
     - `.hc-type-md`: fontSize(16px)
     - `.hc-type-lg`: fontSize(18px)
- Weights (`hc-type-bold` at 600 already exists): 
     - `.hc-type-light`: font-weight: 300
     - `.hc-type-reg`: font-weight: 400
     - `.hc-type-xbold`: font-weight: 700
- Casings:
     - `.hc-type-titlecase`: capitalize first letter of each word
     - `.hc-type-uppercase`: capitalize all letters
     - `.hc-type-lowercase`: all letters lowercase
- `hc-type-ellipsis`:
     -often needed with displaying custom user input, or to help with responsiveness
image

Could have more/fewer helpers than these... but these are what I use in PopBuilder

Nothing changed, just adding new classes

**Additional context**
...

corykon:
No need for this to make into the docs release.

Closes #390 

Left out classes for case (upper/lower/title), as angular does have built in pipes for that stuff, and it might be just me that finds that sort of helper class useful.

Did include classes for size, weight, ellipsis, and non anchor based text-links. We can pull stuff out and tweak it to meet what @andrew-frueh and the group decides it best, but this is basically a cleaned up port of the stuff I'm using in pop builder.


corykon:
Just pushed all the changes we had discussed this morning, including adding the different buttonStyles options in the descriptions until Jeremy has a chance to add in the code needed to automatically output Typescript types.

Could someone give this one last look before I merge? @andrew-frueh  @jtrujill  @benjanderson

projects/cashmere/src/lib/button/button.component.ts
```diff
@@ -4,13 +4,7 @@
...
 import {parseBooleanAttribute} from '../util';
 
-const supportedColors = ['primary', 'primary-alt', 'destructive', 'neutral', 'secondary', 'tertiary'];
-
-export function validateButtonColor(btnColor: string) {
-    if (supportedColors.indexOf(btnColor) < 0) {
-        throw Error('Unsupported color input value: ' + btnColor);
-    }
-}
+export type ButtonStyle = 'primary' | 'primary-alt' | 'destructive' | 'neutral' | 'secondary' | 'link' | 'link-inline';
```
jtrujill: We're going to need to keep the old stuff and update the list of valid styles. This is also causing a test to fail. The buttonStyle is fine and the reason that was removed was that dgeni couldn't generate docs for it

projects/cashmere/src/lib/sass/_typography.scss
```diff
@@ -107,17 +103,72 @@ a {
...
 .hc-type-bold {
     font-weight: 600;
 }
 
+.hc-type-xbold {
+    font-weight: 700;
+}
+
+// case
+.hc-type-titlecase {
```
jtrujill: scope case to `hc-text-*`

projects/cashmere/src/lib/sass/_typography.scss
```diff
@@ -107,17 +103,72 @@ a {
...
+
+.hc-type-uppercase {
+    text-transform: uppercase;
+}
+
+.hc-type-lowercase {
+    text-transform: lowercase;
+}
+
+.hc-type-ellipsis {
```
jtrujill: scope ellipsis to hc-text-nowrap

projects/cashmere/src/lib/button/button.component.spec.ts
```diff
@@ -55,15 +55,15 @@ describe('ButtonComponent', () => {
...
 
-    it('should apply class based on color property', () => {
+    it('should apply class based on buttonStyle property', () => {
         const testApp = new TestAppReference();
-        const buttonColors = ['primary', 'primary-alt', 'destructive', 'neutral', 'secondary', 'tertiary'];
+        const buttonStyles: ButtonStyle[] = ['primary', 'primary-alt', 'destructive', 'neutral', 'secondary', 'link', 'link-inline'];
 
-        buttonColors.forEach(color => {
-            testApp.setColor(color);
+        buttonStyles.forEach(style => {
```
jtrujill: Because we are deprecated but not removed setColor we can't remove those tests. You'll need to test both methods to make sure they work. It's okay to change comments but wherever you see a setColor you'll need a setStyle also.

corykon: fixed :+1:

corykon:
hey folks, looking for a bit more input on this PR again. @jtrujill suggested we be more specific with the new typography classes (see his comments above). perhaps we should establish a naming standard for new classes moving forward, at least in the case of those sass files/classes we expect users of the library to implement directly. thoughts? @andrew-frueh @benjanderson @jacquie

andrew-frueh:
@corykon @jtrujill - any suggestions on what those naming standards might look like for our sass classes?  I'm good establishing a standard - that honestly will make everyone's life easier because you don't have to so hard about what to call your classes.  To your point, Cory, applying those standards back to all of our component stylesheets as well might be a pretty big job, so I agree that it probably makes sense to start with the standalone sass docs.

andrew-frueh:
...and it occurs to me that if we do establish naming standards, we probably should add a guide for that.  Or at least add a section to the "Getting Started" guide.

jtrujill:
@andrew-frueh What I was thinking was that rather than throwing everything under the namespace of typography we could break things out into namespaces that the style has an effect on. 

For example, the font sizes effect the font size, so we would place those under `hc-font-{xs,sm,md,lg,xlg}`

The different cases have a css equivalent [text-transform](https://www.w3schools.com/cssref/pr_text_text-transform.asp), so the following makes sense: hc-`text-{uppercase,lowercase,capitalize}`

The ellipsis does a transformation on text so it could be `hc-text-nowrap`

The weight stuff effects font so it could be `hc-font-weight-{light,reg,bold,xbold}

The idea is that we should keep these scopes organized rather than throwing all of them under one scope. We should plan for the future to avoid name conflicts and possibly confusingly long names

jtrujill:
We should try to follow the css standard as much as possible to make using our classes easier to remember. They should be so easy it's almost intuitive and doesn't require looking up.

andrew-frueh:
Thanks for the additional explanation @jtrujill - that all makes sense to me, and I'm good with that approach.  @corykon @benjanderson @jacquie - any concerns?

jacquie:
I guess if we were going to get serious about naming standards we could look into one like BEM [http://getbem.com/introduction/], but I like the ideas above.

andrew-frueh:
Oh that's cool, @jacquie - I hadn't seen BEM before.  That might be even better because it's already an established syntax standard so we don't have to make up the rules as we go.

jacquie:
There's just a lot of hyphens to remember to get in the right place! :)

corykon:
BEM's got some really good concepts, but I feel like it might be more than we need right now, and might make contributing a little intimidating. I think I'm in favor of a simpler approach

andrew-frueh:
Okay, that makes sense @corykon.  Let's stick with the approach that @jtrujill is proposing for now.  Just so we can all keep our heads straight, let's make sure and document our naming sass naming conventions.  

As I think of it, we should probably have a separate guide called "Naming Conventions" which can document these sass standards, as well as anything contributors need to know about naming their components, etc.  I can create an issue for that unless anyone has any problems with adding a guide.

andrew-frueh:
Updated commit looks good @corykon

PR #405, opened 2018-08-07
docs(readme.md): adds changelog to the npm readme

jtrujill:


jtrujill:
Just realized that the latest version of Cashmere doesn't have a readme, so this PR will need to fix that issue

PR #410, opened 2018-08-08
feat(table): adds table component

### LINKED ISSUE #79: Table
Translate @corykon's incredible work on [table styling](https://github.com/HealthCatalyst/Fabric.Cashmere/pull/73) into a full Cashmere component.

- [x] Sorting (client side)
- [x] Paging (Client side) (toggle between more results button & numbered pagination)
- [ ] Toggle to choose page size
- [ ] Bake in checkboxes (input row type extend some `IsCheckable` class that has an `isChecked` property?)
- [ ] Input Table Options (total number of pages (server side pagination) rows per page, server/client side pagination, checkboxes/selectable rows)
- [ ] Table Output events (current page, selected rows, sortEvents for server side pagination)
- [ ] Documentation
- [ ] Tests

jtrujill:
adds table component along with an ability to sort and filter data with a datasource

close #79


jtrujill:
This is a conversion of the table from material. I'll need to move the stying into the table component itself. For now the styles are being applied through the sass/table.scss file. We'll need to change the old table.scss selector so that it doesn't conflict with the angular component version. We'll also want to deprecate that file

jtrujill:
What I could do is enable ViewEncapsulation until we have that sass/table.scss removed

andrew-frueh:
Wow, this one has been a long time coming.  Very exciting to see it finally moving into the library.  Two thoughts:
- Can we include sorting in the example, just so that all the table functionality is visible?
- Regarding deprecating the current table styles doc, there has been a lot of debate about how to build this component because we have some pretty custom tables out there in our apps.  Are we confident that this component is going to be flexible enough for those edge cases?  Or should we keep the stylesheet around for apps that need something really custom.  @andrewreiland @waldnercharles - I know CORUS is one of those apps with a really custom table, would be curious to hear your thoughts.

jtrujill:
Yep I'm going to get a sort example out. I would need to be shown what the different tables look like before I can answer that question. 

@andrew-frueh if you could show me some examples of what the different tables look like I could better answer that. This table produces actual html5 table elements, so if people are creating tables which follow that html standard for tables it should cover most things.

I'll contact Corus directly to see what they have

andrew-frueh:
Thanks a including the sorting example.

Thanks also for reaching out to CORUS.  My understanding is that they use some of the most custom tables, so my guess is that if we can accommodate that app, we'd probably be good.  But @benjanderson @jonest - maybe you guys can speak to how Atlas and Ops Console are doing their tables and whether there would be any issues for you with this component.

jtrujill:
After talking with Corus, it seems the table will cover most cases that they need. I'll need to get a selectionMode with the following options: none, single, multiple. I'll also need to get pagination implemented, but once I have those two things Corus said that they'd be good to use the table.

jtrujill:
@andrew-frueh Just pushed two more examples for sorting and filtering.

PR #432, opened 2018-08-10
fix(radio): remove circular dependency

### LINKED ISSUE #237: circular dependency in radio button component
When I compile I get the following error

`WARNING in Circular dependency detected:
lib\src\radio-button\radio-button.component.ts -> lib\src\radio-button\radio-group.directive.ts -> lib\src\radio-button\radio-button.component.ts`

And looking at both these components there is a circular dependency. This should be factored out into a third component or removed. 

This was introduced last week in #213

andrew-frueh:
Removed radio-group attributes from radio-button. The group now subscribes to events emitted by the individual buttons.  All the tests pass and the examples work, but I'd feel better if those of you with engineering skills greater than mine could check my work.

closes #237

### DIFF
--- a/projects/cashmere/src/lib/radio-button/index.ts
+++ b/projects/cashmere/src/lib/radio-button/index.ts
@@ -1,3 +1,3 @@
-export {RadioButtonComponent} from './radio-button.component';
+export {RadioButtonComponent, RadioButtonChangeEvent} from './radio-button.component';
 export {RadioButtonModule} from './radio-button.module';
 export {RadioGroupDirective} from './radio-group.directive';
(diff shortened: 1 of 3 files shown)


jtrujill:
@andrew-frueh I have an easier solution. With forwardRef it really only works when the classes are in the same file. It's really difficult to write a radio button with this functionality without there them being tightly integrated. From your PR, it removes functionality that we'd need, and unfortunately there are almost no tests written for this component. I have a [branch](https://github.com/jtrujill/Fabric.Cashmere/blob/fix-radio-circular/projects/cashmere/src/lib/radio-button/radio.ts) that has the fix and I've added the same comments that you have to the RadioButtonChangeEvent

PR #434, opened 2018-08-10
feat(accordion): toolbar trigger

### LINKED ISSUE #422: Accordion content area should trigger menu
**Describe the bug**
Expected behavior for accordion menus is that the content area (the text next to the chevron) will trigger the menu toggle (similar to radio buttons or checkboxes).  Currently only the bounding box for the chevron triggers the menu toggle.

**To Reproduce**
1.  Go to https://cashmere.healthcatalyst.net/components/accordion/examples
2.  Click on the blue chevron (menu toggles)
3.  Click on the text area for "Toolbar" (menu does not toggle)

**Browser:**
*   Firefox, Chrome

**Additional context**
I recognize there are certain situations where the developer won't want the content to trigger the menu, but those are the exceptions.  I think the default behavior should be for the content to fire the toggle function.

andrew-frueh:
adds ability to specify if entire toolbar is clickable or just the arrow; also updates the example to demo the capability

closes #422

### DIFF
--- a/projects/cashmere-examples/src/lib/accordion-overview/accordion-overview-example.html
+++ b/projects/cashmere-examples/src/lib/accordion-overview/accordion-overview-example.html
@@ -1,5 +1,5 @@
 <div class="sample-container">
-    <hc-accordion [triggerAlign]="alignment" [hideToolbar]="hideToolbar" #accordion>
+    <hc-accordion [triggerAlign]="alignment" [hideToolbar]="hideToolbar" [toolbarTrigger]="triggerToolbar" #accordion>
         <hc-accordion-toolbar>Toolbar</hc-accordion-toolbar>
         Content
     </hc-accordion>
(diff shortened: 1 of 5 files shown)


jtrujill:
@andrew-frueh should it always be the case that the accordion toolbar area be clickable? In the issue you say the expected behavior is to always have that toolbar be clickable

andrew-frueh:
Tough to say *always*.  I think it will be the case 90% of the time, but the only thing I can think of would be if the text on the accordion toolbar is a link or something like that.  So I just didn't want to paint us into a corner in case there are situations that need that area capable of catching clicks.

jtrujill:
I never even thought about that. The name is a little confusing but in another PR we could move that property to the toolbar itself.

I've been meaning to do a quick revamp of the accordion. What do you think about an accordion-group that handles the case where you only want one accordion open at a time

jtrujill:
Do you think triggerToolbar is more clear?

andrew-frueh:
Yeah - the name probably isn't ideal, but I couldn't come up with something better.
Regarding accordion groups, yes, that would be **very** useful!  I can think of several prototypes ([like this one](https://xd.adobe.com/view/aa5ba64f-9790-441a-8ad1-f19f3407966e/screen/0766ad30-2a51-4fc7-a3cb-e4197b082e98/Review-Detail-Data-Quality)) that call for that kind of functionality.

jtrujill:
Ignore the comment about the new. Let's get your work in

PR #437, opened 2018-08-10
fix(input): disabled state css

andrew-frueh:
Adds styling to make the input's disabled state more obvious.  Matches the disabled state of the select component.

### DIFF
--- a/projects/cashmere/src/lib/input/hc-form-field.component.scss
+++ b/projects/cashmere/src/lib/input/hc-form-field.component.scss
@@ -118,7 +118,16 @@ $wrapper-padding-bottom: ($error-margin-top + $line-height) * $error-font-scale;
 }
 
 .hc-form-field-disabled {
-    cursor: not-allowed;
+    .hc-form-field-flex {
+        cursor: not-allowed;
+        border: 1.5px solid $gray-200;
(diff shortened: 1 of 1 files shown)


projects/cashmere/src/lib/input/hc-form-field.component.scss
```diff
@@ -118,7 +118,16 @@ $wrapper-padding-bottom: ($error-margin-top + $line-height) * $error-font-scale;
...
 .hc-form-field-disabled {
-    cursor: not-allowed;
+    .hc-form-field-flex {
+        cursor: not-allowed;
+        border: 1.5px solid $gray-200;
+        background-color: $slate-gray-100;
+        color: tint($text, 60%);
+    }
+
+    .hc-input {
```
jtrujill: styling specific to hc-input needs to be moved to input.scss. form-field is meant to be used for multiple elements

andrew-frueh: good call - change made

andrew-frueh:
@jtrujill - you good with the updates to this PR?

jtrujill:
Sure. I'll need to move some scss stuff around, but I'll do that when I get to moving the hc-select to using form-field

PR #438, opened 2018-08-11
fix(tabs): remove multiline

### LINKED ISSUE #257: Tabs title multi-line?
I think that it looks odd to have the tabs title be multiline since it adds a huge margin to the bottom of all the other tabs. I'm thinking that we shouldn't let text wrap. Here's an example of what it looks like:
2018-05-20 23_25_17-claims quality explorer

andrew-frueh:
Removes text wrap from tab titles and truncates if the title is too long for the space (thanks for the `hc-text-ellipsis` class, @corykon) .  I agree with @jtrujill that for layout purposes, the tabs shouldn't go multi-line.  Obviously for the best UX, tabs shouldn't normally be truncated, but this gives a way for them to compress elegantly when screen real estate is limited.  See example:

screen shot 2018-08-10 at 6 49 00 pm

closes #257

### DIFF
--- a/projects/cashmere/src/lib/sass/_tabs.scss
+++ b/projects/cashmere/src/lib/sass/_tabs.scss
@@ -47,6 +47,10 @@
     border-bottom: 1px solid $slate-gray-300;
 }
 
+.hc-tab-width {
+    min-width: 0;
+}
+
 .hc-tab-horizontal {
     background-color: inherit;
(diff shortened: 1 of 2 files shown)


jtrujill:
Good start on this. I'd like to expand on this component so that you can do more than just a string but until I can get to that this is good

PR #442, opened 2018-08-11
feat(drawer): adds a dark themed drawer menu

### LINKED ISSUE #441: Add a new dark themed drawer
We need a dark themed drawer which provides a dark menu for a user. A prototype can be found [here](https://xd.adobe.com/spec/c51e0c10-290a-4f15-7845-2e818ee9951e-b32f/screen/fbc3492a-44bb-4b59-a125-76e0ed659234/Dashboard-Filters-Added/)

jtrujill:
close #441

### DIFF
--- a/projects/cashmere-examples/src/lib/drawer-menu/drawer-menu-example.css
+++ b/projects/cashmere-examples/src/lib/drawer-menu/drawer-menu-example.css
@@ -1,0 +1,20 @@
+.drawer-content {
+    height: 200px; /* used to show menu scrolling effect*/
+}
+
+.drawer-label {
+    margin: 0 50px 0 0;
+}
+
+.fa-plus {
+    color: #00aeff;
+    margin: 0 0 0 auto;
+}
+
+.hc-drawer-right .drawer-label {
+    margin: 0 0 0 50px;
+}
+
+.hc-drawer-right .fa-plus {
+    margin: 0 auto 0 0;
+}
--- a/projects/cashmere-examples/src/lib/drawer-menu/drawer-menu-example.html
+++ b/projects/cashmere-examples/src/lib/drawer-menu/drawer-menu-example.html
@@ -1,0 +1,67 @@
+<div class="drawer-container">
+    <hc-drawer-container>
+        <hc-menu-drawer align="left" menuTheme="dark-theme" #leftDrawer>
+            <div hcDrawerToolbar>Filter Sets</div>
+            <hc-drawer-item>
+                <div class="drawer-label">
+                    Clinical Program
+                </div>
+                <hc-icon fontSet="fa" fontIcon="fa-plus"></hc-icon>
+            </hc-drawer-item>
+            <hc-drawer-item>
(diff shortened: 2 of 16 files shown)


jtrujill:
I plan for this to be a general purpose drawer menu that can be styled slightly different. I know of at least one other menu style that I can do, which is the menu style used for the vertical tab set

projects/cashmere-examples/src/lib/drawer-filter-menu/drawer-filter-menu-example.html
```diff
@@ -0,0 +1,30 @@
...
+        <hc-drawer align="right" #rightDrawer>
+            Right drawer content
+        </hc-drawer>
+
+
+        <div style="height:200px;">
+            <button hc-button buttonStyle="primary" (click)="leftDrawer.toggle()" class="demo-button">
+                Toggle Left Drawer
+            </button>
+            <button hc-button buttonStyle="primary" (click)="rightDrawer.toggle()" class="demo-button">
```
andrew-frueh: I might just be missing something, but what are we demonstrating with the `rightDrawer` in this example?  The left drawer looks great, though!  Excited to see an example of a filter menu using Cashmere components.

jtrujill: I was wanting to show that hc-menu-drawer can be mixed with a hc-drawer. I wanted to show that hc-menu-drawer is nothing more than an extension of hc-drawer.

jtrujill:
@andrew-frueh Should we even include the icon for this drawer element? If we do it'd restrict this element to only one use-case, but maybe we could expand it's use beyond just a filter drawer. It'd be easy for the user to just include a font-awesome icon and have that get projected along with the rest of the menu item using the hc-drawer-item component.

Should we also at some point think about expanding the menu drawer so that it allows a user to define a sub-menu for each main menu item? I was thinking it'd be cool to have an accordion effect that opens the sub-menu up and when you click on another main menu item it closes the currently active sub-menu and opens another.

jtrujill:
Should we also include a toolbar if the user didn't define one with hcDrawerToolbar? It'd be easy to have the toolbar only shown when the user has included the directive on some element

andrew-frueh:
great point on the icon @jtrujill - I didn't even notice until you pointed it out that was included as a part of `hc-menu-drawer`.  I agree that we would do better to break that out and have it projected with the other content.  That gives more flexibility.

The sub-menu idea is also a great one.  I haven't had that come up in any of my prototypes up until now, so I wouldn't make it a super high priority, but I could see it being useful in the future.

And as far as the toolbar goes, I think I've some version of that in all my designs, but for maximum configuration, it probably makes sense to only show it if the user included the directive.  If a user is leveraging this component for something other than a filter drawer, they'd probably want the ability to omit that.

jtrujill:
@andrew-frueh I've made the change so the user has to include the icon themselves and also if a toolbar isn't included than there isn't space taken up by it

PR #443, opened 2018-08-12
feat(button): adds icon button

### LINKED ISSUE #228: Feature request: Icon buttons
UI libraries often have two categories of buttons: regular buttons and icon buttons.

- Regular buttons are rectangular, whereas icon buttons are round
- Icon buttons may or may not have a visible background color or borders
- Icon buttons have very little padding
- On hover, the icon itself will often change color

This is something that shows up frequently in design specs:

image

It would be great to have this built into Cashmere. I recommend using the `[hc-icon-button]` attribute as opposed to (or in addition to) the `[hc-button]` attribute, and making it clear in the documentation that this type of button should have an `hc-icon` as its only child.

jtrujill:
close #228

### DIFF
--- a/projects/cashmere-examples/src/lib/button-icon/button-icon-example.css
+++ b/projects/cashmere-examples/src/lib/button-icon/button-icon-example.css
@@ -1,5 +1,17 @@
-button {
-    margin: 10px;
+h3 {
+    margin: 12px 0;
+}
+
+button:not(:last-child) {
+    margin-right: 10px;
+}
+
+.btn-row {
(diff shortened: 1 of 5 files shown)


jtrujill:
@andrew-frueh should I give a `border-radius: 50%` to give it a round background? I think that if the user decided to apply a background color it'd look better if the button was round

jtrujill:
Also I moved the icon-button to the icons demo. Didn't make sense to separate it into its own page and instead seemed to fit better with the button with icons demo

PR #449, opened 2018-08-15
fix(app-switcher): remove erroneous greater than sign

jonest:
Remove erroneous greater than sign that snuck in. Remove dark blue background on hover and focus.


jonest:
image

PR #456, opened 2018-08-16
 style(about): about modal 

### LINKED ISSUE #107: About Dialog - Guide
The discussion #105 Marla started today made me realize that we should really make the About box triggered from the Help menu a standard component.  It would probably build off the #40 modal dialog component.  But this is something every app is going to need, and for support purposes it's important that the all look the same and contain the same information in the same places.

We developed a pretty nice one for Atlas that @joeskeen did a great job of building.  So we could probably use that as the foundation.  I'm not sure all the content on the Atlas version will be relevent to every app though.  So I'd be curious to hear from others about what standard information should be included on all About modals (and each app could add additional info as well).

Looking at this mockup again, I'm realizing this could be another good opportunity to put @jtrujill's hc-list to use.  It already is set up to do icons on the left and multiple lines of content on the right.

atlasabout

andrew-frueh:
adds a style section and associated style sheet for app modals

closes #107


guides/styles/about-modal.md
```diff
@@ -0,0 +1,72 @@
...
+        <div>
+            <hc-icon fontSet="fa" fontIcon="fa-database" icon-lg></hc-icon>
+            <div>
+                <strong>Database Name</strong>
+                <div class="about-reference-detail">EDWAdmin</div>
+            </div>
+        </div>
+    </div>
+    <div class="about-footer">
+        Copyright 2018 <a href="https://www.healthcatalyst.com/">Health Catalyst</a>. All rights reserved
```
nickrflynn: I don't think it's a problem here but I'll just note that internet access where an application is deployed is not a given.  In some cases a user will not actually be able to access these links on the machine he or she is using.

Issue #455, opened 2018-08-16
Sorting of nulls feels backward in default DataSource sort


corykon:
**Describe the bug**
When sorting A-Z, the order ends ups like this
1. Nulls
2. 0-9
3. A-Z

**To Reproduce**
1.  Go to [https://stackblitz.com/edit/cashmere-demo-sort-nulls](https://stackblitz.com/edit/cashmere-demo-sort-nulls)
2.  Click on the "name" header

**Expected behavior**
Might just be me, but I'd prefer the default to be like this:
1. 0-9
2. A-Z
3. Nulls

...In PopBuilder, we'd actually always just sorted nulls to the bottom regardless of sort direction, but that might not be the right way to go. **Other thoughts?**

**Browser:**
All

andrew-frueh:
Sorting nulls to the bottom would be the behavior I would expect also.

jtrujill:
It doesn't make sense when you think about it. Why when in ascending order should null be greater than all other values? There is no value so it should come first.

jtrujill:
Doesn't that seem odd that descending order would give you nulls at the top and ascending would give you nulls at the bottom?

jtrujill:
@corykon the provided demo seems to be blank

jtrujill:
If this behavior other than what the default it needed sortData can be overriden on the DataSource

jtrujill:
We should keep what logically makes sense. When I think about this you probably never want to actually show 'null' to a user and instead would want to replace that with some string value, which would render this PR nil

andrew-frueh:
Okay - I'll yield on this one.  Your argument makes sense to me.  And if we've learned anything from our past history with a table component, the way every team handles their data is unique.

But I'd really like to hear from @jacquie @benjanderson and @corykon to make sure we have a consensus before we close this issue/PR.

corykon:
I'm fine if we close it, I can just override.

PR #457, opened 2018-08-16
fix(table): fix sorting of nulls

corykon:
455


jtrujill:
@corykon can you edit your commit so the bottom line reads as `fixes #455` the changelog won't link to the issue if you don't have that pound sign

jtrujill:
@corykon before you get this merged in I want to talk about this in #455

jtrujill:
I'll provide some examples of what the original sort does and what the modified version does to sorting.

The `Data Type/String` column provides a label for the datatype for the `value` column or it's an element from the periodic table. For example a row with `Data Type = null` is going to have a null value in the `value` column.

Here's the original in ascending order:
original_asc

Here's the result of the PR in ascending:
mod_asc

PR #466, opened 2018-08-17
fix(select): fixes forms issue with select component

### LINKED ISSUE #464: Reactive Forms cannot set value on hc-select
**Describe the bug**
hc-select does not allow reactive forms to set the value of the underlying select element.

**To Reproduce**
The bug is demonstrated here! Yay!
https://stackblitz.com/edit/cashmere-demo-4erszy?file=app%2Fhome%2Fhome.component.html

**Expected behavior**
We want hc-select to be able to receive values from setValue just like select.

jtrujill:
fixes #464

### DIFF
--- a/projects/cashmere-examples/src/lib/select-forms/select-forms-example.css
+++ b/projects/cashmere-examples/src/lib/select-forms/select-forms-example.css
@@ -1,0 +1,11 @@
+form {
+    max-width: 350px;
+}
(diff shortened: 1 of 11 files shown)


jtrujill:
@andrew-frueh ha no I just like greasy food that'll give me a heart attack. Screw living to 80!

PR #469, opened 2018-08-22
docs(naming): adds naming convention guide

### LINKED ISSUE #411: Naming Conventions Guide
As per the discussion that was started in #392, we need to introduce at least some lightweight naming conventions around our sass classes.  @jtrujill proposed a structure that @corykon followed for the typography additions in that PR.  To keep everyone on the same page, we should have a guide that explains our naming conventions both for sass and for components.

Add an additional markdown file to our guides that describes naming conventions for contributors to follow when adding styles or components.

Could be added to the "Getting Started" guide, but I feel like this is really its own subject.

** If any, which components will have to change, and how **
It would probably worth going back through and auditing our existing styles and components to ensure that they are following the conventions outlined in this guide, but that could be a separate PR.

andrew-frueh:
I'm just the scribe on this one.  Turned the conversation from the thread into a markdown file for our naming conventions guide.  Let me know if any of the content is unclear or needs to change.

@jtrujill - I know you said you are opposed to the "Component" or "Directive" in the component names.  I still included it in this guide because that's how everything is currently named in the library, so I don't want to create confusion.  Seems like that change should be submitted as an issue to wholesale remove those labels from the components in the library.  And then if we do that we can change the guide to match.

closes #411

### DIFF
--- a/guides/naming-conventions.md
+++ b/guides/naming-conventions.md
@@ -1,0 +1,54 @@
+# Source Code Naming Conventions
+
+###### Last updated August 22, 2018
+
+:::
+
+##### Overview
+
+For the highest degree of consistency across the library, the naming conventions below should be followed when creating new components or classes. This will reduce the effort needed to read and understand code from multiple contributors to the library.
+:::
+
+:::
+
+##### CSS Classes
+
+The following general format should be followed when naming classes:
+
+`{company-scope}-{block-modifier}-{element}`
+
+*   {company scope} scopes all of our css to our company
+*   {block modifier} is an element that can standalone within our company scope
+*   {element} is element that can't standalone and requires the block modifier
+
+&nbsp;
+
+Example class name: `.hc-form-field-label-wrapper`
+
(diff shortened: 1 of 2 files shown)


jtrujill:
I don't think k this should be included in the guide section of the user-guide app. This is intended for developers of Cashmere and not users of  the library. It's confusing and seems like we're trying to push our standards on users of Cashmere. Instead this should be linked to through the contribution markdown file

andrew-frueh:
So you want this content at the end of `contribution-guide.md` instead of in its own file?

jtrujill:
Yeah and remove the double colon stuff. It isn't valid markdown. Then at the top add a link to it

andrew-frueh:
The double colons create separate tiles though - which breaks up the page visually so it's easier to scan.

jtrujill:
I didn't realize that that wasn't at the root of the project. Go ahead and merge this and we can look at this later

Issue #451, opened 2018-08-16
Paginator needs a pageSize input


jtrujill:
**Describe the problem this feature request is solving**
The paginator doesn't really allow for you to chunk data. I can't get information from the paginator about how many elements I should grab starting at position y. The only information I have is that initial starting position. 

**Describe the solution you'd like**
I need the ability to be able to set a pageSize or the number of elements that should be shown in the paginator. Without this ability I can't effectively split the data since I don't know how much I can show.

**If any, which components will have to change, and how**
Paginator will need to add a pageSize input and it should probably also add the ability to modify the pageSize since this would be a common use-case.

jtrujill:
Yes, the drop-down would be included be default and you set a pageSize. I need this for a table element. I need to be able to grab some number of elements and I need this built into the paginator. It doesn't make sense to have pagination info split between the table and paginator.

The way it'll work is that I'll grab a reference to the paginator on the page and pass that to the table. The table can then read the current page and the pageSize to determine what elements to grab from a HcTableDataSource or array.

jtrujill:
The user could hide this pageSize selector with an additional input called hide pageSelector or something. This is such a common case that it should be native to the pagination element. It's also necessary to have that pageSize defined on the paginator so the integration with hc-table isn't goofy

PR #470, opened 2018-08-23
451 paginator updates

### LINKED ISSUE #316: Pagination event is raised twice on each click
**Describe the bug**
Event is raised twice when paging. See demo below.

**To Reproduce**
https://cashmere-demo-zkqvml.stackblitz.io
Open the console and you'll see each paging gets logged to the console twice.

**Expected behavior**
I'd expect it to only trigger the event once per click, to avoid making double the requests.

**Screenshots**

**Browser:**

*   Chrome - Version 67.0.3396.62 (Official Build) (64-bit)

### LINKED ISSUE #445: Pagination pageNumber binding throws ExpressionChangedAfterItHasBeenCheckedError
**Describe the bug**
Pagination seems to be setting the page number to `null` before the view is rendered, throwing `ExpressionChangedAfterItHasBeenCheckedError`.

**To Reproduce**
https://stackblitz.com/edit/cashmere-demo-g5nco1?file=app/home/home.component.ts

**Expected behavior**
I expect it to not throw an error 😛 

**Screenshots**
image

corykon:
Adds page size selector to paginator and tightens up some styles. This does have breaking changes, but should allow for us to have this component better interface with the table component.

#451, fixes #445, fixes #316

### DIFF
--- a/projects/cashmere/src/lib/button/button.component.scss
+++ b/projects/cashmere/src/lib/button/button.component.scss
@@ -82,7 +82,7 @@ $btn-icon-sz: 18px;
     &.hc-secondary {
         @include colorButton($gray-100, 5%);
         color: $gray-600;
-        border: 2px solid $gray-300;
+        border: 1.5px solid $gray-300;
         height: 37px;
         line-height: 33px;
         padding-right: 19px;
--- a/projects/cashmere/src/lib/pagination/pagination.component.html
+++ b/projects/cashmere/src/lib/pagination/pagination.component.html
@@ -1,14 +1,13 @@
 <div class="hc-pagination-container">
-    <div *ngIf="!hidePageSize" class="hc-font-xsm hc-page-detail-container">
-        <label>Items per page:</label>
-        <hc-select [value]="pageSize" (selectionChange)="_changePageSize($event.value)">
+    <div *ngIf="!hidePageSize" class="hc-font-sm hc-page-detail-container">
+        <label>Showing:</label>
+        <hc-select [value]="pageSize" (change)="_changePageSize($event.target.value)">
             <option *ngFor="let pageSizeOption of _displayedPageSizeOptions" [value]="pageSizeOption">
                 {{pageSizeOption}}
             </option>
         </hc-select>
+        <label>of <strong>{{length}}</strong> entries</label>
     </div>
-
-    <div class="hc-range-label">{{_rangeLabel}}</div>
 
     <div class="hc-page-btns-container">
         <button hc-button color="secondary" class="hc-page-button hc-left-button" (click)="_previousPage()" [disabled]="!totalPages || _isFirstPage">
@@ -20,7 +19,7 @@
                 class="hc-page-button hc-inner-button expanded-page-button"
                 *ngFor="let page of _visiblePages"
                 [disabled]="!page"
-                [class.currentPage]="page === pageIndex + 1"
+                [class.currentPage]="page === pageNumber"
                 (click)="_goToPage(page)">
             <span *ngIf="!!page">{{page}}</span>
             <hc-icon *ngIf="!page" fontSet="fa" fontIcon="fa-ellipsis-h"></hc-icon>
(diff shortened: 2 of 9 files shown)


jtrujill:
Could you explain why this is a breaking change. I was expecting for this to just be an addition to the component. We need to hold off on breaking changes; otherwise, it'll annoy developers who are currently using this

projects/cashmere/src/lib/button/button.component.scss
```diff
@@ -82,7 +82,7 @@ $btn-icon-sz: 18px;
     &.hc-secondary {
         @include colorButton($gray-100, 5%);
         color: $gray-600;
-        border: 2px solid $gray-300;
+        border: 1.5px solid $gray-300;
         height: 37px;
```
andrew-frueh: Thanks for updating this style as we discussed to match the stroke on the select box.  One thing to change though is to make `height: 36px` instead of 37px now.  You can check the example on the subnav component to see a primary and secondary button side by side.

projects/cashmere/src/lib/button/button.component.scss
```diff
@@ -82,7 +82,7 @@ $btn-icon-sz: 18px;
     &.hc-secondary {
         @include colorButton($gray-100, 5%);
         color: $gray-600;
-        border: 2px solid $gray-300;
+        border: 1.5px solid $gray-300;
```
jtrujill: Lets keep our PRs focused on a single component in the future. It made sense to include those other pagination fixes in this but hiding a button change doesn't make sense.

corykon:
@jtrujill the component used to require the user to input the numberOfPages, but with the page size being adjustable, it makes more sense for that to be calculated based on the number of items being paged through

projects/cashmere/src/lib/pagination/pagination.component.html
```diff
@@ -1,28 +1,42 @@
...
+        <hc-select [value]="pageSize" (change)="_changePageSize($event.target.value)">
+            <option *ngFor="let pageSizeOption of _displayedPageSizeOptions" [value]="pageSizeOption">
+                {{pageSizeOption}}
+            </option>
+        </hc-select>
+        <label>of <strong>{{length}}</strong> entries</label>
+    </div>
+
+    <div class="hc-page-btns-container">
+        <button hc-button color="secondary" class="hc-page-button hc-left-button" (click)="_previousPage()" [disabled]="!totalPages || _isFirstPage">
```
jtrujill: I didn't realize that the last paginator was relying on other cashmere components. I suggest that we keep other cashmere components out of our templates since a subtle change in one could drastically effect the style of another. This can be fixed in a future update

andrew-frueh: But then we are creating similar things that look and act the same way as each other but pull from different code.  So for example, if we change the styling of the button - it comes across here.  If this was something that just looked like the button, we'd have to remember to change its style as well when we updated the button.  And we wouldn't want to rewrite the select box for this component, no?

jtrujill:
@corykon is the numberOfPages change the only breaking change? Could you just map that to the new field and change the description of what it does and mark that as deprecated. I think most people were just passing the number of elements to that field anyways. I don't want to keep updating the major version for small changes like this

projects/cashmere/src/lib/pagination/pagination.component.html
jtrujill: Pagination buttons look nothing like an hc-button. I've been in projects that try to heavily reuse components and extend them where they are used by either changing the default functionality or style. If all of our components start using other components within well eventually run into major issues. By major issues I mean that you can run into cyclic issues. On top of that you have hidden dependencies that the user doesn't know about which could cause issues. Say that you have a drop-down that uses an hc-button and the user the client doesn't import hc-button, then you'll have issues.

Basically there are many reasons not to and few reasons to use cashmere components in other cashmere components templates

jtrujill: It doesn't have to be changed in this PR but we should have the understanding that in the future this will need to be changed and that future components should avoid using other cashmere components in their templates

PR #479, opened 2018-08-28
Set accordion currentlyAnimating prior to emitting

waldnercharles:
Simple change so that when we check the isOpened and isClosed getters in an emitted opened or closed event, they return true.


waldnercharles:
@jtrujill Done. Next time I'll read the contribution guide more thoroughly. 👍

PR #486, opened 2018-08-31
feat(pagination): add page size

### LINKED ISSUE #451: Paginator needs a pageSize input

### LINKED ISSUE #478: Texts get clipped from the bottom for select boxes
**Describe the bug**
While using <hc-select> the text inside the select boxes gets clipped from the bottom. 
For example letters like 'g' and 'y' does not appear fully.

image

image

corykon:
BREAKING CHANGE: changed input property for pagination component

closes #451, fixes #445, fixes #316, fixes #478

### DIFF
--- a/guides/naming-conventions.md
+++ b/guides/naming-conventions.md
@@ -1,0 +1,54 @@
+# Source Code Naming Conventions
+
+###### Last updated August 22, 2018
+
+:::
+
+##### Overview
+
+For the highest degree of consistency across the library, the naming conventions below should be followed when creating new components or classes. This will reduce the effort needed to read and understand code from multiple contributors to the library.
+:::
+
+:::
+
+##### CSS Classes
+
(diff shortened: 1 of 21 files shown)


corykon:
@jtrujill you'll have to tell me if it looks like I've done this correctly... I tried out the interactive rebase thing, squashed all my commits into one. it sort of looks like stuff I merged are just part of this PR.

Feel free to hop on this branch and monkey with it if needed

PR #487, opened 2018-08-31
feat(pagination): add page size

### LINKED ISSUE #316: Pagination event is raised twice on each click

### LINKED ISSUE #445: Pagination pageNumber binding throws ExpressionChangedAfterItHasBeenCheckedError

### LINKED ISSUE #451: Paginator needs a pageSize input

### LINKED ISSUE #478: Texts get clipped from the bottom for select boxes

jtrujill:
BREAKING CHANGE: replaced totalPages input with length input

closes #451, fixes #445, fixes #316, fixes #478

### DIFF
--- a/projects/cashmere/src/lib/pagination/pagination.component.ts
+++ b/projects/cashmere/src/lib/pagination/pagination.component.ts
@@ -1,4 +1,25 @@
-import {Component, EventEmitter, Input, OnChanges, Output, SimpleChanges} from '@angular/core';
+import {Component, EventEmitter, Input, Output, OnInit} from '@angular/core';
+import {coerceNumberProperty} from '@angular/cdk/coercion';
+
+const DEFAULT_PAGE_SIZE = 20;
+
+/**
+ * Change event object that is emitted when the user selects a
+ * different page size or navigates to another page.
+ */
+export class PageEvent {
+    /** The current page index */
+    pageNumber: number;
+
+    /** Index of the page that was selected previously */
+    previousPageNumber?: number;
+
+    /** The current page size */
+    pageSize: number;
+
+    /** The current total number of items being paged */
+    length: number;
+}
 
 /** The pagination control enables the user to navigate across paged content.
  * Although commonly used with tables and data grids, this control may be used any place where paged data is used.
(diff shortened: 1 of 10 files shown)


projects/cashmere/src/lib/pagination/pagination.component.ts
```diff
@@ -176,23 +204,62 @@ export class PaginationComponent implements OnChanges {
...
+        this._displayedPageSizeOptions = this.pageSizeOptions.slice();
+        if (this._displayedPageSizeOptions.indexOf(this.pageSize) === -1) {
+            this._displayedPageSizeOptions.push(this.pageSize);
+        }
+
+        // Sort the numbers using a number-specific sort function.
+        this._displayedPageSizeOptions.sort((a, b) => a - b);
+    }
+
+    private _coerceNumberValue(value: any): number {
```
jtrujill: There is a CDK method that we could use for this called coerceNumberProperty(). I'll see about switching to that

PR #489, opened 2018-09-10
refactor: updates to the latest @angular/cli and @angular/* packages

jtrujill:


guides/contribution-guide.md
```diff
@@ -19,6 +19,10 @@ We would love for you to contribute to Cashmere and be part of the community mak
 4.  Clone your fork.
     Recommendation: name your git remotes `upstream` for `@healthcatalyst/cashmere`
 5.  From the root of the project, run `npm install`.
+6.  Running `npm run build` will build the entire project
+7.  While developing you can use angular cli's `--watch` to automatically build a project when the source files change
```
benjanderson: It may be wise to mention they also need to run `ng serve` in addition to `npm run build`

package.json
```diff
@@ -34,44 +34,37 @@
...
-            "prettier --write",
-            "npm run lint --fix",
-            "git add"
-        ]
+        "*.{js,scss,css,md,json}": ["prettier --write", "git add"],
+        "*.{ts}": ["prettier --write", "npm run lint --fix", "git add"]
     },
     "dependencies": {
-        "@angular/animations": "6.0.3",
+        "@angular/animations": "6.1.7",
```
benjanderson: Nice!

projects/cashmere/src/lib/sort/sort.ts
```diff
@@ -45,7 +45,7 @@ export class HcSort implements OnChanges, OnDestroy, OnInit {
     sortables = new Map<string, HcSortable>();
 
     /** Used to notify any child components listening to state changes. */
-    readonly _stateChanges = new Subject<void>();
+    readonly _stateChanges = new Subject<never>();
```
benjanderson: Today I learned about the `never` type

guides/contribution-guide.md
jtrujill: I'll add that but I think that is getting into documenting `@angular/cli` itself. The reason i'd avoid putting that in here is that we shouldn't bloat our own docs with other tooling docs. If a dev doesn't know a tool then then can easily look up the docs to figure out how to use it.

jtrujill:
closed since travis isn't reporting build status

package.json
```diff
@@ -105,19 +98,19 @@
...
         "semantic-release": "^15.9.5",
         "ts-node": "^5.0.1",
-        "tsickle": ">=0.25.5",
-        "tslib": "^1.7.1",
+        "tsickle": ">=0.32.0",
+        "tslib": "^1.9.3",
         "tslint": "^5.8.0",
         "tslint-config-prettier": "^1.13.0",
-        "typescript": "2.7.2",
+        "typescript": "2.9.2",
```
benjanderson: Why use latest type script when angular explicitly states: `The Angular Compiler requires TypeScript >=2.7.2 and <2.10.0 but 3.0.3 was found instead.`

Issue #412, opened 2018-08-08
Change table sort icon


jtrujill:
**Describe the problem this feature request is solving**
We need to convert the current table sort icon to use a style similar to what is defined in fa-sort-amount-asc 

**Describe the solution you'd like**
I'd like to take a look at using fontello or using css tricks.

** If any, which components will have to change, and how **

PR #495, opened 2018-09-28
refactor(hc-table): update hc-table styles to match official designs

isaaclyman:
This PR proposes changes to the hc-table to bring it into alignment with @andrew-frueh's designs (https://xd.adobe.com/view/55c8964a-fe2f-41c3-4ede-3d1bbd4d5b71-23f9/screen/ea040736-2e21-46f1-a57e-20275e50b077/Expense-Region/). It also adds a footer row to the Example Overview Table in the documentation to better demonstrate the functionality and style of footer rows.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2397,14 +2397,6 @@
                 }
             }
         },
-        "@healthcatalyst/cashmere": {
-            "version": "2.2.2",
-            "resolved": "https://registry.npmjs.org/@healthcatalyst/cashmere/-/cashmere-2.2.2.tgz",
-            "integrity": "sha512-VhiO6ji8T6dNnSoUCi4yDLYLmNPuNf+xYugjACX/cUd7BrPLBpmEcYj5BPoyZXoBgSzpdqRVuN7vf3hSQwsecw==",
-            "requires": {
-                "tslib": "^1.9.0"
-            }
-        },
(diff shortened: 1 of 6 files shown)


andrew-frueh:
This looks great @isaaclyman - thanks for doing this.  Can you make one small addition to the css file?  I noticed on the "Styles" page for tables that the sorting indicator classes were getting their borders cut off in the header.  So on line 161 of tables.scss, in the hc-col-sortable class, can you add `background-clip: padding-box;`.  That will made sure the borders appear above the background even with position:relative turned on.

@corykon @benjanderson - some design thoughts on this.  Something I observed pretty much across the board in our user tests was that many users weren't entirely sure if they could sort a table based on the column with the styling I favored before (no borders around the header).  I think removing a border around the header made them think it was static, even if a sorting indicator was there.  So even though I prefer the no border look in the header row, I'm thinking that for usability - it's probably better to include it.  But of course we'll have to see if this change proves out with actual users.

andrew-frueh:
And just for anyone that was wondering, the table sorting indicator in the component still needs to be updated as per this issue: https://github.com/HealthCatalyst/Fabric.Cashmere/issues/412

isaaclyman:
@andrew-frueh Thanks for pointing that out, I've added a commit.

PR #500, opened 2018-10-08
Opt in examples

isaaclyman:
### This PR is not meant to be merged as-is. It is intended for discussion and documentation and hasn't been thoroughly tested.

See https://catalyst-finance.atlassian.net/browse/LM-167 or talk to one of the following for details:
- Isaac Lyman @isaaclyman 
- Aaron Johnson @?

These changes demonstrate how Cashmere could be converted from an opinionated styles-and-components library to a composable set of SASS mixins and components that allows consumers to opt in to specific styling rulesets and create their own components, or use the full components if said components meet their needs. This solves the problem of straightforward, official look and feel for all HC apps, without attempting to solve for every use case or design need that a developer may have.

The two components that have been converted in this example are hc-button and hc-checkbox. For each, a separate file has been created under the `sass/` folder, which exports a variety of mixins. Then the original component-owned sass file has been updated to import those mixins and implement them in the default way. Consumers can do any of the following, from least flexibility to most flexibility:

- Use the full Angular component
- Adopt the class names used by the Angular component on their own component (this is possible because the components don't use view encapsulation)
- Import the same mixins that the Angular component uses and implement them in their own component

Regardless of the method used, the component will have the correct look and feel for HC apps.


andrew-frueh:
Very cool @isaaclyman - thanks for putting this example together.  This solution feels in a way like having our cake and eating it too.  It allows our teams the autonomy to work with Cashmere in the way that is most effective for them, while ensuring everyone is still pulling from the same style standards.  It also takes some of the pressure off Cashmere to be perfect immediately.  We can continue to build and refine our components, and they'll continue to get better, but in the meantime, it's not going to handicap any of our development teams.

For those new to this conversation, in a nutshell, the Labor Management team has been using Cashmere for a bit, and have found that the components haven't always addressed their use cases, and working around them has slowed their development down.  But knowing that not every team has had the same experience, they proposed this middle ground where everyone can make use of Cashmere to varying degrees as Isaac described, depending on the nature of the product and their dev team.

It appears that this wouldn't introduce any breaking changes to those currently using the library.  Seems like the biggest hassle will be converting all the existing stylesheets over to this approach - but based on your example, it appears we can do that piecemeal.

So thumbs up from me.  I'd leave to hear from @benjanderson, @corykon, @jacquie, @jonest and @iamboo on any thoughts or concerns they have with an adjustment like this.

...and last is the $10 M question, @isaaclyman - is this stylesheet retrofitting process for the Cashmere components something that the finance team is interested in taking on (even if it takes several months)?

isaaclyman:
@andrew-frueh You're right, there shouldn't be any breaking changes. The challenges will be 1) conversion, and 2) making sure to reject pull requests that would diverge from the pattern and create a different experience for component users and style users.

As for my team taking on that work, I'm not in a position to answer that question. Although I'd be happy to pitch in and do a lot of it, starting this week or next I'll be on paternity, and until 2019 I probably won't have any extra time to contribute. I'll talk to Aaron and Brandon and see what kind of commitment we're willing to undertake.

corykon:
Personally, I love this approach! 

We shot down a similar idea in a steering committee meeting months ago, and I think it was for fear that teams would use just the styles and ignore the components, or try to lay them on top of non-angular apps. Maybe the tide is shifting? ...

benjanderson:
I love this idea, I think having a consumable style layer would increase adoption of cashmere. I also like making them mixins because any unused styles wouldn't be included in the final artifacts.

I know this pull request is just a concept, but I hope the final result would be more well architected to be flexible/scalable. For example in your button mixin, it would make sense to have color as a parameter.

isaaclyman:
@corykon that's what we want. It doesn't matter a lot if various teams create their own components or even use a different tech stack; the problem Cashmere exists to solve is a standardized look and feel among Health Catalyst apps.

@benjanderson thanks. There is a button mixin with a color parameter, so I'm not sure what you're referring to. Do you have other examples of how the architecture falls short? This is pretty close to what I'd expect the final version to look like.

benjanderson:
@isaaclyman I think I wrote my comment without fully explaining. If we are to move towards supporting a style library, and using cashmere through referencing css, I would love if key variables in each component were customizable.  

As an example, if you look at bootstrap, almost all the variables have `!default` after their definition. This allows you to further customize the look and feel of components.
https://github.com/twbs/bootstrap/blob/v4-dev/scss/_variables.scss

That said, I dont think it fits in the scope of this task and I think your code is a good first step. If we move towards supporting other platforms through referencing css I would open an issue where we could discuss it further.

isaaclyman:
@benjanderson allowing a lot of customization is a good move for a general-purpose library like Bootstrap, but seems to run counter to the goals of Cashmere, as far as helping the company achieve style consistency. But that may be for @andrew-frueh to make the call on. Making some variables available for override wouldn't be too difficult if we decide to go that route.

andrew-frueh:
Appreciate the discussion, everyone - and it sounds like we're quickly coming to a consensus.  I'd like to just get a thumbs up from @jacquie and @iamboo as well to make sure they don't have any concerns before we give the full green light.

Regarding the level of customization, it sounds like there is agreement that what Isaac has proposed is a good step, and we can continue to debate how much we allow for customization beyond this in the future.

@isaaclyman - given that your team would potentially be shouldering a decent chunk of the grunt work, how would you prefer this effort be organized on GitHub?  Should we make a separate issue for the conversion of each component?  They can then all be added to a GitHub project for easier tracking.  And it would also allow people outside your team to sign up and pick off conversions as they have time.  This is the way we did our conversion to dGeni and it worked pretty well.

isaaclyman:
@andrew-frueh Yes, let's create a GitHub label for this project and an issue for each component. The work isn't too difficult, but it's a great opportunity to organize the CSS a little and make some calls about what main pieces comprise each component. The first thing I'd like to do is write a guide to the process -- does Cashmere have a wiki I could use for that?

andrew-frueh:
Just to make sure everyone else is looped in - @isaaclyman and I were talking on Slack and he has some bandwidth this sprint to work on retrofitting the components.  We decided he'll add a markdown guide to our "Guides" section (instead of the wiki) for how users of the library can decide what level of Cashmere they'd like to use.  We'll also create issues for each component's update and GitHub project to track the progress.  Thank you Isaac!

iamboo:
@isaaclyman and @andrew-frueh - I like it!  This is a great use of mixins that allows much more flexibility in how the styles can be applied.

jacquie:
We on the OPM also agree this would be a good way to go!

jtrujill:
So my take on this is that if something like this is done we should go with a more bootstrap structure where we provide the styling through classes and requires the user to follow a specific html structure. With this current implementation it would require that the users of the mixins to define their own html/css structure and know the appropriate mixin to use, which could be very difficult with more complex components.

If we instead we provided sass files that require the user define their html in a specific form it would make things easier. If we are talking about making it easier for people to implement their own version off cashmere than it'd be best to pull all of these definitions into their own repo. This would also require that we scope the classes to something other than hc.

If we do that we should also probably scope cashmere to something different since hc is has nothing to do with cashmere.

isaaclyman:
@jtrujill I think we're already doing what you're suggesting. The goal is to provide styles and flexibility in how to use them, so consumers are welcome to use the mixins, the CSS classes, or the full component. If they want to see the HTML structure, they can look at the component markup (we could also document it, whatever turns out to be necessary). It might be possible to split the mixins into their own repo, but I'm very resistant to the idea because of how much it would slow down development work. Having to wait for a style change to propagate through two repositories and an NPM package...that doesn't sound good at all.

I'm not sure I understand what you're saying about `hc`, Cashmere is an HC project.

jtrujill:
The issue with hc is that it puts all of cashmere's styles into hc namespace. Instead it should have been scoped to to a cashmere specific space to avoid name clashes with other health catalyst projects. By using mixins there is a hidden predefined html structure that the user must follow but all they are given are mixins which doesn't help with understanding the required structure those mixins need.

It would be better to probably move the whole component's scss file over and remove any scss definitions that have to do with animations. If we hc-button as an example the only thing someone would need to do to get going is have the following:

```
<button class="hc-button hc-primary"></button>
<button class="hc-button hc-primary" [disabled]></button>
<button class="hc-button hc-destructive" [disabled]></button>
```

The user would just need to import the scss file and they are ready to go. With more complex components you'd see more benefit since you aren't needing to redefine the structure.

If needed we could just create another subproject within cashmere to house these styles. The only thing is that I don't know how well it would work with semantic-release. We could keep it nested within cashmere until the styling is more polished and move it to it's own repo, which shouldn't be an issue since styles don't change much

isaaclyman:
Again, I think what you're saying is what we're doing. People using Cashmere can import the mixins if they want to define their own structure, which is a super-flexible option. They can use the CSS classes (like in your example) if they want to do that. Or they can use the full component off the shelf. What you're saying is an option, but in cases where we need to be more flexible, the raw mixins are also available.

jtrujill:
My concern is that if they are required to look up the definition within cashmere to implement the mixins correctly we might as well just copy the whole scss file over.

We also need to make sure that any custom components look the same between different projects which is more difficult when mixins are used since it allows to much freedom in implementation. While freedom is great we need to ensure that the apps look similar

By providing a structure we speed up developers ability in implementing their own components while ensuring the they look the same as what the company has defined.

Is there a case where you see this not working?

jtrujill:
There's also the problem that if people have the freedom to do there own HTML structure, that when the mixins are updated with a different padding, margin or position that it breaks their component.

isaaclyman:
Thanks for the input. I'll keep your concerns in mind, but I'm going to proceed with the plan we agreed to as a team on Oct 2nd. I understand the importance of consistency and correctness but freedom of implementation will drive adoption and I see adoption as the single greatest factor in achieving the goals of the library.

jtrujill:
I'm not trying to argue but I feel I bring up valid points that haven't been brought up previously in this thread. Because Cashmere is used by many teams I'd like to get some input from other people before my concerns are pushed aside. @benjanderson @corykon @andrew-frueh @jacquie do you have any input on this.

Issue #483, opened 2018-08-30
hc-split-button style update


jtrujill:
**Describe the problem this feature request is solving**
Split button needs a style revamp without introducing breaking changes. We need to update the menu to follow the style guide.

**Describe the solution you'd like**
This would be a scss change only and wouldn't introduce any breaking changes

**Describe alternatives you've considered**
#453 - breaking change that can be done in the future

** If any, which components will have to change, and how **
split-button's styling is the only thing that'll need to be changed

jtrujill:
@corykon do you have the example hc-split-button menu that you can upload to this issue

corykon:
this one?

jtrujill:
Yes. Thank you

jtrujill:
The idea was to keep popover abstracted away from the user. I don't want to have a stronger dependency on popover. The easiest solution is to have an 'alignment' input that takes a left or right value.

PR #501, opened 2018-10-09
split button menu

corykon:
#483 adds a more robust split button menu. Replaces popper.js with the angular CDK overlay. Would love to get some feedback on this, and then see if we'd like to try swapping popper out in the navbar and popover components.


projects/cashmere/src/lib/button/split-button/split-button.component.ts
```diff
@@ -102,4 +127,54 @@ export class SplitButtonComponent {
             this.click.emit(new SplitButtonClickEvent(this));
         }
     }
+
+    _menuClicked() {
```
joeskeen: Is this supposed to be private? (Judging by the leading underscore?)

projects/cashmere/src/lib/button/split-button/split-button.component.spec.ts
```diff
@@ -20,29 +21,25 @@ describe('SplitButtonComponent', () => {
...
 
-        it('should not trigger primary button click when menu item is clicked', () => {
+        it('should not trigger primary button click when a menu item is clicked', () => {
             expect(testComponent.isPrimaryClicked).toBe(false);
             expect(testComponent.isMenuItemClicked).toBe(false);
 
             triggerButton.click();
             fixture.detectChanges();
 
+            let buttonItems = splitButtonDebugElement.queryAll(By.directive(ButtonItemDirective));
```
joeskeen: Very cool.  I have never used `By.directive` before!

package.json
```diff
@@ -1,129 +1,129 @@
...
+    },
+    "scripts": {
+        "ng": "ng",
+        "start": "ng serve",
+        "docs": "ts-node scripts/docs.ts && node scripts/generate-examples-docs.js && node scripts/generate-usage-docs.js",
+        "test": "ng test cashmere",
+        "test:unit": "ng test cashmere --watch=false --code-coverage --no-progress",
+        "lint": "npm run lint:tslint && npm run lint:prettier",
+        "lint:tslint": "tslint -p tsconfig.json",
+        "lint:prettier": "prettier --write '**/*.{ts,js,scss,css,md,json}'",
```
jtrujill: The problem with this is that instead of checking to see if there are lint issues the server will re-write the files with the proper formatting, but it won't commit those changes. Basically it'll allow people to commit code that doesn't match our linter

jtrujill:
This is an awesome job Cory!

projects/cashmere/src/lib/button/button.module.ts
```diff
@@ -2,13 +2,37 @@ import {NgModule} from '@angular/core';
...
 import {AnchorComponent} from './anchor.component';
-import {ButtonItemDirective} from './split-button/button-item.directive';
+import {ButtonItemDirective} from './split-button/directives/button-item.directive';
+import {MenuIconDirective} from './split-button/directives/menu-icon.directive';
+import {MenuTextDirective} from './split-button/directives/menu-text.directive';
+import {MenuSubTextDirective} from './split-button/directives/menu-sub-text.directive';
+import {DividerDirective} from './split-button/directives/divider.directive';
 import {SplitButtonComponent} from './split-button/split-button.component';
 import {PopoverModule} from '../popover/popover.module';
+import {OverlayModule} from '@angular/cdk/overlay';
```
benjanderson: very nice! I have been wanting to get rid of popper.js in other components. This is a great first step

package.json
joeskeen: That change was accidental.  Good catch.

joeskeen: fixed

benjanderson:
I pulled down the source and ran it locally. It works very well and a good start to using angular cdk. Great job!

PR #537, opened 2018-10-11
refactor(button): convert button to mixins

### LINKED ISSUE #511: (Opt-in) Convert the Button component

See https://github.com/HealthCatalyst/Fabric.Cashmere/pull/500/files

isaaclyman:
Is there a way to test this as if it were a published NPM package so I can document the process of using Cashmere via mixins?

Fixes #511

### DIFF
--- a/guides/using-and-customizing-components.md
+++ b/guides/using-and-customizing-components.md
@@ -14,7 +14,7 @@ You can use Cashmere components in a few different ways. They're built from thre
 1.  A set of CSS classes that import the mixins and implement them in a default way. Cashmere components don't use view encapsulation, so these CSS classes are global.
 1.  An Angular component, which includes HTML markup and basic functionality.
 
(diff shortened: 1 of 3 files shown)


jtrujill:
From my perspective this could be problematic with more complex components. Button is extremely simple and just has classes appended to the button.

With other components you're going to have structure. What I mean by that is that it'll require some HTML structuring and css to go with it for it to be styled correctly.

If this is supposed to be similar to bootstrap then we should provide styling through classes rather than mixins. Are there plans to pull this into it's own repo? Is this going to be discussed at the next cashmere meeting?

isaaclyman:
@jtrujill I haven't heard anything about trying to be like Bootstrap. We'll take it on a component-by-component basis, I'm sure some of them will be more difficult. Discussion about architecture is at https://github.com/HealthCatalyst/Fabric.Cashmere/pull/500 and https://github.com/HealthCatalyst/Fabric.Cashmere/pull/536.

PR #541, opened 2018-10-12
style(table): updated sorting indicators

### LINKED ISSUE #412: Change table sort icon

andrew-frueh:
Updates the table component sort headers to match our [Health Catalyst style](https://cashmere.healthcatalyst.net/styles/table).  Uses encoded svgs for the icons, so no dependency on FontAwesome.

closes #412

### DIFF
--- a/projects/cashmere-examples/src/lib/table-sort/table-sort-example.html
+++ b/projects/cashmere-examples/src/lib/table-sort/table-sort-example.html
@@ -8,13 +8,13 @@
 
     <!-- Name Column -->
     <ng-container hcColumnDef="name">
-        <th hc-header-cell *hcHeaderCellDef hc-sort-header>Name</th>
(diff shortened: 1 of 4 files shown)


isaaclyman:
Net -50 lines of code! I love to see that. I'm especially curious about what made sort-header.html so compact, it looks like we've been able to get rid of a lot of that logic.

andrew-frueh:
@isaaclyman a lot of the code in the Google version of that class looked to be supporting animated transitions of the up/down arrow.  Our implementation is a lot simpler, and honestly there is probably quite a bit that could be cleaned out of the typescript file now as well that we aren't using.  But I thought that might be a better job for a programmer than a designer :-)

PR #552, opened 2018-10-19
feat(utils): parseBoolean throws error

### LINKED ISSUE #551: parseBooleanAttribute doesn't throw an error
**Describe the bug**
@jonest pointed this out in #550 - that the utility function, parseBooleanAttribute, doesn't currently throw an error if a non-boolean value is passed into it (it just returns false)

**To Reproduce**
Pass a non-boolean value or string into parseBooleanAttribute.

**Expected behavior**
The function should return true or false if the value is either already a boolean value or is a string of "true" or "false" (regardless of the case).  The function throw an error if it is passed a non-boolean value.  However, an empty string should return true because if an input is added to tag without a value it implies true.

andrew-frueh:
parseBooleanAttribute now throws an error when passed non-boolean values.  Note that an empty string ("") returns true because when an input is added to a tag without a value it implies that the value should be true (ie `<input disabled>`).

closes #551


projects/cashmere/src/lib/util.ts
```diff
@@ -1,3 +1,11 @@
 export function parseBooleanAttribute(value: any): boolean {
-    return value != null && value.toString() !== 'false';
+    if (typeof value === 'boolean') {
+        return value;
+    } else if (value.toLowerCase() === 'false') {
+        return false;
+    } else if (value.toLowerCase() === 'true' || value === '') {
```
jonest: I'd expect that if it's not explicitly `'true'` or `'false'` that it'd throw an error.

What's the scenario you'd see where we would want `''` to return `true`?

jonest: I could maybe see that if it wasn't included it could default to `''` and then we'd just default to `false` maybe?

andrew-frueh: This is an interesting one because I originally thought the same as you.  But I realized when I made this update that this was throwing an error:
`<input hcInput placeholder="Enter username" disabled>`
...and it's because the component is using `parseBooleanAttribute` on the "disabled" value.  So it's correct syntax to write that as either `<input disabled>` or `<input disabled=true>` and it shouldn't be throwing an error.  And in the first example where the value isn't included, it's expected that the value is true.
So that's how I arrived where I did - open to suggestions though.

jonest: Interesting. That makes sense! Thanks for the info.

PR #553, opened 2018-10-19
style(app-switcher): thumbnail sizing

### LINKED ISSUE #503: Only width of icons is constrained in App Switcher
App switcher icons should be a consistent size.  Currently the width is constrained to 100px, but the height isn't constrained at all.  This means small icons have labels ride higher than square or tall icons.  Both width and height (for the image container) should be set to 100px for consistency.

**To Reproduce**
Reproduce it on stackblitz fork
https://stackblitz.com/edit/cashmere-demo-7lm9ah

**Expected behavior**
Each icon uniformly fitting within a square 100x100px container.

**Screenshots**
This issue occurs in internal Health Catalyst app(s), but as I can't share screenshots from internal apps on GitHub, here is an example of the issue from my Dominion app (blue borders around images added for emphasis):

image

**Browser:**

*   All

**Additional context**
It's important that the fix for this doesn't simply stretch out the image height, but instead constrain the max-height and vertically align the image within the 100x100px image container.

andrew-frueh:
app switcher thumbnails are now proportionally constrained in 100px box.  Switched from an `` to a `<div>` with a background-img.  This takes advantage of CSS' ability to proportionally resize an image within our bounding box so it won't be stretched.

closes #503

### DIFF
--- a/projects/cashmere/src/lib/app-switcher/app-switcher.component.html
+++ b/projects/cashmere/src/lib/app-switcher/app-switcher.component.html
@@ -15,7 +15,7 @@
         <div class="apps">
             <div class="app" *ngFor="let app of applications">
                 <a class="thumbnail" title="{{app.Description}}" href="{{app.ServiceUrl}}" target="_blank">
-                    <img class="popover-icon img-responsive" src={{app.Icon}}>
+                    <div class="thumbnail-img" style="background-image: url('{{app.Icon}}')"></div>
                     <div class="title">{{app?.FriendlyName | ellipsis:18}}</div>
                 </a>
             </div>
--- a/projects/cashmere/src/lib/app-switcher/app-switcher.component.scss
+++ b/projects/cashmere/src/lib/app-switcher/app-switcher.component.scss
@@ -36,8 +36,12 @@
                 padding: 10px;
                 text-decoration: none;
 
-                & > img {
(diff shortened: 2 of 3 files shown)


projects/cashmere/src/lib/app-switcher/app-switcher.component.html
```diff
@@ -1,26 +1,26 @@
...
+</div>
+
+<hc-popover-content #appSwitcherContent>
+    <div class="popover-content">
+        <h4>Applications</h4>
+        <pre *ngIf="!applications">loading applications...</pre>
+        <div class="apps">
+            <div class="app" *ngFor="let app of applications">
+                <a class="thumbnail" title="{{app.Description}}" href="{{app.ServiceUrl}}" target="_blank">
+                    <div class="thumbnail-img" style="background-image: url('{{app.Icon}}')"></div>
```
joeskeen: replace this line with the following:
```suggestion
                    <div class="thumbnail-img" [style.background-image]="'url(' + app.icon + ')'"></div>
```

andrew-frueh: Thanks @joeskeen !

joeskeen:
@andrew-frueh looks like this is ready to merge!

PR #563, opened 2018-10-23
style(table): optional borders

### LINKED ISSUE #558: Table: borders parameter
Based on data collected from our recent user feedback lab, it was determined that users were struggling to understand that they could sort table columns by clicking their header cell.  To improve the usability we adjusted the style to have borders on by default.  But it does make the table significantly heavier, and there will likely still be situations where borders are not needed.

Include a boolean parameter to hc-table called `borders` that defaults to true but may be turned off.

Let users override the styles themselves.

Table

andrew-frueh:
As per the conversation on Slack today, adds a parameter to set whether the table component has solid borders or not.

closes #558

### DIFF
--- a/projects/cashmere/src/lib/sass/tables.scss
+++ b/projects/cashmere/src/lib/sass/tables.scss
@@ -61,6 +61,7 @@ table.hc-table {
 
     tbody {
         font-size: $tbody-font-size;
+        border: $table-border-interior;
 
         tr {
             &:nth-child(2n) {
@@ -89,14 +90,6 @@ table.hc-table {
 
         td {
             vertical-align: top;
-
-            &:first-of-type {
-                border-left: $table-border-interior;
-            }
-
-            &:last-of-type {
-                border-right: $table-border-interior;
-            }
         }
     }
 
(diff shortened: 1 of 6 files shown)


projects/cashmere/src/lib/sass/tables.scss
```diff
@@ -26,10 +26,10 @@ $row-selected-font-color: $white;
 $row-selected-border-color: $white;
 
 table.hc-table {
-    border: $table-border;
     width: 100%;
     max-width: 100%;
     color: $tbody-font-color;
+    border: 1px solid transparent;
```
jacquie: We could make a variable for this border combo:
```suggestion
    border: $table-border-transparent;
```

projects/cashmere/src/lib/sass/tables.scss
```diff
@@ -26,10 +26,10 @@ $row-selected-font-color: $white;
 $row-selected-border-color: $white;
```
jacquie: Add variable for transparent border
``` $table-border-transparent: 1px solid transparent;```

andrew-frueh: Thanks @jacquie - makes sense and made the change.  This one is actually an annoying bit of styling that I had to add.  What I was finding was that when I added the 2px border to `<tbody>` on a table with `width: 100%`, the table was overflowing it's container by the width of the border (so horizontal scroll bars were appearing).  By adding this transparent border to the overall table, it seemed to inset the table to allow the tbody borders to appear without extending past 100%.  @corykon - not sure if there's something I could have done differently to avoid this, but this was the only solution I could come up with that worked.

PR #564, opened 2018-10-25
feat(select): hc-form-field support for select

### LINKED ISSUE #168: Forms.scss
We should create a sass file with general styles for form elements, including things like:
- Borders (width, color, etc)
- Error/focus states
- Font sizes
- padding/spacing
- etc

Two reasons for this:
1. Some styles are shared across many form controls, and as such should be variablized for easier changes in the future
2. It would be nice to have styles available so that someone could have their forms look style guide approved even if they cannot use the official angular components.

### LINKED ISSUE #290: hc-form-field needs to work with hc-select
hc-form-field should work with hc-select so that we can have a uniform handling of errors and labels for form elements.

- Interface will need to be created that hc-form-field can use.
- InputDirective and hc-select will need to implement
- Pull hc-form-field out of input directory and place it along with it's helper directives and components(hcPrefix, hcSuffix, hcError) in its own directory

### LINKED ISSUE #423: Deprecate hc-label
**Describe the bug**
Not sure if this would be a bug or a feature, but let's go with bug :)

It appears there are two separate `LabelComponent`s. 

The first, which requires `LabelModule` targets elements such as `<label hc-label></label>`.
The second, which requires `InputModule` targets elements such as `<hc-label></hc-label>`.

When using both in the same file, importing `LabelModule` causes an error to be thrown for the `<hc-label></hc-label>` element, whereas importing the `InputModule` appears to cover both.

andrew-frueh:
This is a massive PR, so thanks in advance for any time you can give to review it.  It follows the approach we agreed to on our webex last week to break hc-form-field out into wrapper class that can be used to unify all our form elements.  It's a pretty huge upgrade for cashmere because it will standardize our form display including labels, errors, stacked/inline layouts, etc (something on our wishlist since #168).

`hc-form-control` is the new wrapper component added that all the form components can extend.  I've also depreciated the old `<label hc-label>` component as it's redundant to the `<hc-label>` in hc-form-field and has been causing confusion (see #423).  Depending on how we want to define it, that could be considered a breaking change.  It won't break anyone's app who was using it, but it also won't display as expected.  Note that hc-select still does work just fine when it is not included in an hc-form-control wrapper.  So nothing else about this PR introduces anything breaking.

I've also gone back through and updated any examples that were using the old label and updated them to take advantage of hc-form-field.  

closes #290, closes #423, closes #168


projects/cashmere/src/lib/form-field/hc-form-field.component.html
```diff
@@ -0,0 +1,32 @@
+<div [ngClass]="{ 'hc-form-field-wrapper': !inline, 'hc-form-field-wrapper-inline': inline }">
+    <span [ngClass]="{ 'hc-form-field-label-wrapper': !inline, 'hc-form-field-label-wrapper-inline': inline }">
```
corykon: just a coding style preference on my part, but this might be cleaner:
`[ngClass]="inline ? 'hc-form-field-label-wrapper-inline' : 'hc-form-field-label-wrapper'"`

projects/cashmere-examples/src/lib/form-field-overview/form-field-overview-example.html
```diff
@@ -0,0 +1,29 @@
+<form>
```
benjanderson: This is great! Can it work with text areas and checkboxes? It would be great to have it in this example

projects/cashmere-examples/src/lib/form-field-overview/form-field-overview-example.html
```diff
@@ -0,0 +1,29 @@
...
+    <div class="form-sample">
+        <hc-form-field>
+            <hc-label>Item Name:</hc-label>
+            <input hcInput required [formControl]="inputControl">
+            <hc-error>An item name is required</hc-error>
```
benjanderson: This is more of a question than a suggestion, but is there a way to handle different types of errors? 

What if I wanted to validate required OR max length?

projects/cashmere/src/lib/form-field/hc-form-field.component.html
andrew-frueh: ah, yes that's way better - good call

projects/cashmere/src/lib/form-field/hc-form-field.component.html
```diff
@@ -0,0 +1,32 @@
...
+        <label [ngClass]="{ 'hc-form-field-label': !inline, 'hc-form-field-label-inline': inline }"
+                [attr.for]="_control.componentId">
```
benjanderson: This is clever and I love it. Very good for screen readers :)

projects/cashmere/src/lib/select/select.md
```diff
@@ -0,0 +1,14 @@
+The hc-select component may be nested within a hc-form-field component. hc-form-field acts as a coordinator between multiple components including label and error elements.
```
benjanderson: Good explanation!

projects/cashmere-examples/src/lib/form-field-overview/form-field-overview-example.html
andrew-frueh: Yep, in that example I've got select and input.  And once this PR gets finalized, I'm planning on incorporating it with checkboxes (#565) and radio buttons (#566) as you said - and then I'll get them added to the example as well.

projects/cashmere-examples/src/lib/form-field-overview/form-field-overview-example.html
andrew-frueh: Yup, I think we're in good shape there.  Jeremy had everything set up so you can use [Validators](https://angular.io/api/forms/Validators) on [FormControl](https://angular.io/api/forms/FormControl) which should cover just about any type of errors you'd want to check for.  Take a look at the example `input-required` to see that being used.

PR #579, opened 2018-10-26
feat(tabs): support HTML titles

### LINKED ISSUE #259: Tab titles only allow string values
As an enhancement to the tabs component, we may want to consider allowing tab titles to include icons as well as string values. An example of when this scenario might come up is using the tabs as part of a multi-page form and indicating that there is an invalid field on a tab that is not visible with a warning icon.

andrew-frueh:
allows tab titles to include HTML markup instead of only plain text

closes #259

### DIFF
--- a/projects/cashmere-examples/src/lib/tabs-horizontal/tabs-horizontal-example.html
+++ b/projects/cashmere-examples/src/lib/tabs-horizontal/tabs-horizontal-example.html
@@ -1,18 +1,18 @@
 <div class="tab-demo">
     <hc-tab-set direction="horizontal">
-        <hc-tab tabTitle="Tab 1">
+        <hc-tab tabTitle='Pending Tasks'>
             <div class="tab-content">
-                Tab content 1
+                This is where pending tasks would go.
             </div>
         </hc-tab>
-        <hc-tab tabTitle="Tab 2">
+        <hc-tab tabTitle='<i class="fa fa-check" style="color: #00a859; font-size: 20px;"></i>  Complete Tasks'>
             <div class="tab-content">
-                Tab content 2
+                These are completed tasks
             </div>
         </hc-tab>
-        <hc-tab tabTitle="Tab 3">
+        <hc-tab tabTitle='<i class="fa fa-times fa-2x" style="color: #f13c45; font-size: 20px;"></i> Incomplete Tasks'>
             <div class="tab-content">
-                Tab content 3
+                Incomplete tasks belong here.
             </div>
         </hc-tab>
     </hc-tab-set>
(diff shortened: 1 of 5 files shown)


projects/cashmere/src/lib/tabs/tab.component.ts
```diff
@@ -8,11 +9,17 @@ import {Component, Input} from '@angular/core';
...
     /** Router path that the tab routes to. If one tab uses the routerLink in a tab set, all must use the router link.
      * Can be specified as '/path/2' or ['path', '2']
      */
     @Input() routerLink: any[] | string;
     _active: boolean = false;
+
+    constructor(private sanitizer: DomSanitizer) {}
+
+    _safeHtmlTitle(): SafeHtml {
+        return this.sanitizer.bypassSecurityTrustHtml(this.tabTitle);
```
andrew-frueh: This was necessary to keep the standard HTML sanitizer from stripping out a lot of standard markup like inline styles, etc.  Would like to verify with those of you who know more about security than me that this isn't introducing a significant security risk.

projects/cashmere-examples/src/lib/tabs-horizontal/tabs-horizontal-example.html
```diff
@@ -1,18 +1,18 @@
...
     <hc-tab-set direction="horizontal">
-        <hc-tab tabTitle="Tab 1">
+        <hc-tab tabTitle='Pending Tasks'>
             <div class="tab-content">
-                Tab content 1
+                This is where pending tasks would go.
             </div>
         </hc-tab>
-        <hc-tab tabTitle="Tab 2">
+        <hc-tab tabTitle='<i class="fa fa-check" style="color: #00a859; font-size: 20px;"></i>  Complete Tasks'>
```
joeskeen: It seems strange to me that HTML would be inside an element attribute.  Usually when this kind of thing is needed you just have an optional subelement, i.e.

```suggestion
<hc-tab tabTitle="Pending Tasks">
   <hc-tab-content>
      This is where pending tasks would go.
   </hc-tab-content>
</hc-tab>
<hc-tab>
   <hc-tab-title>
      <i class="fa fa-check" class="complete-tasks"></i> Complete Tasks
   </hc-tab-title>
   <hc-tab-content>
      These are completed tasks
   </hc-tab-content>
</hc-tab>
```

projects/cashmere/src/lib/tabs/tab.component.ts
joeskeen: There is some risk in doing this such as:

```
<hc-tab [tabTitle]="someUserProvidedValue">
```

In this case, the user could provide whatever HTML they want to and it would be injected here without being sanitized.  This could pose a security risk.

PR #580, opened 2018-10-28
Tab Event Handling

### LINKED ISSUE #255: tab-set doesn't emit currently selected tab
It would be immensely helpful to have an observable that people can subscribe to on the tab-set that notifies them when and what new tab has been selected.

```
<hc-tab-set direction="horizontal" (selectChange)="selectionChanged($event)">
    <hc-tab tabTitle="Tab 1">
        Tab content 1
    </hc-tab>
    <hc-tab tabTitle="Tab 2">
        Tab content 2
    </hc-tab>
</hc-tab-set>
```

We should allow the user to listen for a selectChange event and that event should be the following object

```
{
  index: number,
  tab: TabComponent
}
```

### LINKED ISSUE #548: Cannot bind to (click)="doSomething()" on `hc-tab`
**Describe the bug**
<hc-tab tabTitle="Please Do Something" (click)="doSomething()"> does not work.

**To Reproduce**
Reproduce it on stackblitz fork:
    - https://stackblitz.com/edit/cashmere-demo-vjveyq?file=app%2Fhome%2Fhome.component.html
    - https://cashmere-demo-vjveyq.stackblitz.io

**Expected behavior**
I would expect (click) would allow the user of `hc-tab` to bind a click event to clicking the tab.

**Additional context**
As a user of `hc-tab`, I'd like clicking on the tab to be configurable, so that I may assign behavior as needed.

### LINKED ISSUE #556: Tabs: Allow "add tab" button use case

We are in the process of converting IDEA from AngularJS / Bootstrap to Angular / Cashmere.  One use case we have that I have discovered Cashmere can't really accommodate (well) at the moment is the ability to dynamically add tabs.  Consider this AngularJS / Bootstrap page:
image
Notice that the tab set allows the user to add tabs by clicking the '+' icon at the right side of the tab set.  There are a few things that make doing this with Cashmere tabs today impossible:

* #548 Can't bind click events to tab headers
* #259 Can't use the 'plus' icon as a tab title (using the `+` character looks super lame)
* Can't programmatically select/activate a tab (or it isn't documented at all)

I can think of a handful of viable solutions, but this is the one I think would work the best.  Feel free to propose a better way:

Add an option to show an add button (by default `false`).  Then we could surface the `newTab` event to bind to.  The tricky part is we would still have to be able to control which tab is selected programmatically so that the newly created tab can be selected.

If the issues referenced above were fixed, we could make the tab header be a plus icon, then create the tab without content and bind it to a click event which modifies the tab list backed by the binding.  This is how I tried to implement it:

```html
<div class="content">
    <hc-tab-set direction="horizontal">
        <hc-tab tabTitle="Details">
            <div class="tab-content">
                Details
            </div>
        </hc-tab>
        <hc-tab tabTitle="Main">
            <div class="tab-content">
                Main
            </div>
        </hc-tab>
        <ng-container *ngIf="application">
            <hc-tab *ngFor="let subset of application.subsets" [tabTitle]="subset.name">
                <div class="tab-content">
                    {{subset.name}}
                </div>
            </hc-tab>
        </ng-container>
        <hc-tab tabTitle="+" (click)="newSubset()">
        </hc-tab>
    </hc-tab-set>
</div>
```

Overall, it feels a bit clunky (having a tab that's not a tab, but rather a button).

*   Components changed? : Tabs
*   Are these breaking changes to public interfaces? not necessarily

andrew-frueh:
- Adds a `selectChange` emitter to `TabSetComponent` that fires when the select tab in the set is changed.  
- Add a `selectTab` function to `hc-tab-set` that allows the selected tab to be set programmatically via either a numerical index or a `TabComponent` object
- Adds a `tabClick` emitter on `TabComponent` that fires when the tab is clicked.  `(click)` doesn't work because the tabs are generated in the tab-set template separately from the content.  So the click is being attached to something that doesn't exist.  But this emitter gets around that and will allow apps to bind events to tab clicks.
- Updated the Horizontal Tab example to illustrate all of the above.

closes #255, closes #556, closes #548

### DIFF
--- a/projects/cashmere/src/lib/tabs/tab-set.component.ts
+++ b/projects/cashmere/src/lib/tabs/tab-set.component.ts
@@ -64,12 +64,29 @@ export class TabSetComponent implements AfterContentInit {
         });
     }
 
+    /** Sets the currently selected tab by its numerical index  */
+    selectTabByIndex(index: number) {
+        let i: number = 0;
+
+        this._tabs.forEach(t => {
+            if (i === index) {
+                this._setActive(t);
+            }
+            i++;
+        });
+    }
+
+    /** Sets the currently selected tab by its `TabComponent` object  */
+    selectTabByObject(tab: TabComponent) {
+        this._setActive(tab);
+    }
+
     _setActive(tab: TabComponent) {
         let selectedTab: number = 0;
         let index: number = 0;
 
         this._tabs.forEach(t => {
-            if( t === tab) {
+            if (t === tab) {
                 selectedTab = index;
             }
             t._active = false;
--- a/projects/cashmere/src/lib/tabs/tab-set.component.ts
+++ b/projects/cashmere/src/lib/tabs/tab-set.component.ts
@@ -64,24 +64,23 @@ export class TabSetComponent implements AfterContentInit {
         });
     }
 
-    /** Sets the currently selected tab by its numerical index  */
-    selectTabByIndex(index: number) {
-        let i: number = 0;
+    /** Sets the currently selected tab by either its numerical index or `TabComponent` object  */
+    selectTab(tab: number | TabComponent) {
+        if (typeof tab === 'number') {
+            let i: number = 0;
 
-        this._tabs.forEach(t => {
-            if (i === index) {
-                this._setActive(t);
-            }
-            i++;
-        });
+            this._tabs.forEach(t => {
+                if (i === tab) {
+                    this._setActive(t);
+                }
+                i++;
+            });
+        } else {
+            this._setActive(tab);
+        }
     }
 
-    /** Sets the currently selected tab by its `TabComponent` object  */
-    selectTabByObject(tab: TabComponent) {
-        this._setActive(tab);
-    }
-
-    _setActive(tab: TabComponent) {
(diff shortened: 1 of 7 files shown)


projects/cashmere/src/lib/tabs/tab-set.component.ts
```diff
@@ -30,6 +34,9 @@ export class TabSetComponent implements AfterContentInit {
 
     @ContentChildren(TabComponent) _tabs: QueryList<TabComponent>;
 
+    /** Emits when the selected tab is changed */
+    @Output() selectChange: EventEmitter<TabChangeEvent> = new EventEmitter();
```
benjanderson: I am not a fan of "selectChange" as the event. What about "selectedTabChange" or "tabSelected"?
Angular Material - "selectedTabChange"
https://material.angular.io/components/tabs/api

Kendo UI - "tabSelect"
https://www.telerik.com/kendo-angular-ui/components/layout/tabstrip/

andrew-frueh: works for me - I like `selectedTabChange` mostly because it feels more clearly differentiated from the `tabClick` event on TabComponent

PR #591, opened 2018-11-02
fix(navbar): remove dependency on app switcher service and cleanup

### LINKED ISSUE #340: `hc-navbar-mobile-menu` has a dependency on the App Switcher
**Describe the bug**
`hc-navbar-mobile-menu` appears to have a dependency on the App Switcher. Adding in `hc-navbar` works fine without the app switcher, but adding in `hc-navbar-mobile-menu`, leaving the `appSwitcher` property off or setting to `false` causes this error in the console:

```
ERROR
Error: StaticInjectorError(AppModule)[IAppSwitcherService]: 
StaticInjectorError(Platform: core)[IAppSwitcherService]: 
NullInjectorError: No provider for IAppSwitcherService!
```

**To Reproduce**
https://stackblitz.com/edit/cashmere-demo-a69f58

On this particular page ([navbar.component.html](https://stackblitz.com/edit/cashmere-demo-a69f58?file=app%2Fnavbar%2Fnavbar.component.html,`)) if you uncomment the `hc-navbar-mobile-menu` it will no longer load the page correctly

**Expected behavior**
I wouldn't expect to need the app switcher provided when adding in a mobile menu.

benjanderson:
- removed dependency on app switcher service and closed #340 
- added app switcher and mobile menu examples
- cleaned up dependency injection so services are properly injected with injection token
- added documentation on how to work with app switcher including customizing app switcher service


projects/cashmere/src/lib/app-switcher/app-switcher.component.html
```diff
@@ -10,7 +10,7 @@ <h4>Applications</h4>
         <div class="apps">
             <div class="app" *ngFor="let app of applications">
                 <a class="thumbnail" title="{{app.Description}}" href="{{app.ServiceUrl}}" target="_blank">
-                    <div class="thumbnail-img" [style.background-image]="'url(' + app.icon + ')'"></div>
```
joeskeen: I seem to remember the `style.background-image` being made that way as a fix for #503 .  Can you verify that this still works with the change?

projects/cashmere/src/lib/app-switcher/app-switcher.component.scss
```diff
@@ -39,9 +39,6 @@
                 & > .thumbnail-img {
                     width: 100px;
                     height: 100px;
-                    background-size: contain;
```
joeskeen: Can you confirm that this is not a regression of #503 ?

projects/cashmere/src/lib/app-switcher/app-switcher.service.ts
```diff
@@ -1,31 +1,23 @@
...
-    constructor() {}
-
-    public getApplications(): Observable<IDiscoveryRequest> {
-        return of({value: getMockApplication(5)}).pipe(delay(500));
-    }
-}
+import {IAppSwitcherConfig, IAppSwitcherService, IDiscoveryRequest, APP_SWITCHER_CONFIG} from './app-switcher-interfaces';
 
 @Injectable()
 export class AppSwitcherService implements IAppSwitcherService {
```
joeskeen: Don't have to change this, but I was thinking, would it be better to call this `DiscoveryAppSwitcherService`?  It would be more clear the coupling between the DiscoveryService and this service this way, and not cause naming conflicts if there is a custom app switcher service not dependent on DiscoveryService.  Just my 2¢.

projects/cashmere/src/lib/app-switcher/app-switcher-application-link/app-switcher-links.component.html
```diff
@@ -0,0 +1,5 @@
+<div class="bottomSpacer"></div>
+<a class="app-switcher" href={{_appSwitcherService?.allApplicationsUri}}>
```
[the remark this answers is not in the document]
benjanderson: Good suggestion!

I abstracted the ability to show or hide this link by including/excluding the `<hc-app-switcher-links></hc-app-switcher-links>` element. See `navbar-app-switcher-example.html`. This was a part of the work to not require app-switcher service when using the navbar #340

projects/cashmere/src/lib/app-switcher/app-switcher-application-link/app-switcher-links.component.html
```diff
@@ -0,0 +1,5 @@
...
+    <i class="fa fa-th"></i>
```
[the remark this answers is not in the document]
benjanderson: good suggestion

projects/cashmere/src/lib/app-switcher/app-switcher.component.html
```diff
@@ -10,7 +10,7 @@ <h4>Applications</h4>
...
+                    <img class="thumbnail-img" src={{app.Icon}}>
```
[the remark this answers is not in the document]
benjanderson: Good catch! These curly braces were originally in the code, but it for sure cleans things up to use this way

projects/cashmere/src/lib/app-switcher/app-switcher.component.scss
benjanderson: see above

projects/cashmere/src/lib/app-switcher/app-switcher.service.ts
benjanderson: I like this, I think I want to focus on more important issues, so feel free to open an issue

projects/cashmere/src/lib/app-switcher/app-switcher.component.html
benjanderson: The discovery service returns a base64 image. I found this most recent change broke the discovery service integration. I have tested my fix and it seems to work

projects/cashmere/src/lib/app-switcher/app-switcher-application-link/app-switcher-links.component.ts
```diff
@@ -0,0 +1,11 @@
+import {Component, Inject} from '@angular/core';
+import {IAppSwitcherService, APP_SWITCHER_SERVICE} from '../app-switcher-interfaces';
+
+@Component({
+    selector: 'hc-app-switcher-links',
+    templateUrl: './app-switcher-links.component.html',
+    styleUrls: ['./app-switcher-links.component.scss']
+})
+export class AppSwitcherLinksComponent {
+    constructor(@Inject(APP_SWITCHER_SERVICE) public _appSwitcherService: IAppSwitcherService) {}
```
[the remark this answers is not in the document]
benjanderson: This happens quite a few places in cashmere, I guess its confusing here because the default accessor is public, so this is superfluous. It is public to allow the template access, but has the underscore to not show up in the documentation.

projects/cashmere/src/lib/app-switcher/app-switcher.component.html
andrew-frueh: hey @benjanderson - the main reason for using `background-img` on a div instead of an img was that it allows us to scale any image proportionally that's added to the app switcher.  With the switch as you've changed it, an image that isn't 1:1 will be distorted.  Do base64 images not work as a background-image?

PR #597, opened 2018-11-06
refactor(navbar): separates app switcher

### LINKED ISSUE #593: Help users know that the flame icon is an app switcher
It isn't clear to users that the flame icon is an app switcher.  We have to verbally tell them, which seems to indicate that there is a problem.

I like how when my bank, credit card company, or google does a short walkthrough of their new UI the first/second time I login.  They highlight where things have moved by fading the background and pointing out where things are.  I can click "next" once or twice and then click "got it" when I'm done.  They don't over-do it... they just highly 1-3 things and then let me dismiss the little tutorial.  (Sometimes they have a link that says, "Show this again next time" next to the "dismiss" link).

Get rid of the flame icon and make it a waffle.  I understand this may not be an option.

andrew-frueh:
Based on consistent feedback from a variety of sources, we had a UX problem with the location of our app switcher.  It was previously activated by clicking on the tri-flame box in the top left, but there was no visual indication to the user to click there. As a result, the app switcher was getting lost.

After reviewing other industry practices regarding app switching, it was decided that including it as an icon next to the help menu would be a significant improvement.  It groups all global app functionality together on the right side of the navbar, and drastically improves the discoverability of the button.

This *BREAKING CHANGE* removes the app switcher from the navbar's template and allows the user to attach it to a `hc-icon` (or anywhere else for that matter).  Component examples and usage markdown has also been updated to reflect the change.

closes #593

### DIFF
--- a/projects/cashmere/src/lib/navbar/navbar.md
+++ b/projects/cashmere/src/lib/navbar/navbar.md
@@ -15,11 +15,11 @@ All Health Catalyst apps should include a help menu in their navbar. This guide 
 
 You may not have all these items available. However, include what you have in this order:
 
--   **Help Topics** (if available—reach out to the Content Team for support)
--   **Release Notes** (if applicable, again reach out to the Content Team for support)
--   **About** (a modal containing app reference information - see the [associated style page](https://cashmere.healthcatalyst.net/styles/about) for guidelines)
--   **Health Catalyst Community** (link to community space specific to the app)
--   **Send us your feedback** (see the [User Feedback Guide](https://cashmere.healthcatalyst.net/components/typeform-survey/usage))
+*   **Help Topics** (if available—reach out to the Content Team for support)
+*   **Release Notes** (if applicable, again reach out to the Content Team for support)
+*   **About** (a modal containing app reference information - see the [associated style page](https://cashmere.healthcatalyst.net/styles/about) for guidelines)
+*   **Health Catalyst Community** (link to community space specific to the app)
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -28,7 +28,6 @@
         "postbuild": "npm run copy-assets",
         "precopy-assets": "npm run docs",
         "copy-assets": "node scripts/guide-copy-assets.js",
-        "precommit": "lint-staged",
```
benjanderson: I am good with this, we already have linting as a part of the pr validation. If we are removing this, we should remove the "lint-staged" section of the package.json

projects/cashmere/src/lib/navbar/navbar.md
```diff
@@ -43,18 +43,25 @@ You may not have all these items available. However, include what you have in th
...
-    <hc-app-switcher brandIcon="./assets/TriFlame.svg"></hc-app-switcher>
+    ...
+    <hc-icon class="hc-navbar-icon" fontSet="fa" fontIcon="fa-th" [hcPopover]="appSwitcher" popperPlacement="bottom"></hc-icon>
     ...
     <hc-navbar-mobile-menu>
         ...
         <hc-app-switcher-links></hc-app-switcher-links>
     </hc-navbar-mobile-menu>
+    ...
+    <hc-popover-content #appSwitcher>
```
benjanderson: I like giving them the option and ability to customize it being in a popover 👍

package.json
andrew-frueh: Cool, good catch - just pushed a commit that removes that lint-staged section.

PR #600, opened 2018-11-07
599 load more paginator

### LINKED ISSUE #599: "Load more" pagination
This is a common pagination approach in pop builder:
image

A new component that is just a simple "load more" button with the built in functionality for pulling another page of data from table-data-source.ts

None

I'll create a new abstract pagination class which the current `hc-pagination` and the new `hc-load-more-pagination` can extend.

In the screenshot above, you'll notice we use a smaller button. I'll just implement using the standard button size. Maybe somewhere down the road we had button size to our button component... sm, reg, lg... not sure it's that important though.

corykon:
closes #599 

I think it'd be cool to do an auto-load on scroll type paginator, too


corykon:
Can I just mention... the prettier plugin kills me. I think I've a good understanding of its benefits, and I'm all for it, but for the life of me, I can't seem to get consistent results locally and on the travis build server. It's crapshoot as to what'll happen with the build every time I submit a PR...

PR #605, opened 2018-11-11
Toaster messages

andrew-frueh:
Toaster system written from the ground up for Cashmere and fully leveraging the Angular CDK for displaying the overlays.  Includes four standard styles, as well as the ability to create an entirely custom toast by passing in a template.  Allows the user to place toasts in any corner of the window, or the center top or bottom, or full-width top or bottom.  Multiple toasts in a given location will stack, and will reorder when one is closed.  Toasts can auto-dismiss on a set time interval and/or be clicked to dismiss.

Note that it was necessary for me to upgrade our CDK from 6.4.2 to the newest version, 7.0.3.  The upgrade was needed because the OverlayRef object in our version was missing an important function that allows you to update the PositionStrategy.  When a toast is closed, I need to be able to re-parent the remaining toasts based on the new order in the stack, and this function allows me to do that.


projects/cashmere/src/lib/toaster/hc-toast-options.ts
```diff
@@ -0,0 +1,18 @@
...
+    type?: string;
+    /** Position and sizing of the toaster message. Defaults to bottom-right.
+     * Options are: `top-right`, `top-center`, `top-left`, `top-full-width`,
+     * `bottom-right`, `bottom-center`, `bottom-left`, `bottom-full-width`.*/
+    position?: string;
+    /** Timeout value in milliseconds sets the amount of time the toast is displayed.
+     *  Defaults to 5000. A value of 0 means the toast will not auto-dismiss.*/
+    timeout?: number;
+    /** When set to true, the user may click the Toast to dismiss it.  Defaults to false.*/
+    clickDismiss?: boolean;
```
corykon: Just a thought... perhaps add a close button if clickDismiss is true?

projects/cashmere/src/lib/toaster/hc-toast-options.ts
```diff
@@ -0,0 +1,18 @@
...
+     * Options are: `success`, `info`, `warning`, `alert`, `custom`*/
+    type?: string;
+    /** Position and sizing of the toaster message. Defaults to bottom-right.
+     * Options are: `top-right`, `top-center`, `top-left`, `top-full-width`,
+     * `bottom-right`, `bottom-center`, `bottom-left`, `bottom-full-width`.*/
+    position?: string;
+    /** Timeout value in milliseconds sets the amount of time the toast is displayed.
+     *  Defaults to 5000. A value of 0 means the toast will not auto-dismiss.*/
+    timeout?: number;
+    /** When set to true, the user may click the Toast to dismiss it.  Defaults to false.*/
```
corykon: another thought... maybe we create an issue for this... a click callback might be a cool option. This could be probably be done today with custom toasts, I'm assuming

projects/cashmere/src/lib/toaster/hc-toast.component.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    constructor(public _el: ElementRef) {}
+
+    get _styleType(): string {
+        return this._styleTypeVal;
+    }
+    set _styleType(typeVal: string) {
+        this._styleTypeVal = typeVal;
+        if (typeVal === 'success') {
+            this._toastIcon = 'fa-check-circle';
```
corykon: I think it would valuable to just base64 encode these icons rather that relying on font awesome to be available

andrew-frueh:
Great suggestions @corykon - just pushed an update that should address your three suggestions.  Let me know what you think.

Issue #200, opened 2018-04-12
Many components do not render when navigating between tabs in Microsoft Edge


benjanderson:
Reproduced on Edge 41.16299.15.0
![image](https://user-images.githubusercontent.com/1934453/38686199-08446798-3e31-11e8-89de-5031a1f825ab.png)

The following error was in the console, I am not sure if its related
`HTML1500: Tag cannot be self-closing. Use an explicit closing tag.`

benjanderson:
I think this is a tabs issue, as many components are not fully rendered unless you do a full refresh of the page. I cannot get this to reproduce outside of tabs in microsoft edge. I am renaming this issue

jtrujill:
@benjanderson do you know if this is still an issue

benjanderson:
Yes, about half of them do not work with the following error

![image](https://user-images.githubusercontent.com/1934453/44874452-71bb4d00-ac58-11e8-8459-a1c2b6976e4e.png)

PR #610, opened 2018-11-13
fix(all): demo fix in ie11

### LINKED ISSUE #491: Demo fails to load in Internet Explorer 11
**Describe the bug**
IE11 fails to load https://cashmere.healthcatalyst.net/

**To Reproduce**
1.  Open IE11 
2.  Goto https://cashmere.healthcatalyst.net/

**Browser:**
Internet Explorer 11

**Additional context**
When trying to access https://cashmere.healthcatalyst.net/ through IE11 it fails to load. Opening the dev console reveals the error "Syntax error vendor.js (134317, 27)". Navigating to the code shows the following line
```
const ucs2encode = array => String.fromCodePoint(...array);
```
This piece of code is tied to punycode.js https://github.com/bestiejs/punycode.js/ and it appears this is a current issue with angular

https://github.com/angular/angular-cli/issues/9508#issuecomment-381104612

benjanderson:
Fixes issues with loading the demo in IE11 by adding a hard dependency on an older version of punycode. According to this comment, they updated their code to only support ES6 https://github.com/angular/angular-cli/issues/9508#issuecomment-391490414 and I tested out just adding the older version to our node_modules directory and it worked ;)

Would someone with a windows machine validate this works as well?

Closes #491


corykon:
@benjanderson  here's what I'm seeing:
**In IE11:** Everything renders, but there are several small bugs. I can submit a new issue for all of these.

**In Edge:** Accordion & Form Field don't render at all. Navbar and Picklist sometimes show up, but on a second visit to that tab seemed not to render. Not seeing anything in the console. :/

corykon:
Logged IE 11 bugs here: #614

benjanderson:
@corykon Wow, thanks for the thorough feedback. My goal was just to get IE to load and it looks like you successfully did that. I think we are on a good course to making the library more robust with IE and edge by logging #614 and #200 .

That said, I think we can merge these changes as a first step. Could you please approve it?

PR #611, opened 2018-11-14
feat(button): minimal button style

### LINKED ISSUE #585: New Button Style
image

A new minimal button style to be used for non-primary functions.

Dark grey font color for text and icon. Can be just text, or just icon, or both. Has hover effect as shown in screenshot. The height of these button is much smaller than our standard button size (only 20px)

new `buttonStyle` in `hc-button`

**Are these breaking changes to public interfaces?**
not breaking!

andrew-frueh:
adds new 'minimal' button style; also removes the "disabled" button example - it was redundant to others and wasn't adding any additional value.

closes #585


corykon:
Looks perfect. My only ask would be if we could also introduce button size. The minimal buttons I use are usually tucked up in the right corner, and fit better when they're not quite so tall.
image

@andrew-frueh thoughts on that? I could introduce a new buttonSize param.

andrew-frueh:
I think introducing buttonSize as a separate param is a good way to go on that.  Because we'd probably want that parameter to work with all button styles, not just minimal.  There might be a reason why someone would need a small primary button.  So maybe a separate PR for that?

Issue #612, opened 2018-11-14
ButtonSize param


corykon:
**Describe the problem this feature request is solving**
Allow for smaller buttons when needed

**Describe the solution you'd like**
New parameter, `buttonSize`

**If any, which components will have to change, and how**
`hc-button` and `hc-split-button`

**Additional context**
Probably will use something like this:
```
font-size: 14px;
height: 26px;
padding: 3px 10px;
line-height: 14px;
min-width: auto;
```

corykon:
discussed in #611

andrew-frueh:
Also a reminder to update one of the existing button examples to demonstrate this functionality when the parameter is added.

PR #616, opened 2018-11-15
feat(button): add button sizes

corykon:
add buttons sizes and clean up some demo examples

re #612 

one note--- You might notice some wacky positioning of split button dropdowns in the demo code. There's a cdk bug causing this, but it has been fixed in the version of the cdk andrew is updating us to in #605.


projects/cashmere/src/lib/sass/button.scss
```diff
@@ -67,6 +62,30 @@ $btn-icon-sz: 18px;
     }
 }
 
+@mixin hc-button-sm() {
+    padding: 3px 12px;
+    height: 26px;
+    min-width: auto;
+    font-size: 14px;
```
andrew-frueh: Should we be using the fontSize() mixin on these instead of using `font-size`?

projects/cashmere/src/lib/sass/button.scss
```diff
@@ -67,6 +62,30 @@ $btn-icon-sz: 18px;
...
+    padding: 0 20px;
+    height: 35px;
+    min-width: 150px;
+    font-size: 15px;
+    line-height: 15px;
+}
+
+@mixin hc-button-lg() {
+    padding: 0 30px;
+    height: 45px;
```
andrew-frueh: For `lg` I'd suggest we do 40px rather than 45.  45 is pretty huge and I'm sure sure I'd ever see a use for that.  But in prototypes before I have used 40px buttons, like on the close button at the bottom of a modal.

projects/cashmere/src/lib/sass/button.scss
```diff
@@ -67,6 +63,30 @@ $btn-icon-sz: 18px;
...
+    min-width: 150px;
+    @include fontSize(15px);
+    line-height: 15px;
+}
+
+@mixin hc-button-lg() {
+    padding: 0 30px;
+    height: 40px;
+    min-width: 200px;
+    @include fontSize(18px);
```
andrew-frueh: Sorry for being nitpicky - I just noticed this now with the 40px size, 18 font size looks a bit big to me.  Can we change this to 17px instead?  Then I think we're good.  Thanks!

PR #631, opened 2018-11-15
fix(pagination): fix pagination event for invalid input

### LINKED ISSUE #630: Pagination: two way data binding for pageNumber and pageSize broken
**Describe the bug**

It is easy enough to subscribe to the event for when page number or size changes, using `(page)=""`, but if you don't always want to start with page 1 and page size 20 every time you refresh the page, it gets ugly.

See, you can bind those values into the pagination component using `[pageSize]` and `[pageNumber]`, but if you attach those bindings to the current page size and page number, it quickly becomes an issue.  As soon as the user changes the page, you get the signal from `(page)` telling you to change the page and what the new page number and page size are.  But the `[pageNumber]` is one-indexed (as it would make little sense for page 1 to be 0, page 2 to be 1, etc...), but the number coming out of the event is sometimes zero-based, so the behavior is unpredictable and leads to `ExpressionChangedAfterItHasBeenCheckedError`s.

**To Reproduce**
Reproduce it on stackblitz fork
https://stackblitz.com/edit/cashmere-demo-seiqvh?file=app%2Fhome%2Fhome.component.ts

**Expected behavior**
You should be able to seed the paginator with values for pageSize and pageNumber on load (using data binding, not ViewChild and setting it manually).

**Screenshots**
image

**Browser:**

-   All

**Additional context**

This worked previously before #594, and is a regression.  Previously the `[(pageNumber)]` was a two-way data binding, and it worked great.  I think we may need to revisit using `[(pageNumber)]` and `[(pageSize)]` instead of having a separate event for `(page)`.

CC @corykon

joeskeen:
provide output properties to bind to both pageNumberChange and pageSizeChange in addition to the
page event; schedule a macrotask for correcting bad input for pageNumber of the pagination component

fix #630

### DIFF
--- a/projects/cashmere-examples/src/lib/pagination-simple/pagination-simple-example.html
+++ b/projects/cashmere-examples/src/lib/pagination-simple/pagination-simple-example.html
@@ -1,0 +1,5 @@
+<hc-form-field> <hc-label>Total number of items</hc-label> <input hcInput type="number" [(ngModel)]="totalItems" /> </hc-form-field>
+<hc-form-field> <hc-label>Page number</hc-label> <input hcInput type="number" [(ngModel)]="pageNumber" /> </hc-form-field>
+<hc-form-field> <hc-label>Page size</hc-label> <input hcInput type="number" [(ngModel)]="pageSize" /> </hc-form-field>
+
+<hc-pagination [(pageNumber)]="pageNumber" [(pageSize)]="pageSize" [length]="totalItems"></hc-pagination>
--- a/projects/cashmere/src/lib/pagination/base-pagination.ts
+++ b/projects/cashmere/src/lib/pagination/base-pagination.ts
@@ -2,6 +2,8 @@ import {EventEmitter, Input, Output, OnInit, Component} from '@angular/core';
 import {coerceNumberProperty} from '@angular/cdk/coercion';
 import {Initailizable} from '../shared/initializable';
 import {PageEvent} from './page-event';
+import {map, distinctUntilChanged} from 'rxjs/operators';
+import {Observable} from 'rxjs';
 
 /**
  * Base Pagination class for shared functionality
@@ -24,6 +26,7 @@ export class BasePaginationComponent extends Initailizable implements OnInit {
     }
     private _length: number = 0;
 
+    private _pageNumber: number = 1;
     /** The currently displayed page. *Defaults to 1.* */
     @Input()
     get pageNumber(): number {
@@ -31,10 +34,15 @@ export class BasePaginationComponent extends Initailizable implements OnInit {
     }
     set pageNumber(value: number) {
         const prevPageNumber = this._pageNumber;
-        this._pageNumber = this._sanitizePageNumber(value);
-        this._emitPageEvent(prevPageNumber);
+        this._pageNumber = value;
+
+        const sanitizedValue = this._sanitizePageNumber(value);
+        if (sanitizedValue !== value) {
+            setTimeout(() => (this.pageNumber = sanitizedValue));
+        } else {
+            this._emitPageEvent(prevPageNumber);
+        }
     }
-    private _pageNumber: number = 1;
 
     /** Number of items to display on a page. *Defaults to 20.* */
     @Input()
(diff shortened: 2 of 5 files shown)


projects/cashmere-examples/src/lib/pagination-simple/pagination-simple-example.html
```diff
@@ -0,0 +1,5 @@
+<hc-form-field> <hc-label>Total number of items</hc-label> <input hcInput type="number" [(ngModel)]="totalItems" /> </hc-form-field>
```
joeskeen: Even with the table examples, I think there is still value in having a non-table example for the pagination component.  I have found myself actually using it without tables more than with tables, and it is nice to see how to set that up.

projects/cashmere/src/lib/pagination/base-pagination.ts
```diff
@@ -24,17 +26,23 @@ export class BasePaginationComponent extends Initailizable implements OnInit {
...
     }
     set pageNumber(value: number) {
         const prevPageNumber = this._pageNumber;
-        this._pageNumber = this._sanitizePageNumber(value);
-        this._emitPageEvent(prevPageNumber);
+        this._pageNumber = value;
+
+        const sanitizedValue = this._sanitizePageNumber(value);
+        if (sanitizedValue !== value) {
+            setTimeout(() => (this.pageNumber = sanitizedValue));
```
joeskeen: This kills the `ExpressionChangedAfterItHasBeenCheckedError`, by scheduling the value correction for after the current change detection cycle.  See https://blog.angularindepth.com/everything-you-need-to-know-about-the-expressionchangedafterithasbeencheckederror-error-e3fd9ce7dbb4#8c66

projects/cashmere/src/lib/pagination/base-pagination.ts
```diff
@@ -48,8 +56,10 @@ export class BasePaginationComponent extends Initailizable implements OnInit {
     private _pageSize: number = BasePaginationComponent._DEFAULT_PAGE_SIZE;
 
     /** Event emitted when the paginator changes the page size or page index. */
-    @Output()
-    readonly page: EventEmitter<PageEvent> = new EventEmitter<PageEvent>();
+    @Output() readonly page: EventEmitter<PageEvent> = new EventEmitter<PageEvent>();
+
+    @Output() readonly pageNumberChange: Observable<number> = this.page.pipe(map(e => e.pageNumber));
```
joeskeen: These allow two-way data binding to those individual properties, if you are not interested in getting the entire `(page)` event.  See `pagination-simple-example` above.

projects/cashmere/src/lib/pagination/base-pagination.ts
```diff
@@ -81,9 +91,9 @@ export class BasePaginationComponent extends Initailizable implements OnInit {
     _changePageSize(pageSize: number) {
         // Current page needs to be updated to reflect the new page size. Navigate to the page
         // containing the previous page's first item.
-        const startIndex = (this.pageNumber - 1) * this.pageSize + 1;
+        const startIndex = (this.pageNumber - 1) * this.pageSize;
```
joeskeen: I was getting some off-by-one errors, fixed by these changes.

joeskeen:
@corykon @andrew-frueh or @benjanderson I would appreciate help with the following lint error I'm getting:

```
> @healthcatalyst/cashmere-docs@2.1.0 lint:prettier /home/travis/build/HealthCatalyst/Fabric.Cashmere
> prettier --list-different "**/*.{ts,js,scss,css,md,json}"
projects/cashmere/src/lib/pagination/base-pagination.ts
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @healthcatalyst/cashmere-docs@2.1.0 lint:prettier: `prettier --list-different "**/*.{ts,js,scss,css,md,json}"`
npm ERR! Exit status 1
```

I can't reproduce it on my machine, as running Prettier on that file doesn't cause any pending changes in my repository.

andrew-frueh:
hey @joeskeen - you should be all set now.  Wish I knew why it wasn't working for you.  I ran `npm run lint:prettier` and it gave me the same error Travis was throwing.  So I ran Format Document on the offending file and it made the correction that I pushed - which got the build to pass.  ::shrug::

joeskeen:
@andrew-frueh I wonder if there is a difference of version of Prettier in play here: I do know on one of their releases they changed the default behavior for inline annotations.  Perhaps we should make that explicit in our Prettier config so it's not to the mercy of what version of Prettier is being run...

PR #634, opened 2018-11-16
feat(icons): add icon styles page

corykon:
re #58 

Adds a page with brief recommendations for implementing icons, and a small set of medical icons. Most of these icons were made custom by Russ Tabet or myself. Some are based heavily on font awesome icons, which are also open source.


benjanderson:
Great job! It will be good having these in the repo 👍

PR #638, opened 2018-11-27
Unit tests pagination 617

jacquie:
Unit tests for the LoadMorePaginationComponent.


projects/cashmere/src/lib/shared/initializable.spec.ts
```diff
@@ -4,7 +4,7 @@ import {Observable, Subscriber} from 'rxjs';
 class MockInitializable extends Initializable {}
 
 let testInitializable: Initializable;
-fdescribe('Initializable', () => {
+describe('Initializable', () => {
```
benjanderson: wow! Good catch!

PR #653, opened 2018-12-06
feat(banner): notification banners

### LINKED ISSUE #329: Feedback blocks
There is a need on a team for an alert that lights up when users are using the application in a non-optimal way (that is not a pop-up requiring an extra click to close) -- something like:

2018-06-15 10_21_00-adobe xd cc

This is an example on MailChimp's style guide:

2018-06-15 10_20_03-feedback _ mailchimp

[http://ux.mailchimp.com/patterns/feedback](http://ux.mailchimp.com/patterns/feedback)

Sorry if I missed something in the style guide that already exists -- the closest I can see are the alert chips. Thank you!

andrew-frueh:
Adds notification banner component.  Thanks @corykon for providing the style sheets that made up the basis for this component.  

Also added a deprecated label to the `fixedTop` property of subNav because it won't set the position correctly if a banner is included in between the navBar and the subnav.  Better to set the fixed position of any content below the navbar in the app instead.

closes #329


projects/cashmere/src/lib/banner/hc-banner.html
```diff
@@ -0,0 +1,6 @@
+<div class="hc-banner {{type}}" [ngClass]="{'hc-banner-dismiss': clickDismiss}" (click)="_bannerClick($event)">
+    <ng-content></ng-content>
+    <span *ngIf="clickDismiss" class="hc-banner-close">
+        <div [ngClass]="type != 'info' ? 'hc-banner-close-icon' : 'hc-banner-close-icon-dark'"></div>
```
corykon: I'd consider dropping the logic in the template here, just use `hc-banner-close-icon`, and then when you need the dark close icon you could target with `.info .hc-banner-close-icon`

andrew-frueh: great suggestion, much cleaner that way.  just pushed a commit that updates as you recommended

Issue #661, opened 2018-12-12
HcSort Question


corykon:
**Describe the bug**
Maybe not a bug...

HcSort by default sorts like this "ABCabc", but I prefer "AaBbCc". Technically the way it sorts is programmatically a correct approach, but I'd rather have things alphabetized regardless of case, and I think that would feel logical to most users. Do others have thoughts?

**To Reproduce**
https://stackblitz.com/edit/table-sort-question
Sort the Name column.

jonest:
As long as it's documented, I'm good with whatever! We had a discussion around this sorting on our team a little while back and we went with the latter as well (`AaBbCc`).

corykon:
Also tweaking so sort happens like this : [Null]123ABC... previously the numbers would come last

PR #662, opened 2018-12-12
fix(table data source): change default sort

corykon:
Sort like this 'AaBbCc' instead of this 'ABCabc'

re #661

### DIFF
--- a/projects/cashmere/src/lib/table/table.md
+++ b/projects/cashmere/src/lib/table/table.md
@@ -129,6 +129,8 @@ If you are using the `HcTableDataSource` for your table's data source, provide t
 directive to the data source and it will automatically listen for sorting changes and change the
 order of data rendered by the table.
 
+The default sort order of values is as follows: "[Null]AaBbCc123".
+
 By default, the `HcTableDataSource` sorts with the assumption that the sorted column's name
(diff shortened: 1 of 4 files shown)


projects/cashmere/src/lib/table/table.md
```diff
@@ -129,6 +129,8 @@ If you are using the `HcTableDataSource` for your table's data source, provide t
 directive to the data source and it will automatically listen for sorting changes and change the
 order of data rendered by the table.
 
+The default sort order of values is as follows: "[Null]AaBbCc123".
```
joeskeen: if you're doing `toLocaleLowerCase()`, isn't the sort order actually `ABC123` (case-insensitive)?  Wouldn't this be the sort order?
`aBc`
`AdE`

joeskeen: Also, don't numbers usually sort before letters in alphabetical order?

corykon: @joeskeen I'll make that change you suggested on the docs. The examples you're showing look correct to me. 

Yup typically numbers are sorted first, but the way the component is written they come second. Not my doing. I suppose we could change that too if we'd like

PR #668, opened 2018-12-22
feat(all): demo site navigation

### LINKED ISSUE #655: Improved Demo Site Navigation
We're in the fortunate position that we have so many great Cashmere components, that they've stretched our current navigation structure past what it was really designed for.  It's pretty poor UX currently that you have to scroll all the way down the page to select a component like "Tabs" and then scroll all the way back up to see its content.  With our long list of components, it's also more difficult to find what you you're looking for.

I think we can take a page from Angular Material's playbook and group our components into sections.  Instead of using vertical tabs for our navigation, I'd recommend that we use a pinned hc-tile on the left with hc-accordion elements containing an hc-list.  So it won't require any new components or changes to components, just reworking of our container pages.

We discussed keeping the vertical tabs, but breaking the components into sections and showing the current section in breadcrumbs.  I think the left tile is overall better UX, because keeps the navigation always visible, and shows all available components in one place.

**Additional context**
This is my mockup for how the new UI would look:
components example components new side nav

I'd propose that we organize the components into the following groups:

**Form Controls**
- Checkbox
- Form Field
- Input
- Radio Button
- Select

**Navigation**
- Breadcrumbs
- Navbar
- Pagination
- Subnav

**Layout**
- Accordion
- Drawer
- List
- Picklist
- Tabs
- Tile

**Buttons & Indicators**
- Button
- Chip
- Icon
- Progress Indicators

**Popups & Notifications**
- Banner
- Modal
- Popover
- Toaster Messages
- Typeform Survey

**Data Table**
- Sort
- Table

andrew-frueh:
Some much needed improvements to the navigation of the demo site.  You'll see that I kept the `hc-tabs` because I didn't want to reinvent the wheel on all its routing functionality, but I hid the left portion and replaced it with my new fixed navigation tile (leveraging our `hc-accordion` component)

For anyone on a Windows computer, can you send me a screenshot of what the scrollbar on the left nav tile looks like when you expand several sections?  Not totally sure whether the tile will expand horizontally to accommodate it or if it will stick out off the edge.

closes #655

### DIFF
--- a/src/app/components/components.component.ts
+++ b/src/app/components/components.component.ts
@@ -6,11 +6,13 @@ import {Subject} from 'rxjs';
 
 @Component({
     selector: 'hc-demo-components',
-    templateUrl: './components.component.html'
+    templateUrl: './components.component.html',
+    styleUrls: ['./components.component.scss']
 })
 export class ComponentsComponent implements OnDestroy {
     docItems: DocItem[];
     thisPage = '';
+    thisCategory = '';
     selectOptions: Array<string> = [];
     private unsubscribe = new Subject<void>();
 
@@ -22,6 +24,11 @@ export class ComponentsComponent implements OnDestroy {
             if (event instanceof NavigationEnd) {
                 if (activatedRoute.firstChild) {
                     this.thisPage = activatedRoute.firstChild.snapshot.params['id'];
+                    this.docItems.forEach(element => {
+                        if (this.thisPage === element.id) {
+                            this.thisCategory = element.category;
+                        }
+                    });
                 }
             }
         });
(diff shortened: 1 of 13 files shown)


src/app/components/components.component.ts
```diff
@@ -30,6 +37,13 @@ export class ComponentsComponent implements OnDestroy {
     // Handle changes to the select component and navigate
     selectUpdate(event: any) {
         this.router.navigate(['/components/' + event]);
+        window.scrollTo(0, 0);
+    }
+
+    // Handle nav changes via the sidebar
+    navUpdate(id: any) {
+        this.router.navigate(['/components/' + id]);
+        window.scrollTo(0, 0);
```
benjanderson: Nice!

src/app/core/document-items.service.ts
```diff
@@ -3,94 +3,115 @@ import {Injectable} from '@angular/core';
 export interface DocItem {
     id: string;
     name: string;
+    // Category options are: 'forms', 'nav', 'layout', 'buttons', 'popups', 'table'
+    category: string;
```
benjanderson: very nice!

benjanderson:
Should we care if it works in IE11? Here is what I see

andrew-frueh:
I'm going to merge this one recognizing that the IE11 problems @benjanderson noticed are still a problem.  But unfortunately it's part of a bigger IE11 challenge we have in many places across the site, so that's probably something we need to bite the bullet and address this month for good.

PR #670, opened 2019-01-04
Convert popover

### LINKED ISSUE #522: (Opt-in) Convert the Popover component

isaaclyman:
Fixes #522.

Hi, I got back from parental leave several weeks ago but I've been catching up on stuff within my team. Anyhow, here's a mixin conversion to keep the ball rolling.

Couple of things:
- There was already a "popover.scss" but it didn't seem to have anything to do with popovers, so I renamed it "menu-panel.scss". Something in the back of my mind says we've discussed this exact file before, but I can't remember how the conversation went or what we decided. Maybe one of you has a better memory of it.
- A couple of the mixins were tricky to name (hc-popover-placed and hc-popover-list-item-middle). Let me know if you have suggestions for better names.
- I don't know why npm decided to HTTPS all our URLs in package-lock. Is that something I should revert?
- Is there anyone else I should request a review from?


isaaclyman:
Ah, I found that conversation and it was with @corykon. Looks like I created `popover.scss` because split buttons automatically have dropdown menus; CDK automatically adds `.hc-menu-panel` to dropdown menus; therefore, `.hc-menu-panel` is applicable to popovers. As far as a home for global styles in this situation, `menu-panel.scss` probably isn't too bad.

PR #674, opened 2019-01-09
Add "Edit on StackBlitz" functionality to Cashmere Docs

### LINKED ISSUE #377: Add ability to edit examples in StackBlitz
In addition to viewing examples and their code, it would be useful for developers to be able to dig into the example and test out changes.

In the top right of each example tile, add a "Edit in StackBlitz" button that opens a new tab with the example loaded up.  See the prototype below for an illustration:

https://github.com/HealthCatalyst/Fabric.Cashmere/blob/dev/CHANGELOG.md

joeskeen:
Fixes #377

image

This is a complete refactor of the examples for the Cashmere documentation, in order to support dynamically generated StackBlitz projects.

Some things to note:

* Strict naming conventions for everything in the examples has been applied.  It is critical that those naming conventions are used so that both the live example on the docs site and the dynamic StackBlitz project both keep working.  I have added some checks in the build process that should help identify if anything isn't following the conventions.
* I added a new guide to the Cashmere docs site entitled "Working with Examples", which explains the entire process of adding a new example to Cashmere docs.
    image
* Support for multi-component examples has been added by allowing you to create a module for the example.  This is particularly useful for examples such as Modal which uses multiple components.
* General refactoring has been done to ensure that the dynamically generated StackBlitz project adheres to good practices (so as to encourage consumers of Cashmere to do the same)

One issue with the dynamic StackBlitz projects is that although we are including the assets in the project, StackBlitz currently won't serve them up properly (see https://github.com/stackblitz/core/issues/617).
image

A BIG thanks to @bskeen , without his help over Nov and Dev Open Space we wouldn't be to this point by now.

Please take some time to pull down this branch and check to make sure the examples and StackBlitz projects work as expected.


projects/cashmere-examples/src/lib/cashmere.module.ts
```diff
@@ -1,92 +1,7 @@
...
-        SelectModule,
-        SortModule,
-        SubnavModule,
-        TableModule,
-        TabsModule,
-        TileModule,
-        ToasterModule,
-        TypeformSurveyModule
-    ]
+    exports: [GeneratedCashmereModule]
```
benjanderson: ohhh Nice!

guides/working-with-examples.md
```diff
@@ -0,0 +1,47 @@
...
+Adding meaningful examples can greatly enrich the documentation of components in the Cashmere library. Here are some things to remember when creating or modifying examples.
+
+## Examples files
+
+The following files are used in the creation of examples for the Cashmere documentation website.
+
+### Files you may need to modify
+
+-   `projects/cashmere-examples/src/lib/{EXAMPLE_NAME}` - the folder containing all files for an example
+-   `src/app/core/document-items.service.ts` - the file that adds the example to the appropriate component page in the docs site
```
benjanderson: This sentence is a little hard to read. perhaps:
> Service listing all the example components in the docs site

guides/working-with-examples.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Files used in the build process (which you shouldn't need to modify)
+
+-   `src/assets/docs/examples` - the output directory for StackBlitz example project files
+-   `projects/cashmere-examples/src/lib/cashmere.module.ts` - if your example contains a module file, you will need to import this CashmereModule
```
benjanderson: This probably belongs in the "Files you may need to modify"

guides/working-with-examples.md
```diff
@@ -0,0 +1,47 @@
...
+-   `projects/cashmere-examples/src/project-template` - the folder containing Angular CLI project files used in the StackBlitz
+-   `scripts/build-examples.ts` - the script used to generate the support files needed for the docs site and StackBlitz
+
+## Creating a new example
+
+_Please pay close attention to the naming conventions used here, otherwise your example may not work on the documentation site and/or StackBlitz._
+
+1. Name the example (the [hardest part](https://bit.ly/2VI8ZI5)). Figure out what you want the title of the example to be on the documentation site.
+2. Derive the example name from the example title by converting it to kebab case (for instance, if your example is titled "Format Hard Drive", the example name would be `format-hard-drive`)
+3. Create a folder under `projects/cashmere-examples/src/lib/` named your example name, i.e. `projects/cashmere-examples/src/lib/format-hard-drive`
```
benjanderson: nitpicky, but its usually called directory, not folder

guides/working-with-examples.md
```diff
@@ -0,0 +1,47 @@
...
+
+Adding meaningful examples can greatly enrich the documentation of components in the Cashmere library. Here are some things to remember when creating or modifying examples.
+
+## Examples files
+
+The following files are used in the creation of examples for the Cashmere documentation website.
+
+### Files you may need to modify
+
+-   `projects/cashmere-examples/src/lib/{EXAMPLE_NAME}` - the folder containing all files for an example
```
benjanderson: nitpicky, but its usually called directory, not folder

guides/working-with-examples.md
```diff
@@ -0,0 +1,47 @@
...
+2. Derive the example name from the example title by converting it to kebab case (for instance, if your example is titled "Format Hard Drive", the example name would be `format-hard-drive`)
+3. Create a folder under `projects/cashmere-examples/src/lib/` named your example name, i.e. `projects/cashmere-examples/src/lib/format-hard-drive`
+4. Create the component file `{EXAMPLE_NAME}-example.component.ts`, i.e. `projects/cashmere-examples/src/lib/format-hard-drive/format-hard-drive-example.component.ts`
+5. In the example component, export the example class `{EXAMPLE_NAME}ExampleComponent`, i.e. `FormatHardDriveExampleComponent`
+6. Add the corresponding HTML template file and (only if needed) the SCSS style file and register them with the component, i.e. `format-hard-drive-example.component.html`
+7. Register your example with the appropriate component(s) by adding the example name to the list of examples in `src/app/core/document-items.service.ts`. For example, you may add your example to the SubNav component examples by adding it thus:
+    ```
+    {id: 'subnav', name: 'Subnav', examples: ['subnav-overview', 'format-hard-drive']},
```
benjanderson: :heart: loving these examples, it makes it easy to follow :heart:

guides/working-with-examples.md
```diff
@@ -0,0 +1,47 @@
...
+    ```
+8. If you are creating an example to a new Cashmere component, make sure that the corresponding module is added to `src\app\shared\cashmere.module.ts`. The example build process will use this file to generate the `CashmereModule` for both the example modules as well as the StackBlitz project.
+9. Implement your example.
+
+### Creating a more complex example
+
+Sometimes your example needs more than a single component, such as when you are demonstrating modals. In this case you can create an example module file named `{EXAMPLE_NAME}-example.module.ts` which will declare all components used in your example. When doing this it is important to do the following in your module file:
+
+-   Make sure the exported module name is `{EXAMPLE_NAME}ExampleModule`
```
benjanderson: Make sure the exported module **class** name is

projects/cashmere-examples/src/lib/button-primary/button-primary-example.component.scss
```diff
@@ -0,0 +1,3 @@
+button {
+    margin: 10px;
```
benjanderson: This seems to be a very common thing to add to each of the examples. What was the reason to add this to so many style sheets? Could it be something that should be pushed to a more global style sheet?

projects/cashmere-examples/src/project-template/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "ng": "ng",
+    "start": "ng serve",
+    "build": "ng build",
+    "test": "ng test",
+    "lint": "ng lint",
+    "e2e": "ng e2e"
+  },
+  "private": true,
+  "dependencies": {
+    "@angular/animations": "6.1.7",
```
benjanderson: I am guessing this `package.json` is used to send to stack blitz. Does the angular versions need to match? If so is there a way we can match the root `package.json`? We may need to add documentation on how this works and what to do when we update package versions...

projects/cashmere-examples/src/project-template/src/app/cashmere.module.ts
```diff
@@ -0,0 +1,61 @@
...
+        SelectModule,
+        SubnavModule,
+        TableModule,
+        TabsModule,
+        TileModule,
+        ToasterModule,
+        TypeformSurveyModule
+    ]
+})
+export class CashmereModule {}
```
benjanderson: as we already have a cashmere module in the root project, would it make sense to name this CashmereExamplesModule?

guides/working-with-examples.md
joeskeen: Actually, no.  All this file does is import the generated Cashmere module, which is copied from `src/app/shared/cashmere.module.ts` to make it so there is only one CashmereModule file needing to be touched.
I probably should, however add `src/app/shared/cashmere.module.ts` to the list of files you may need to modify.

projects/cashmere-examples/src/project-template/package.json
joeskeen: good point.  I should probably have this file generated from the root `package.json`

joeskeen:
> Please keep reformatting contents to a minimum in the future. It made it quite a bit longer to review
@benjanderson isn't this what Prettier should be doing for us?  When I was preparing to check in I ran the linter and Prettier said that tons of files were malformatted.  I ran `prettier --write` on all the files to fix this.

guides/working-with-examples.md
joeskeen: noted

guides/working-with-examples.md
joeskeen: noted

guides/working-with-examples.md
joeskeen: noted

guides/working-with-examples.md
joeskeen: noted

projects/cashmere-examples/src/lib/button-primary/button-primary-example.component.scss
joeskeen: We could promote it to `projects\cashmere-examples\src\project-template\src\styles.scss`...

I just want to remove as much smoke and mirrors as I can for the examples.  As Cashmere buttons don't have margin, but you almost never use them without margin, users may assume after looking at the examples that the margins are a part of Cashmere buttons... but they are not.  I don't think it hurts to put this in the example stylesheet if you will have to do this in order to achieve the style displayed.

guides/working-with-examples.md
joeskeen: Fixed in latest push

projects/cashmere-examples/src/project-template/package.json
joeskeen: Just curious, why the `angular.json`?  In order to support new features added to the docs site?  The `angular.json` file in the StackBlitz won't necessarily have all the same config as the docs site, as it is only for StackBlitz.  I feel like changes to that should be more manual.

joeskeen: I suppose we could auto-insert whatever is in the `"assets"`, `"styles"`, and `"scripts"` from the docs site project, though...

benjanderson: I'm ok with the way it is

benjanderson:
I approved even though there were minor wording changes, I trust you

PR #677, opened 2019-01-22
datepicker and date-range components

benjanderson:
This is a clone of material date picker https://github.com/angular/material2 and https://github.com/ashishgkwd/ngx-mat-daterange-picker but highly modified to match cashmere standards. Here are the following things I did

- with the help of @andrew-frueh I styled them cashmere style
- renamed from `MAT*` to `HC*`
- organize file structure
- removed ability to change color
- added a today button
- lots of documentation


projects/cashmere/src/lib/datepicker/datepicker.component.scss
```diff
@@ -0,0 +1,296 @@
+@import '../sass/colors';
+@import '../sass/typography';
+@import '../sass/functions';
+@import '../sass/mixins';
+
+.hc-datepicker-content {
+    width: 296px;
```
andrew-frueh: Looks like our container isn't quite big enough for the content with padding (see attached).  If you increase width to 310 it should account for the 8px padding being applied.
screen shot 2019-01-28 at 10 29 57 am

projects/cashmere/src/lib/datepicker/datepicker.component.scss
```diff
@@ -0,0 +1,296 @@
...
+    height: 354px;
```
andrew-frueh: I think we could use a little extra padding on the bottom of the container too - maybe height: 360px

projects/cashmere/src/lib/datepicker/datepicker.component.scss
andrew-frueh: If this change is made - it looks like the `.hc-datepicker-content .hc-calendar` is getting set inline somewhere, but that should also change to 310px so the right side of the header is aligned with the wider width.  This value might be calculated automatically, I just wasn't sure where.

projects/cashmere/src/lib/datepicker/calendar-body/calendar-body.component.html
```diff
@@ -0,0 +1,53 @@
+<!--
+  If there's not enough space in the first row, create a separate label row. We mark this row as
+  aria-hidden because we don't want it to be read out as one of the weeks in the month.
+-->
+<tr *ngIf="_firstRowOffset < labelMinRequiredCells" aria-hidden="true">
+    <td class="hc-calendar-body-label" [attr.colspan]="numCols" [style.paddingTop]="_cellPadding" [style.paddingBottom]="_cellPadding">
```
andrew-frueh: I think we decided on our review call to remove the calendar body label because it's redundant with the dropdown above and will save us some vertical space.

projects/cashmere/src/lib/datepicker/datepicker.component.scss
benjanderson: Thank you for pointing this out, It was a simple style sheet change

projects/cashmere/src/lib/datepicker/datepicker.component.scss
benjanderson: I ended up removing the height and width and let the elements size themselves and things "fit" much better

benjanderson: nevermind, the month and year views need the height and width :P

projects/cashmere/src/lib/datepicker/calendar-body/calendar-body.component.html
benjanderson: Looking at it, it didn't save us any vertical space but it was redundant so I removed it

projects/cashmere/src/lib/datepicker/datepicker.component.scss
andrew-frueh: Datepicker container is looking much better!  Now that we got rid of the month label though, it's a bit long.  For `.hc-datepicker-content`, 330px looked about right to me.

projects/cashmere/src/lib/date-range/picker-overlay/picker-overlay.component.scss
```diff
@@ -0,0 +1,41 @@
...
+    display: flex;
+    justify-content: space-around;
+    min-width: 650px;
+}
+
+.hc-date-range-calendar-item {
+    flex-basis: 33.33%;
+    min-width: 210px;
+    padding: 1em;
+    font-family: Roboto, 'Helvetica Neue', sans-serif;
```
andrew-frueh: Just noticed this one.  Should be open sans, not roboto - that's Angular Material showing through :-)

projects/cashmere/src/lib/datepicker/calendar/calendar.component.scss
```diff
@@ -0,0 +1,145 @@
...
+
+.hc-calendar-next-button::after {
+    border-right-width: $hc-calendar-prev-next-icon-border-width !important;
+    transform: $hc-calendar-next-icon-transform;
+}
+
+.hc-calendar-today-button.hc-md.hc-button.hc-link {
+    font-weight: 600;
+    color: $primary-brand;
+    @include fontSize(12px);
```
andrew-frueh: Go ahead and remove this `@include fontSize(12px);` line.  We want the font size of "Today" to match the size of the label on the left.

projects/cashmere/src/lib/date-range/picker-overlay/picker-overlay.component.scss
benjanderson: Wow! good catch

projects/cashmere/src/lib/datepicker/calendar/calendar.component.scss
benjanderson: done

PR #681, opened 2019-01-28
Navbar 671

jacquie:
This fixes some of the issues in bug 671.


projects/cashmere-examples/src/lib/cashmere-example.module.ts
```diff
@@ -0,0 +1,358 @@
+
```
benjanderson: @joeskeen is this supposed to be checked in?

projects/cashmere/src/lib/navbar/navbar.component.spec.ts
```diff
@@ -43,38 +43,43 @@ describe('NavbarComponent', () => {
     it('should create', () => {
```
benjanderson: Very nice tests!

projects/cashmere-examples/src/lib/cashmere-example.module.ts
joeskeen: No, this should not be checked in.  It should be deleted.

jacquie: Thanks - I was wondering!

jacquie: Should it be added to .gitignore?

jacquie: I removed it

projects/cashmere/src/lib/navbar/navbar.component.html
```diff
@@ -1,22 +1,22 @@
...
-    <div class="hc-navbar-link-container">
+    <div #navlinks class="hc-navbar-link-container">
         <ng-content select="hc-navbar-link"></ng-content>
         <div class="hc-navbar-more-links" title="More" *ngIf="_collapse" [hcPopover]="navbarMore" popperPlacement="bottom">
             More<hc-icon fontSet="fa" fontIcon="fa-chevron-down" hcIconSm></hc-icon>
         </div>
     </div>
 
-    <div class="hc-navbar-right-container">
+    <div #rightcontainer class="hc-navbar-right-container">
```
jonest: Should we prefix any of these ids? I could see a possible collision on this one

projects/cashmere-examples/src/lib/cashmere-example.module.ts
joeskeen: You shouldn't need to add it to .gitignore.  The new example generation process doesn't create this file.

Also, don't forget to push!

projects/cashmere/src/lib/navbar/navbar.component.html
joeskeen: This is an Angular variable name, and is scoped into the component.  There won't be any collisions (that would result in a compile error).

andrew-frueh: That's a good call @jonest - maybe #hc-rightcontainer

jonest: Doh! Good call @joeskeen :) I completely mixed the two up :) I saw the `#` and thought "jQuery ID tag!" 😆

joeskeen: Again, not necessary - it's just a template variable.  I don't think you can even put hyphens in those.

Issue #680, opened 2019-01-26
Dates sort as strings using HcSort


corykon:
**Describe the bug**
Dates sort like there string using HcSort. Guessing I broke this when I worked on #661.

**To Reproduce**
1) Go here: https://stackblitz.com/edit/angular-a5xpcy
2) Sort by the date column

**Screenshots**
![image](https://user-images.githubusercontent.com/12416432/51779518-138a7d00-20c5-11e9-95e9-a3ac0f0c4f81.png)

PR #682, opened 2019-01-28
Sort issues

### LINKED ISSUE #679: Sort Demo doesn't work
**Describe the bug**
Sorting doesn't work on the table demo page.

**To Reproduce**
Go here https://cashmere.healthcatalyst.net/components/table/examples, can also look at the stackblitz.

**Expected behavior**
Sorting should be there and should work

**Screenshots**
No sort indicators:
image

**Additional context**
I think it's broken because the SortModule is missing or isn't imported properly. I was able to fix on stackblitz by adding the module to cashmere.module.ts

corykon:
resolve #679 & #680


projects/cashmere-examples/src/lib/cashmere-example.module.ts
```diff
@@ -0,0 +1,358 @@
+
```
benjanderson: This shouldn't be checked in, it is part of the autogenerated files. 

https://github.com/HealthCatalyst/Fabric.Cashmere/pull/681#pullrequestreview-197126718

jacquie: I asked about adding it to .gitignore and @joeskeen said no?

PR #683, opened 2019-01-29
fix(hc-banner): flex start justify content

jonest:
There are some flex box issues with IE that can be solved with setting the justify-content to
flex-start

#### Before
image

#### After
(Ignore the small chunk of scroll bar that's covering some of the banner, there are tons of other IE issues at the moment...)
image

Thanks @YaroBear for finding this solution and helping me out!

Fixes issue #684


jonest:
FYI I checked that Chrome still looks the same, as well as Firefox (both on Windows)

PR #688, opened 2019-02-01
fix(navbar): navbar-brand min-width

### LINKED ISSUE #678: When using tables styles, icons should be base64 encoded instead of relying on FontAwesome
Title pretty much says it

Base64

Tables.scss

image

andrew-frueh:
prevent navbar-brand from shrinking below 55px width

closes #678


andrew-frueh:
Thanks for the IE help on this one @corykon

PR #730, opened 2019-02-14
fix(date-range): options set after nginit not taking effect

### LINKED ISSUE #729: Date range options cant change after "ngOnInit"
**Describe the bug**
Options set after "ngOnit" aren't taking effect. Since this is a fairly new component I wanted to get this out before anyone adopts this and finds the bug

benjanderson:
BREAKING CHANGE: moved date binding from options to selectedDate

fix #729


projects/cashmere-examples/src/lib/date-range/date-range-example.component.ts
```diff
@@ -23,9 +23,7 @@ export class DateRangeExampleComponent implements OnInit {
         this.options = {
             presets: this.presets,
             format: 'mediumDate',
-            range: {fromDate: today, toDate: today},
             applyLabel: 'Apply',
-            // excludeWeekends:true,
```
joeskeen: should this be removed?

PR #741, opened 2019-02-20
Breadcrumb query parameters

### LINKED ISSUE #740: Breadcrumbs do not preserve query parameters
**Describe the bug**
We're trying to keep track of some state using query parameters, but the breadcrumbs don't seem to work with them.

I've tried messing around with the library, but nothing seems to work correctly. Hoping someone else has some thoughts

image

jonest:
Propogate the query parameters when using the breadcrumb component

fix #740

### DIFF
--- a/projects/cashmere/src/lib/breadcrumbs/breadcrumbs.component.html
+++ b/projects/cashmere/src/lib/breadcrumbs/breadcrumbs.component.html
@@ -1,12 +1,12 @@
 <ol class="breadcrumb">
     <li *ngFor="let breadcrumb of _breadcrumbs; let last = last" class="breadcrumb-item">
-        <a *ngIf="!last" [routerLink]="[breadcrumb.url]" queryParamsHandling="preserve" [innerText]="breadcrumb.label"></a>
+        <a *ngIf="!last" [routerLink]="[breadcrumb.url]" queryParamsHandling="{{queryParamsHandling}}" [innerText]="breadcrumb.label"></a>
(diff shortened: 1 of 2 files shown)


projects/cashmere/src/lib/breadcrumbs/breadcrumbs.component.html
```diff
@@ -1,6 +1,6 @@
 <ol class="breadcrumb">
     <li *ngFor="let breadcrumb of _breadcrumbs; let last = last" class="breadcrumb-item">
-        <a *ngIf="!last" [routerLink]="[breadcrumb.url, breadcrumb.params]" [innerText]="breadcrumb.label"></a>
+        <a *ngIf="!last" [routerLink]="[breadcrumb.url" queryParamsHandling="preserve" [innerText]="breadcrumb.label"></a>
```
benjanderson: I would really like to see this as an option in the component, users of the library may not want this behavior

jonest: Good call @benjanderson !

jonest: I also made a typo there and left off the `]`

jonest:
Great suggestion @benjanderson! I added some code to handle it. Let me know your thoughts!

jonest:
> Looks great! I think this a a good incremental change :)

Yup! There's still some other things we _could_ add at some point, like the fragment usage, but one thing at a time, right? :D

PR #751, opened 2019-02-26
docs(guide): update guide docs to use Cashmere packages

Br00t4L17y:
These changes take into account that not all users or contributors have globally installed npm
packages. This also helps negate differences between versions of globally and project specific
packages.


benjanderson:
This is awesome! When setting people of with cashmere I have had to tell them to install the angular cli. This will help people get up and going quicker

PR #755, opened 2019-03-02
feat(tabs): defaultTab property

### LINKED ISSUE #754: Tabs - ability to set the default tab
Adding this on behalf of @jonest and the Ops Console team - who are creating their tabs dynamically and don't want a tab clicked event thrown on their first tab automatically.

Potentially add a parameter called `defaultTab` that a user can set to tell the app what tab to select on display.  And if that parameter is set to `null`, no tab will be selected by default.

Tabs

**Additional context**
@jonest - feel free to add any additional color commentary that I may have missed.

andrew-frueh:
Adds a defaultTab property to hc-tab-set which includes the ability to set to "none" - which prevents any tab from being selected initially.  How does this work to address the issue you ran into @jonest?

closes #754

### DIFF
--- a/projects/cashmere-examples/src/lib/tabs-vertical/tabs-vertical-example.component.html
+++ b/projects/cashmere-examples/src/lib/tabs-vertical/tabs-vertical-example.component.html
@@ -1,5 +1,5 @@
 <div class="tab-demo">
-    <hc-tab-set>
+    <hc-tab-set defaultTab="1">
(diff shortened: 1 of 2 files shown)


jonest:
I had a new thought just now: would it make more sense to have the default tab be a directive placed on one of the tabs instead of having to choose it on the tab set component via number?

jonest:
Merged this and submitting a pull request to get a new package so we can fix our bug. Thanks!

PR #772, opened 2019-03-14
fix(appswitcher): thumbnail img sizing

andrew-frueh:
Corrects app icons that were being stretched if they weren't square.  Adds an app switcher parameter to set the height to something other than 100px if needed.  Also fixes an issue where the names weren't being centered correctly under the icons.


benjanderson:
Great job @andrew-frueh

PR #783, opened 2019-04-04
502 cdk popup

corykon:
~~Don't want to merge this in quite yet, but would like some eyes on it.~~ This PR creates a new popover component based on the CDK. The code was adapted from another 3rd party component (that also uses the CDK). Let me know what you think about the interface. Am I missing anything?

~~Still to be done/decided:~~
- ~~What to actually name the new component, if we leave the old one around but depreciated? hc-pop or hc-popper or hc-popup or hc-overlay~~
-~~Swap out the old popover in the navbar, app switcher, other examples in the docs~~
- ~~To depreciate or bridge the old popover? Bridging seems like a significant lift and somewhat risky/hard to get right, but take a look and let me know what you think. Depreciating in my mind would include: console.warn when using old to let devs know, make it clear in the docs site that the old one is depreciated, and also let current users of the library know via email/slack announcements~~


benjanderson:
Great job Cory! This looks amazing.

Before we remove the old popover we for sure need to replace any instances of it from other components.  That said, I have been reviewing it this morning and have been going back and forth on naming and what to do with the old popover. If this were a large library with a large base of consumers, we would be forced to migrate it slowly. However, with just a handful of teams using it and this change has been coming for quite a while, I would vote to remove the old component. This would be a very welcome breaking change for me.

I would like to see to get an opinion from others if this would be too dramatic of a change? Does anybody use the old popover extensively that it would be too painful to switch when upgrading cashmere?

andrew-frueh:
I think we've been pretty mindful of introducing breaking changes to keep them to a minimum, but this might be one of those instances where it's justified.  It'll just be very important message this clearly to our users that the new number release contains a breaking change on popover.  But I'm with @benjanderson that I'd rather see this added as a breaking change than have two popover components existing side by side in the library.

corykon:
I'm going to move forward then with totally pulling out the old popover component.

andrew-frueh:
I think that's the right thing to do @corykon .  Btw - just had a chance to check out your new component, and it looks great!  Love the tooltip callout too, that's a great idea.  The only thing I miss in this new version is not having the little arrow off of the popover that points to the item which triggered it.  I think it has some UI value in terms of anchoring the popover visually to its parent, but I also recognize it has some limitations.  So I'm okay with letting it go, but I at least wanted to run it by you to see what you thought.

corykon:
@andrew-frueh Certainly can be done, and I agree it's a nice-to-have in a lot of cases. I'll look into in after I get everything else cleaned up.

corykon:
@andrew-frueh That is on purpose actually, but it is configurable via `[autoFocus]`. Feels a little different on menu like that though. I'll make a change to have it turned off by default!

It is a nice-to-have for keyboard accessibility/saving clicks, but is really meant more for when you have an input element inside of a popover, which I don't have demoe anywhere.

benjanderson:
I am not the biggest fan of animating something like the app switcher. Looking at both google and office 365, their app popup just appears without any motion or fade. @andrew-frueh what is your opinion?

andrew-frueh:
It does look cool - but you're probably right @benjanderson.  For a utility item like the app switcher, it should probably be instant on.

package.json
```diff
@@ -54,7 +54,6 @@
         "markdown-it-container": "^2.0.0",
         "ngx-highlightjs": "^3.0.0",
         "npm-font-open-sans": "^1.1.0",
-        "popper.js": "^1.12.9",
```
benjanderson: woohoo! Good riddance

projects/cashmere/src/lib/pop/popover.component.scss
```diff
@@ -0,0 +1,135 @@
+@import '~@angular/cdk/_overlay';
```
benjanderson: It is a little hard to see the arrow and top of the popup, especially with our standard bg. Maybe a border? This is similar styling as the last popover but I wanted to point it out. Not a huge deal

image

corykon:
@benjanderson @andrew-frueh Fixed the animation and border issues. If there aren't any objections I'm going to merge after the build goes through

PR #825, opened 2019-04-15
Popover + Unit tests and minor fixes

corykon:
Includes breaking change, removal of the old popover


corykon:
We ignoring this guy ?
image

PR #830, opened 2019-04-19
feat(toasts): custom component toasts

### LINKED ISSUE #829: Toast Supports Custom Components
Currently hcToast only supports `TemplateRef` for custom toasts.

@isaaclyman has requested that it support a component as an input as well (similar to modal) so that multiple custom toasts can be up at once with different variables displaying in each.

-   hcToast

andrew-frueh:
Allows for custom toasts with a component as the input. @isaaclyman - take a look at the updated toasts example for the version that uses a component.  It shows how you can use a component as the input in your custom toast, and set values in that component instance in the generated toast.  Let me know if this gets you what you need for your use case.

closes #829


andrew-frueh:
Thanks for all the help today @joeskeen with getting CDK portal incorporated.  Let me know if you see anything else in the update you'd do differently.

PR #855, opened 2019-05-07
User/joe/dev mode

joeskeen:
This finally answers the problem of library development for Cashmere.

As a developer, you don't have to do anything different, just `npm start`.

Under the hood, the TSConfig is patched to point the `cashmere` and `cashmere-examples` libraries to the source locations (as opposed to the `dist` folder as it does during release builds).  This enables you to make changes (almost) anywhere in the repo and have the browser live-reload in a couple of seconds, as opposed to having to rebuild the library after each change.

And to keep our PRs clean, a pre-commit hook was added that automatically sets things back to release mode when you commit.  Easy peasy.

image


andrew-frueh:
@joeskeen - should we document these updated run calls in the "Guides" docs?

joeskeen:
@andrew-frueh good catch - it's probably a good time to revisit the contribution guide as things are much simpler now.

JaredDraper:
love it!!!

PR #858, opened 2019-05-07
build(travis): push npm package to @next on dev builds

joeskeen:
Enough said :P

But seriously, I won't know if this works until we actually get this into the dev branch.


scripts/publish-next.sh
```diff
@@ -0,0 +1,22 @@
...
+npm run build
+
+# Go to Cashmere build output directory
+cd $(dirname $0)/../dist/cashmere
+
+# Make it possible to install this version using `npm i -S @healthcatalyst/cashmere@next`
+npm config set tag next
+
+# Set version to timestamped beta version
+npm version 0.0.0-beta.$(date +"%Y%m%d%H%M%S") --git-tag-version false
```
benjanderson: I think the proper semver way to do this is 
`npm version prerelease --preid=beta`
https://docs.npmjs.com/cli/version

joeskeen: I'm a little fuzzy on how that would work - do we keep the current version number in package.json? I thought that normally the version number is auto-gen'd at release time and not committed to git.
Would it be something like this?
version 6.0.0 is released on `master`, then version 6.0.0.beta is released after on `dev`?  That would be a bit backwards I would think.
I looked at using semantic-release, but that only works with git tags, so we would be tagging every dev release and our master releases would be jumping numbers a lot more, which I don't think we want to do. (also semantic-release doesn't work with a branch named `dev`, we would have to use the name "next" instead)
I think keeping a timestamp in the version number somewhere would be helpful, regardless of how we decide to version these prerelease packages.

joeskeen:
@benjanderson I have fixed the script to increment the version as we discussed.  In my tests this is how it worked:
Current Cashmere version (@latest) is `5.1.4`
The script stamped the package with version `5.1.5-dev.20190513172517`
That way they are connected to both the version and the date.
👍 ?

PR #868, opened 2019-05-20
864 scroll nav

corykon:
New component that Care Management and Pop Builder are hoping to share


corykon:
Look like the build failure is just that publishdev step:
```
npm ERR! code ENEEDAUTH
npm ERR! need auth auth required for publishing
npm ERR! need auth You need to authorize this machine using `npm adduser`
npm ERR! A complete log of this run can be found in:
npm ERR!     /home/travis/.npm/_logs/2019-05-20T23_48_54_431Z-debug.log
The command "./scripts/travis-script.sh" exited with 1.
```

projects/cashmere/src/lib/scroll-nav/nav/scroll-nav.component.scss
```diff
@@ -0,0 +1,31 @@
...
+.hc-scroll-nav-links-container {
+    margin-right: 30px;
+
+    .hc-scroll-nav-link {
+        padding: 10px 15px;
+        color: $offblack;
+        font-weight: 500;
+        background-color: $white;
+        cursor: pointer;
+
```
andrew-frueh: Looks awesome @corykon and @Techgeekster - love it!  Only small suggestion I have is to reduce the font-size of the `hc-scroll-nav-link` to 14px (instead of 16px).  Otherwise, visually it's great and functionality works exactly as I would have expected.

joeskeen:
> Look like the build failure is just that publishdev step:
> 
> ```
> npm ERR! code ENEEDAUTH
> npm ERR! need auth auth required for publishing
> npm ERR! need auth You need to authorize this machine using `npm adduser`
> npm ERR! A complete log of this run can be found in:
> npm ERR!     /home/travis/.npm/_logs/2019-05-20T23_48_54_431Z-debug.log
> The command "./scripts/travis-script.sh" exited with 1.
> ```

Sorry @corykon ... I really don't know what's causing that, but I'm trying to look into it...

benjanderson:
For those like me who didn't quite know what it was here is a gif I made

benjanderson:
It tragic that it doesn't work with the `<a id="link">` http://something.com#link https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Linking_to_another_section_on_the_same_page

Is this something you considered? How hard would it be to integrate automatically scrolling to a section when a user lands on the page? Its not a deal breaker but I use it all the time

projects/cashmere/src/lib/scroll-nav/nav/scroll-nav.component.spec.ts
```diff
@@ -0,0 +1,92 @@
+import {async, ComponentFixture, TestBed} from '@angular/core/testing';
```
benjanderson: Nice tests!

projects/cashmere/src/lib/scroll-nav/nav/scroll-nav.component.ts
```diff
@@ -0,0 +1,58 @@
...
+            if (i === 0) {
+                e.classList.add(this.ACTIVE_CLASS);
+            }
+            this.renderer.listen(e, 'click', () => {
+                this.onNavClick(e);
+            });
+        });
+    }
+
+    public setActiveClassById(id: string): void {
```
benjanderson: Should this have documentation?

projects/cashmere/src/lib/scroll-nav/nav/scroll-nav.component.ts
```diff
@@ -0,0 +1,58 @@
...
+        const link = this._links.find(e => e.getAttribute('hcScrollLink') === id);
+        if (!link) {
+            console.warn(`Failed to mark active class. Could not find the element with the data target for id: ${id}.`);
```
benjanderson: I would rather throw an error. It seems like this is the sort of thing that would pass QA and become tech debt forever because of a bad setup.

corykon:
> It tragic that it doesn't work with the `<a id="link">` http://something.com#link https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Linking_to_another_section_on_the_same_page
> 
> Is this something you considered? How hard would it be to integrate automatically scrolling to a section when a user lands on the page? Its not a deal breaker but I use it all the time

Definitely considered it! That's the reason use IDs as my targets. It is possible to navigate via those bookmark IDs if you're already on a particular page: 

anchorBookmark


You will run into problems navigating from another page though, like you mentioned, because angular router wants to do different things with the '#'. I think you could make a tweak in the router to allow for that though?

Either way I'm thinking that may a good enhancement for a v2.

projects/cashmere/src/lib/scroll-nav/nav/scroll-nav.component.spec.ts
corykon: credit goes to @Techgeekster !

Techgeekster:
Maybe if we used something like this, we could grab what is currently in the route and then act on it.

`constructor(
    route: ActivatedRoute
  ) {
    route.parent.params.subscribe((p) => this.navLocation= p["navLocation"]);
  }`

Issue #877, opened 2019-05-28
ScrollNav Upgrades/Fixes


corykon:
**Describe the problem this feature request is solving**
Ran into a few issues implementing the scroll nav.

- Was getting an error because I had an outdated version of the cdk in my app. Would be good to mention a min version for the CDK somewhere
- Binding to the `hcScrollLink` @Input doesn't work, only takes string literals
- The code for making the last section the height of the container is imperfect. Default the setting to false.
- Sometime I need to manually update/refresh the currently active section. Add a public function for this

**If any, which components will have to change, and how**
Will update scrollnav, no breaking changes.

benjanderson:
> Was getting an error because I had an outdated version of the cdk in my app. Would be good to mention a min version for the CDK somewhere

I know we have peer dependencies listed, https://github.com/HealthCatalyst/Fabric.Cashmere/blob/dev/projects/cashmere/package.json
It should give a warning if they are not met

What version of the cdk were you using?

corykon:
@benjanderson Ah ok. So I was on 6.x, but because of the new scroll nav component you need at least 7.x. Should I just update the version number in the peer dependency section of package.json?

PR #880, opened 2019-05-28
877 scroll nav updates

corykon:
resolves issues in #877


corykon:
Build failure is just this guy: (which we're ignoring until someone has an opportunity to fix)
image

PR #888, opened 2019-06-05
docs(issue template): simplify existing templates and add doc template

isaaclyman:
My goals here are to:

1. Significantly reduce friction for Cashmere outsiders who want to report issues
2. Make documentation a first-class citizen by inviting Cashmere users to recommend improvements to it

Potential objections: I understand the impulse to ask for as much information as possible from issue creators, but in my experience this makes people feel like submitting an issue isn't worth the trouble. Most software bugs and feature requests can be effectively described in a sentence or two. If a bug report is severely lacking in detail or we're unable to reproduce it, we should take a moment to follow up and ask for specific clarification.

In short, if we want participation from the Health Catalyst community, we should make it easy for them to participate.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,32 +1,23 @@
 ---
 name: Bug report
-about: Create a report to help us improve
+about: Report a bug or unexpected behavior
 ---
 
-**Describe the bug**
-A clear and concise description of what the bug is.
+**Summary**
 
-**To Reproduce**
-Reproduce it on stackblitz fork
-https://stackblitz.com/edit/cashmere-demo?file=app%2Fhome%2Fhome.component.ts
-_OR_
-Steps to reproduce the behavior:
+*Let us know what went wrong. The rest of this form is optional.*
 
-1.  Go to '...'
-2.  Click on '....'
-3.  Scroll down to '....'
-4.  See error
+**Reproduction**
 
-**Expected behavior**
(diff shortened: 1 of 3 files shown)


benjanderson:
I agree with simplifying the templates. I first created the templates with the vision of helping people submit issues and give proper detail, but as I used the templates I found them cumbersome and almost always reduced the template down to something similar to what your were proposing. Out of curiosity I looked up all the issues I created and sure enough, I never used the template. 

https://github.com/HealthCatalyst/Fabric.Cashmere/issues?utf8=%E2%9C%93&q=is%3Aissue+author%3Abenjanderson+

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,32 +1,23 @@
...
+1. 
+2. 
+3. 
+4. 
 
-**Browser:**
-
--   Browser [e.g. chrome, safari]
--   Version [e.g. 22]
+Stackblitz reproduction (you can fork from the Stackblitz link in a component example):
```
benjanderson: Great change! We were using the old stackblitz that I created to help people repro issues, but now we have the "edit in stack" button so it really iesn't needed.  You may want to be more explicit:

Stackblitz reproduction (optional). 
Click the "Edit in Stackblitz" button next to the most relevant sample and fork the generated angular template

isaaclyman: Updated, thanks for the suggestion.

PR #900, opened 2019-06-27
fix(table): remove onpush change detection

isaaclyman:
Putting this up for discussion.

I've had multiple issues where the content of hc-table cells doesn't update when bound data on the component changes. My hypothesis is that this is because hc-table uses ChangeDetectionStrategy.OnPush, which (theoretically, actual usage may be jankier) only detects changes in the current component when Input() props change. hc-table has only one Input() prop and I don't know what it does. The documentation says to use `renderRows()` when the data changes, but that doesn't cover cases where the data stays the same but the table has, say, a footer row whose values are calculated, or interactive behavior that animates cells when certain conditions are met.

My solution thus far has been to use an ng-template outside the table to represent the contents of the cell, then use an ng-container with *ngTemplateOutlet to include it into the table. It's bothersome to have to do things this way, and it doesn't work when an entire column, for example, needs reactive features, since all the elements of hc-table are so interdependent.

I could make the argument that it's a mistake to use OnPush change detection in a component that has content projection. We won't always (ever) be updating the component's content via props, and users shouldn't expect that they have to. They'll expect to control the content they're projecting via their own component. Maybe this is a gotcha in Angular itself, but still.

The counterargument to this is that some tables may have a lot of content (e.g. thousands of cells), and constant dirty-checking could reduce performance to an unacceptable level. If we want to prioritize this use case, better documentation is needed to indicate that `renderRows()` is not only necessary when the data source's rows change, but _any time_ the table is expected to update in any way. Also, we'd need to fix `renderRows()`, since right now it doesn't seem to do much.


isaaclyman:
NOTE: if we go the latter route, I suspect the fix to renderRows involves creating an override method that invokes the _changeDetectorRef, then calls the inherited method, like so:

image

isaaclyman:
Dang. Looks like the change detection issues go all the way down to the CDK. Still digging on this one...

isaaclyman:
I've done everything I can think of and my conclusion is that _Angular CDK does not support interactive, animated or calculated elements in tables_. They work to an extent, but become unresponsive far too often to be acceptable in production code.

Things I attempted:
- Removing `ChangeDetectionStrategy.OnPush`, as in this PR. Did not work.
- Using `changeDetectorRef.detectChanges()` or `.markForCheck()` in the `renderRows()` method, then calling that method from our app every time we needed an update. Did not work.
- Using `CdkTable.removeRowDef(null)` to trigger an update (something similar showed up in an old GitHub thread). Did not work, although `removeFooterRowDef(null)` did seem to trigger an update to the table footer.
- Setting `HcTableDataSource.data` to a new object reference (with the same data). Did not work.
- Using JavaScript to manually trigger a click on a DOM element when it needs to update. This works if the element has a click listener and you can reasonably get a reference to it. This turned out to be prohibitively difficult in my case--it would have worked if I'd spent a few more days on it, but it's a terrible hack, especially when multiple cells may need to update.

The only effective workaround I've found is the one I mentioned before, using an external ng-template for anything you want to control: https://github.com/angular/components/issues/13030#issuecomment-419480494

PR #916, opened 2019-07-10
style(drawer): ADA color contrast

### LINKED ISSUE #913: Updating Drawer Colors for ADA Standards Compliance
The current colors on the example drawer given in Cashmere are difficult to read and not in accordance with ADA Compliance standards.

"WCAG 2 level AA requires a contrast ratio of at least 4.5:1 for normal text 
3:1 for large text, and a contrast ratio of at least 
3:1 for graphics and user interface components (such as form input borders). 
Level AAA requires a contrast ratio of at least 7:1 for normal text and 4.5:1 for large text."

Screen Shot 2019-07-09 at 1 25 15 PM

The header bar would be #303A46
the body background would be #384655
This color combination provides ADA Compliance

andrew-frueh:
adjusts the suggested filter drawer colors to be ADA compliant

closes #913

### DIFF
--- a/projects/cashmere-examples/src/lib/drawer-basic/drawer-basic-example.component.scss
+++ b/projects/cashmere-examples/src/lib/drawer-basic/drawer-basic-example.component.scss
@@ -1,4 +1,4 @@
 .hc-drawer {
-    background-color: #6e6f71;
-    color: #b9babb;
+    background-color: #384655;
+    color: #c0c5cc;
 }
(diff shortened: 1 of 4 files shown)


projects/cashmere-examples/src/lib/drawer-basic/drawer-basic-example.component.scss
```diff
@@ -1,4 +1,4 @@
 .hc-drawer {
-    background-color: #6e6f71;
-    color: #b9babb;
+    background-color: #384655;
```
corykon: any reason not to use color variables here instead of a hardcoded hex?

andrew-frueh: Yeah - those are in examples, and it seemed like we weren't pulling sass variables into any other examples.  I'm not even sure what the path would be to do that.  Any idea?

corykon: Should be able to add this to the top of the file I think:
```
@import "~@healthcatalyst/cashmere/scss/colors";
```

Issue #914, opened 2019-07-09
Add an option for aligning split-button menus to the left


isaaclyman:
**Describe your idea**

Currently there are three options for the `menuPosition` attribute of an `<hc-split-button>`.

`start`:

![image](https://user-images.githubusercontent.com/9139369/60923220-a70ecf00-a25b-11e9-86a4-bfb3f23715ae.png)

`center`:

![image](https://user-images.githubusercontent.com/9139369/60923280-d32a5000-a25b-11e9-9c0d-4c8f48411fd2.png)

`end`:

![image](https://user-images.githubusercontent.com/9139369/60923108-72027c80-a25b-11e9-81c8-288471d00295.png)

If no menu option is chosen, the default seems to be `end`, which looks okay if there's enough room for the menu on the left side (see Cashmere examples under "Button Split"). But in this case, where the menu is wide enough that it would go off the left side of the screen, it flips to the other side and looks disconnected.

I think there should be an option that looks like this:

![image](https://user-images.githubusercontent.com/9139369/60923373-02d95800-a25c-11e9-8e5d-9ea26549c7e8.png)

PR #919, opened 2019-07-10
fix(split-button): change dropdown anchoring

corykon:
anchor the dropdown to the entire split button, not just arrow btn trigger portion

re #914


andrew-frueh:
@corykon - I pushed a couple small changes to the branch to use the default 'end' alignment on most of the examples except one that a put the 'start' alignment on and called it out in the dropdown.

Issue #917, opened 2019-07-10
Loading ring slides around on Cashmere docs site (Chrome only)


isaaclyman:
**Summary**

On Chrome, the Cashmere loading circle slides around while the docs site is loading, much like a hula hoop around the triflame. It should probably remain in the same place and just rotate the white bar.

See gif:

![cashmere-loading](https://user-images.githubusercontent.com/9139369/60988972-0c1d0000-a302-11e9-9f34-4d4aafb837d8.gif)

**Reproduction**

Steps to reproduce:

1. Visit https://cashmere.healthcatalyst.net on Windows 10/Chrome

NOTE: problem does not occur on Firefox, nor does it occur on the Progress Indicator example on any browser.

corykon:
'much like a hula hoop' haha

PR #921, opened 2019-07-10
fix(loading animation): get rid of hula hoop effect

corykon:
re #917

We'll want to communicate to teams that if they copied the old loading screen styles, they'll want to make note of this change.

Not sure if its worth the effort... but I wonder if there should be styles page that describes how to implement this. It's a bit trickier than other components because it needs to show before angular has bootstrap (so it can't be a regular component).


andrew-frueh:
A style page for this would probably be really good - since there are multiple apps using it now.  @corykon - can you add an issue for that?

PR #928, opened 2019-07-11
fix(form-control): remove component decorator

### LINKED ISSUE #927: Angular 8 + Ivy error: "Directives cannot inherit Components"
**Summary**

Running an Angular 8 + Ivy app that uses Cashmere Input causes a module load error:

```
Uncaught Error: Directives cannot inherit Components
    at ɵɵInheritDefinitionFeature (core.js:22125)
    at core.js:4347
    at Array.forEach (<anonymous>)
    at Object.toString (core.js:4343)
    at noSideEffects (core.js:4202)
    at Module.ɵɵdefineComponent (core.js:4330)
    at Module../node_modules/@healthcatalyst/cashmere/__ivy_ngcc__/fesm2015/healthcatalyst-cashmere.js (healthcatalyst-cashmere.js:9579)
```

Looking at the source, I found the issue:

in `projects/cashmere/src/lib/input/input.directive.ts`:
```ts
/** Directive that allows a native input to work inside a HcFormFieldComponent */
@Directive({
    selector: '[hcInput]',
    providers: [{provide: HcFormControlComponent, useExisting: forwardRef(() => InputDirective)}]
})
export class InputDirective extends HcFormControlComponent implements DoCheck {
```

in `projects/cashmere/src/lib/form-field/hc-form-control.component.ts`:
```ts
@Component({
    selector: 'hc-form-control',
    template: '<ng-content></ng-content>',
})
export class HcFormControlComponent {
```

This is the issue - a directive cannot extend a component.  But I am not sure that this will be too hard to fix.  We don't ever use `hc-form-control` on its own, right?  I mean, no one does `<hc-form-control>...</hc-form-control>`; instead we use the input directive or some other component that inherits from `HcFormControlComponent`.  So I don't think it's necessary to actually decorate this class with `@Component(...)` at all.  In fact, we could probably make it into an `abstract` class to make it impossible to use without extending it.  If we remove the `@Component` decoration from this class, it should fix the error.

joeskeen:
also make it an abstract class

Closes #927

### DIFF
--- a/projects/cashmere/src/lib/form-field/hc-form-control.component.ts
+++ b/projects/cashmere/src/lib/form-field/hc-form-control.component.ts
@@ -1,12 +1,5 @@
-import {Component, ViewEncapsulation} from '@angular/core';
-
 /** Wrapper component that textual form controls extend to work with hc-form-field */
-@Component({
-    selector: 'hc-form-control',
-    template: '<ng-content></ng-content>',
-    encapsulation: ViewEncapsulation.None
-})
-export class HcFormControlComponent {
+export abstract class HcFormControlComponent {
     /** Whether the control should be displaying an associated error */
     _errorState: boolean = false;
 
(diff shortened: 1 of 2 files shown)


joeskeen:
I tested all the form controls with these changes and they still work.

PR #968, opened 2019-09-02
update README.md - Running the demo application

JoshK2:
Add the missing run command before the start.


corykon:
Good tweak here, I've definitely tripped over missing `npm run build`.

Looks like the build is failing because of the formatting of the git commit message. We've got some tight rules around how those are supposed to look. It talks about that in the contribution guidelines, which are somewhat buried: https://cashmere.healthcatalyst.net/guides/contribution-guide.

Easiest way to get the commit right is to run `npm run cm` instead of `git commit`.

JoshK2:
So you want that I create a new pr with the `npm run cm` command?

PR #971, opened 2019-09-06
943 toast items

### LINKED ISSUE #943: Toast Styles Questions/Suggestions
**Describe your idea**
Couple questions/suggestions. While implementing the cashmere toasts, there were two things we found ourselves wanting to change.

1. Long messages make for really wide, awkward to read toasts. Feels like this should have a max width somewhere from 300-400px and then the text should wrap.
image

2. The message text font size seems small. We're sitting at about 12px by default, which is "extra small" on our typography guidelines. I'd love to have that be at least 14px ("small") by default.

JDavidStevens:
Added the option to include a progress bar. 
Added a width configuration and set the default toast width to 300px.
Increased toast body font size.

closes #943

### DIFF
--- a/projects/cashmere-examples/src/lib/toaster-overview/toaster-overview-example.component.html
+++ b/projects/cashmere-examples/src/lib/toaster-overview/toaster-overview-example.component.html
@@ -28,11 +28,6 @@
         </hc-select>
     </hc-form-field>
 
-    <hc-form-field class="example-field">
-        <hc-label>Toast Width:</hc-label>
-        <input hcInput type="number" [(ngModel)]="toastWidth" />
-    </hc-form-field>
-
 </div>
 <div class="example-column">
     <hc-form-field class="example-field">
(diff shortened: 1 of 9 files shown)


andrew-frueh:
Great work and thanks for submitting this @JDavidStevens.  @corykon @joeskeen and I finally reviewed it yesterday (HAS slowed everything down) and we realized that in addition the indeterminate progress bar you added, it would be really helpful to give users the option for a determinate progress bar.  For example, if they wanted the progress bar on the toast to indicate how long before the toast closes.

So I made that addition and push the change up to your branch.  Joe and Cory - please take a look at how I implemented this.  I actually kept Cory's boolean value as it was, because I realized that the place to set the progress value is on the toastRef, not the options (where the boolean is).  Because a user would want to be able to update the progress value multiple times after the toast is open, not just once when it is created.  So I added a `progress` value to the toastRef, similar to how progress Indicator works.

Let me know if you have any suggestions.

andrew-frueh:
That's a great suggestion @corykon - that occurred to me too that not everyone would understand that you can hold onto that ref.  Maybe we had an example called using `HcToastRef`s or something like that.  Probably something for another PR.

PR #979, opened 2019-09-17
User/philip.heebner/rework tabs

pheebner:
Using the actual TabComponent template instead of generating anchor elements with a for loop that copies attributes/data from the TabComponent children of a tab set. This allows us to treat the TabComponents more like actual html elements (put custom attributes on them, etc.) since their templates are actually rendered into the view. It also allows us to render additional content inside the tab bar after the TabComponents if the user desires (maybe a button, etc.).

Doing things this way breaks the old way of rendering the tab content and it is now done in a different way. The TabComponent template now includes an ng-template which contains the TabComponent's content. The parent TabSetComponent then uses an ng-container with a ngTemplateOutlet directive to reference the content of whatever tab is selected.

I created a TabsService to communicate 'direction' from the TabSetComponent to its children on initialization.

The parent TabSetComponent now subscribes to its children's tab clicks.

It's possible that if anyone has overridden the hc-tab-width css class it would cause a breaking change. I've taken the class out and instead applied a :host styling to the TabComponent.

Let me know if anything here will not work for whatever reason.

NOTE: I moved some files around to organize some of the new content. If you want to see only the changes to the files before they were moved, look at the first commit of this PR.

Changes to examples to demonstrate additional functionality:

Adding attributes to hc-tab components (hcPop for example)
image

Additional content in tab bar
image


joeskeen:
I don't see any changes to the tabs examples.  Would this be a non-breaking change for consumers of the library?

pheebner:
> I don't see any changes to the tabs examples. Would this be a non-breaking change for consumers of the library?

This should be a non-breaking change as far as I can see, unless they have overridden the css class 'hc-tab-width'. I took this class out since I was able to apply the same styling using :host for the TabComponent.

pheebner:
@joeskeen Do you think it would be a good idea to add a tab example? Maybe something that shows extra content in the tab bar?

projects/cashmere/src/lib/tabs/tab-set/tab-set.component.ts
```diff
@@ -70,7 +82,11 @@ export class TabSetComponent implements AfterContentInit {
         }
     }
 
-    constructor(private router: Router, private route: ActivatedRoute) {}
+    constructor(private router: Router, private route: ActivatedRoute, @Self() private _tabsService: TabsService) {}
```
mhollier: What is the reasoning for introducing a service here?

pheebner: I wanted to be able to tell the TabComponent children what their parent's direction was without having to initialize the content of the TabSetComponent first. This way the TabComponent's direction would be correct before it ever initializes its view or content. If there's a better way to do this I'm all ears, I just read a service was good practice for parent-child communication.

joeskeen: Services are Singleton in Angular; what if you have multiple tab sets, or nested tabs?

joeskeen:
> @joeskeen Do you think it would be a good idea to add a tab example? Maybe something that shows extra content in the tab bar?

Absolutely. Otherwise people won't know they can.

projects/cashmere/src/lib/tabs/tab-set/tab-set.component.ts
pheebner: I believe it should create a new TabsService for every TabSetComponent since the TabSetComponent lists "TabsService" under "providers" in the @Component decorator. The @Self decorator in the constructor should also prevent TabSetComponent from getting the dependency from anywhere other than its own injector. Each TabComponent should get the dependency from their parents because they don't provide it themselves and the next place to look would be their parent component.

projects/cashmere-examples/src/lib/tabs-vertical/tabs-vertical-example.component.html
```diff
@@ -17,13 +17,22 @@
...
+                        <span>You can place additional content after the tabs</span>
+                    </div>
+                </div>
             </hc-tab-set>
         </hc-tab>
-        <hc-tab tabTitle="Tab 2">
+        <hc-tab tabTitle="Tooltip Tab" [hcPop]="tabTooltip" trigger="hover">
             <div class="tab-content">
-                Vertical Tab content 2
+                This tab shows a Cashmere popover tooltip when you hover over it.
```
mhollier: :joy:

mhollier: Can you provide a screenshot or two of the demo app?

pheebner: Ah right, good idea, one sec..

pheebner: image

image

mhollier: Yay thanks

pheebner:
@joeskeen I added some examples of the new functionality (see above in the description)

PR #987, opened 2019-09-21
fix(form-field): support for no labels

### LINKED ISSUE #986: Form Field Support for No Labels
**Summary**

@snicol21 noted that with an `hc-form-field`, if no label is included, the label wrapper is still added.  So there is a lot of unnecessary padding - both for normal and inline modes.  See screenshots below:

image (8)

image (7)

andrew-frueh:
Removes label wrappers if no labels are present.

closes #986

### DIFF
--- a/projects/cashmere/src/lib/form-field/hc-form-field.component.html
+++ b/projects/cashmere/src/lib/form-field/hc-form-field.component.html
@@ -1,5 +1,5 @@
-<div [ngClass]="inline ? 'hc-form-field-wrapper-inline' : 'hc-form-field-wrapper'">
-    <span [ngClass]="inline ? 'hc-form-field-label-wrapper-inline' : 'hc-form-field-label-wrapper'">
+<div [ngClass]="{'hc-form-field-wrapper-inline': inline, 'hc-form-field-wrapper': !inline, 'hc-form-field-no-label': !_labelChildren.length}">
+    <span *ngIf="_labelChildren.length" [ngClass]="inline ? 'hc-form-field-label-wrapper-inline' : 'hc-form-field-label-wrapper'">
         <label [ngClass]="inline ? 'hc-form-field-label-inline' : 'hc-form-field-label'" [attr.for]="_control._componentId">
             <ng-content select="hc-label"></ng-content>
             <span class="hc-required-marker" aria-hidden="true" *ngIf="(_control._isRequired )&& !_control._isDisabled">&nbsp;*</span>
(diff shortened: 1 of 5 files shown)


projects/cashmere/src/lib/sass/form-field.scss
```diff
@@ -138,6 +138,10 @@ $wrapper-padding-bottom: ($error-margin-top + $line-height) * $error-font-scale;
     vertical-align: baseline;
 }
 
+@mixin hc-form-field-no-label() {
+    border-top: none !important;
```
corykon: I can't see a reason why anybody would want to override here and put their own border on this particular DOM element, but I still  have to ask - any chance we can rearrange things or make selectors more specific in some way to avoid the use of !important?

projects/cashmere/src/lib/form-field/hc-form-field.component.html
```diff
@@ -1,5 +1,5 @@
-<div [ngClass]="inline ? 'hc-form-field-wrapper-inline' : 'hc-form-field-wrapper'">
-    <span [ngClass]="inline ? 'hc-form-field-label-wrapper-inline' : 'hc-form-field-label-wrapper'">
+<div [ngClass]="{'hc-form-field-wrapper-inline': inline, 'hc-form-field-wrapper': !inline, 'hc-form-field-no-label': !_labelChildren.length}">
```
joeskeen: IMHO this line is really hard to read and reason with.  Perhaps if you broke it up a bit:
```html
<div [class.hc-form-field-wrapper-inline]="inline"
     [class.hc-form-field-wrapper]="!inline"
     [class.hc-form-field-no-label]="!_labelChildren.length">
```

projects/cashmere/src/lib/form-field/hc-form-field.component.html
```diff
@@ -1,5 +1,5 @@
...
+    <span *ngIf="_labelChildren.length" [ngClass]="inline ? 'hc-form-field-label-wrapper-inline' : 'hc-form-field-label-wrapper'">
```
joeskeen: Also, `_labelChildren.length` is seen multiple times, maybe just change it to a property and use that?
```ts
get hasLabel() {
  return !!_labelChildren.length;
}
```

projects/cashmere/src/lib/sass/form-field.scss
andrew-frueh: Cool - looks like I don't need `important` to override the base style (thought I did).  Removing it.

projects/cashmere/src/lib/form-field/hc-form-field.component.html
andrew-frueh: There are actually two other spots in the template that do something similar, so I'm going to change all three as you indicated.

PR #988, opened 2019-09-21
fix(datepicker): improved header labeling

### LINKED ISSUE #703: UX: Date Picker
Funcationality for the arrows around the word today is unclear to me. I felt like the clicking them should move my selection by one day, but they move forward and backward by month. Might make more sense to wrap those arrows around the month instead of the word today.

image

### LINKED ISSUE #985: Daterange component is difficult to use
Based on usability testing at HAS, we learned that the daterange component is difficult to use.  
Screen Shot 2019-09-20 at 4 22 07 PM

While there was a high completion rate, people rated it as difficult to use and watching them struggle changing a date range confirms what we saw.

Based on what we saw, I think what can solve the issue is just changing some labels!

Where it says for example "Sep 2019", just change the label to "2019 (the current year)" since when you click on that you see years and the user can change the year accordingly.

Where it say "Today", replace that with the month that the user is on since the left right arrows are used to navigate the months in that year.

andrew-frueh:
Adjusted header elements to address UX concerns that Alejo uncovered in his usability testing.

Closes #985 
Closes #703

### DIFF
--- a/projects/cashmere-examples/src/lib/date-range/date-range-example.component.ts
+++ b/projects/cashmere-examples/src/lib/date-range/date-range-example.component.ts
@@ -15,9 +15,9 @@ export class DateRangeExampleComponent implements OnInit {
     ngOnInit() {
         const today = new Date();
         const fromMin = new Date(today.getFullYear(), today.getMonth() - 2, 1);
-        const fromMax = new Date(today.getFullYear(), today.getMonth() + 1, 0);
+        const fromMax = new Date(today.getFullYear(), today.getMonth() + 1, 1);
         const toMin = new Date(today.getFullYear(), today.getMonth() - 1, 1);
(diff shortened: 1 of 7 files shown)


joeskeen:
> Or maybe used the space at the bottom for a text-based link:

I second this.

andrew-frueh:
Good point @corykon and I agree that no one will think to click on the month to jump to the current today.  I like your idea of adding a link/button in the space below to jump to the current day.  I'll get that added.

joeskeen:
IMHO I think you could just go with 'Today'... not sure you need the 'jump to'

andrew-frueh:
I thought about that one Joe, and I think if it was styled as a button that might work.  But as link text, I think the extra instruction is useful.  Especially because we had the single word button before (in a less than ideal place) - and no one in the tests thought to click on it.  Also aesthetically it fills up that bottom space a little better than a single word.

PR #990, opened 2019-09-21
fix(modal): restrict multiple modals

### LINKED ISSUE #959: Modal should not open more than one instance of itself at a time
**Summary**

You can open the same modal as many times as you want and they will just continually stack on top of each other. The modal should either prevent the same one from being opened multiple times or it needs to expose a function that allows a consumer to check and see if it is already open.

**Reproduction**

Steps to reproduce:
1. On the example page click on the either of the buttons to open the modal.
2. The button will still have the focus so if you hit the enter key a few times now you will open a few more of the same modal. The overlay gradually gets darker and you have more modals to close now.

**Additional context**

image

image

andrew-frueh:
Adds a flag to ModalService to restrict multiple modals being opened.

closes #959

### DIFF
--- a/projects/cashmere/src/lib/modal/modal.service.ts
+++ b/projects/cashmere/src/lib/modal/modal.service.ts
@@ -20,7 +20,10 @@ export type ModalContentType = Type<{}> | TemplateRef<any>;
 @Injectable()
 export class ModalService {
 
-    /** Restricts multiple modals from being opened on top of each other (an error will be thrown if attempted). */
+    /** Defaults to false. Restricts multiple modals from being opened on top of each other
+     * (an error will be thrown if attempted). It's generally considered bad practice to open multiple
+     * models at once, so only change this with good reason.
+    */
     allowMultiple: boolean = false;
 
     // start at 2000 (reserved range for modals, see _variables.scss)
--- a/projects/cashmere/src/lib/modal/modal.service.ts
+++ b/projects/cashmere/src/lib/modal/modal.service.ts
@@ -19,9 +19,14 @@ export type ModalContentType = Type<{}> | TemplateRef<any>;
 
 @Injectable()
 export class ModalService {
+
+    /** Restricts multiple modals from being opened on top of each other (an error will be thrown if attempted). */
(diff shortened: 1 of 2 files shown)


projects/cashmere/src/lib/modal/modal.service.ts
```diff
@@ -19,9 +19,14 @@ export type ModalContentType = Type<{}> | TemplateRef<any>;
 
 @Injectable()
 export class ModalService {
+
+    /** Restricts multiple modals from being opened on top of each other (an error will be thrown if attempted). */
+    allowMultiple: boolean = false;
```
joeskeen: This is a BREAKING CHANGE unless you default it to `true`.

joeskeen: As the commit is not flagged as breaking change, I would set this to `true` as default.

andrew-frueh:
This is kind of gray area to me @joeskeen.   As @corykon and I were talking in our UX meeting, we were thinking this is more of a bug than a breaking change.  While it's true that if an app was currently stacking modals, this would prevent it from working - that's also not behavior that we really want our apps to be using.  So I hesitate to make this true by default.

If we need to follow the letter of the law though on this, I can understand that.  And so would I need to go back and update the commit to include the words `BREAKING CHANGE`?

joeskeen:
@andrew-frueh yes if this is the default behavior we want to have, we should designate it as a BREAKING CHANGE. I know that at least one app has been using this 'bug' as a 'feature', so we should be kind to our consumers and give them a heads up.

joeskeen:
Another avenue we could explore... what if stacking the same modal on top of itself causes an error, but not if it's not the same component class?  It doesn't sound great, but it would make it possible to separate the the 'bug' from the 'feature'...

andrew-frueh:
@joeskeen - I updated the description text for this property explaining that it's set to false by default.  I also included `BREAKING CHANGE` in the commit message.  So hopefully this will trigger versioning appropriately.  We'll also be sure and announce this on our slack channels.

PR #993, opened 2019-09-22
fix(split-button): minimal split buttons

### LINKED ISSUE #890: Minimal styled split buttons have invisible drop-down arrow
**Summary**

When styling a split button with minimal, the drop-down is invisible.
Stackblitz reproduction: 
https://stackblitz.com/edit/angular-um8nzq

**Additional context**

This was used to select an icon, which was displayed in lieu of the button text.  I was able to work around the issue by tying a button (click) event to manually triggering the dropdown, but it feels like the dropdown should be visible if present.

andrew-frueh:
adds split-button support for the minimal button style

@corykon - what do you think about this styling for a minimal split button?

closes #890

### DIFF
--- a/projects/cashmere-examples/src/lib/button-split/button-split-example.component.html
+++ b/projects/cashmere-examples/src/lib/button-split/button-split-example.component.html
@@ -14,12 +14,17 @@
     </button>
 </hc-split-button>
 
-<hc-split-button buttonStyle="destructive" menuPosition="start">
+<hc-split-button buttonStyle="secondary" menuPosition="start">
     Simple Dropdown
     <div hcButtonItem>Dropdown with menuPosition set to 'start'</div>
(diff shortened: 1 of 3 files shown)


corykon:
Feeling like I'd need to see it in context. I tried a couple other approaches just cause, but I think yours is best.
image
2019-09-23 13_07_07-Cashmere
2019-09-23 13_12_35-Cashmere



Since we allow implementers to do it, obviously we ought make it work, but frankly I'm not sure I'd want "minimal" to be used with a split button. Feels a bit weird.

andrew-frueh:
Yup good thoughts on both of those - I thought about those two. But it felt like putting a border around either broke from the “minimal” style of the normal buttons.

So I’m with you, not sure there is a strong use case for this, but at least it’s supported now.

PR #996, opened 2019-09-23
fix(checkbox): indeterminate checkbox state

### LINKED ISSUE #953: Checkbox: Indeterminate Example
**Where does our documentation fall short?**
Would be great to have an example of the `indeterminate` property for the `hc-checkbox`. 

Note: I'm not sure if this property was ever fully implemented. I see that there's a `hc-checkbox-indeterminate` class applied, but I don't think there are any styles targeting that class. 

I imagine it would look something like this:
image

andrew-frueh:
adds support of the indeterminate property on the checkbox component and adds an example to illustrate it.

closes #953


projects/cashmere-examples/src/lib/checkbox-indeterminate/checkbox-indeterminate-example.component.ts
```diff
@@ -0,0 +1,45 @@
...
+    childChecks = [true, false, true, false];
+
+    isChecked() {
+        return this.childChecks[0] && this.childChecks[1] && this.childChecks[2] && this.childChecks[3];
+    }
+
+    isIndeterminate() {
+        let numChecked = 0;
+        for (let i = 0; i < this.childChecks.length; i++) {
+            if (this.childChecks[i]) {
```
joeskeen: This could probably be reduced to this:
```ts
const checkedCount = this.childChecks.filter(c => c).length;
return checkedCount === 0 || checkedCount === childChecks.length;
```

projects/cashmere-examples/src/lib/checkbox-indeterminate/checkbox-indeterminate-example.component.html
```diff
@@ -0,0 +1,10 @@
+<hc-checkbox [checked]="isChecked()" [indeterminate]="isIndeterminate()" (change)="parentClick()">
+    Parent Checkbox (indeterminate)
+</hc-checkbox>
+
+<div class="checkbox-group">
+    <hc-checkbox [checked]="childChecks[0]" (change)="childClick($event)" id="0">Child Checkbox</hc-checkbox>
```
joeskeen: Although this works, I think it would be better to have our example use Angular Reactive Forms instead. (as it's more or less a best practice.  In this example, the model is changed by the UI, but you can't initialize the form with previously saved values.  IMHO it makes this example less compelling.

projects/cashmere-examples/src/lib/checkbox-indeterminate/checkbox-indeterminate-example.component.ts
```diff
@@ -0,0 +1,45 @@
...
+        }
+    }
+
+    childClick(event: CheckboxChangeEvent) {
+        this.childChecks[parseInt(event.source.id, 10)] = event.checked;
+    }
+
+    parentClick() {
+        if (!(this.childChecks[0] && this.childChecks[1] && this.childChecks[2] && this.childChecks[3])) {
+            this.childChecks[0] = this.childChecks[1] = this.childChecks[2] = this.childChecks[3] = true;
```
joeskeen: This just feels a bit hard-coded to me. I'd be uncomfortable putting it out there as "the way you should use this component".

andrew-frueh:
@joeskeen - take a look at the update I've pushed.  It uses reactive forms as requested and no longer has anything hard coded in the typescript outside of the init.

PR #1004, opened 2019-09-25
524 convert radio to mixins

### LINKED ISSUE #524: (Opt-in) Convert the Radio Button component

isaaclyman:
Fixes #524.

I added the `disabled` class to a couple of internal elements. This is admittedly a little redundant, but it allows us to use gentler CSS selectors, which is a win for overridability.


projects/cashmere/src/lib/radio-button/radio-button.component.html
```diff
@@ -1,5 +1,7 @@
-<label class="hc-radio-container" [ngClass]="{'disabled': disabled}">
-    <input type="radio"
+<label class="hc-radio-container" [class.disabled]="disabled">
+    <input class="hc-radio-input"
+           [class.disabled]="disabled"
```
joeskeen: I'm curious, do we also need to set [disabled]="disabled" here?

isaaclyman: That would improve accessibility. Good call.

isaaclyman: Oh, looks like it's already set. See line 9.

joeskeen: So it is! :)

PR #1005, opened 2019-09-25
feat(age.pipe): AgePipe returns age in terms of mos or yrs

Techgeekster:


projects/cashmere/src/lib/pipes/age-pipe/age.pipe.spec.ts
```diff
@@ -0,0 +1,105 @@
+
+import {async, inject, TestBed} from '@angular/core/testing';
+
+import { AgePipe } from './age.pipe';
+import { Date } from 'sugar';
+
+describe('AgePipe', () => {
+    beforeEach(async(() => {
+        TestBed.configureTestingModule({
```
joeskeen: I would prefer to not use TestBed for testing Pipes, unless it is necessary to test integration with other Angular things.  You should just be able to `new()` up the pipe and test it.

projects/cashmere/src/lib/pipes/age-pipe/age.pipe.ts
```diff
@@ -0,0 +1,22 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { Date } from 'sugar';
```
joeskeen: This would require a peerDependency to be added to Cashmere (currently the DatePicker doesn't strictly require Sugar). I'm working through how to open things up so you can depend on other libraries without requiring additional peer dependencies to the Cashmere library itself.

projects/cashmere/src/lib/pipes/age-pipe/age.pipe.ts
```diff
@@ -0,0 +1,22 @@
...
+})
+export class AgePipe implements PipeTransform {
+
+  public transform(value: string): string {
+    const birth: Date = Date.create(value);
+
+    // if under 12 months then show in months
+    let ageInMonths: number = Date.monthsAgo(birth);
+    if (ageInMonths < 12) {
+      return `${ageInMonths} ` + ((ageInMonths === 0 || ageInMonths > 1) ? 'mos' : 'mo');
```
joeskeen: We want Cashmere to work with localization out of the box.  How would you handle alternative locales?

projects/cashmere/src/lib/pipes/age-pipe/age.pipe.ts
joeskeen: Also, I think we would prefer to not shadow the global `Date`; IMHO it would be better to either:

* create an alias: `import { Date as SugarDate }` from 'sugar';
* import the functions directly: `import createDate from 'sugar/date/internal/createDate';`

projects/cashmere/src/lib/pipes/age-pipe/age.pipe.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+        // arrange
+        const testDate: string = adjustDate(-49, 0, 1).toISOString();
+
+        // act
+        const result: string = pipe.transform(testDate);
+
+        // assert
+        expect(result).toEqual('48 yrs');
+    }));
+});
```
joeskeen: A few edge cases I would include:
* `null`/`undefined`
* some object / other unexpected value

joeskeen:
Even though we don't have a place for pipe examples to live yet, I think documentation and examples should be created before this is merged in.

projects/cashmere/src/lib/pipes/age-pipe/age.pipe.ts
Techgeekster: I honestly have no clue as I've never really done localization before.... what would you suggest?

projects/cashmere/src/lib/pipes/age-pipe/age.pipe.spec.ts
Techgeekster: Fixed

projects/cashmere/src/lib/pipes/age-pipe/age.pipe.ts
Techgeekster: Fixed

projects/cashmere/src/lib/pipes/age-pipe/age.pipe.spec.ts
Techgeekster: Added tests

projects/cashmere/src/lib/pipes/age-pipe/age.pipe.ts
joeskeen: Take a look at how Angular handles it in the DatePipe: https://github.com/angular/angular/blob/fa699f65d7a2cb4d27b508e816acf256d03ec259/packages/common/src/pipes/date_pipe.ts#L156

projects/cashmere/src/lib/pipes/age-pipe/age.pipe.ts
```diff
@@ -0,0 +1,36 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { Date as SugarDate } from 'sugar';
+import chalk from 'chalk';
+
+@Pipe({
+  'name': 'age'
```
joeskeen: This should be a `pure: false` pipe, right?

corykon:
@andrew-frueh @isaaclyman @joeskeen We may want to abandon this PR for now. I think it might be best that our pipes not require peer dependencies (this one requires sugarJS) Unless there are additional teams specifically asking for a pipe like this.

andrew-frueh:
I believe that concern was raised earlier as well @corykon and I think the consensus was to avoid a SugarJS dependency if at all possible.  So I'll go ahead and close this and we can revisit in the future if the need comes up again.

PR #1007, opened 2019-09-25
feat(null-or-empty.pipe): returns alternative string on null or empty

Techgeekster:


projects/cashmere/src/lib/pipes/null-or-empty/null-or-empty.pipe.ts
```diff
@@ -0,0 +1,15 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+    'name': 'nullOrEmpty'
+})
+export class NullOrEmptyPipe implements PipeTransform {
+
+    public transform(value: string, altText: string): string {
+        if (value) {
+            value = value.trim();
```
joeskeen: What if an object is passed? What if the value is falsey like `0` or `false` or `NaN` but not `null` or `''`?

projects/cashmere/src/lib/pipes/null-or-empty/null-or-empty.pipe.spec.ts
```diff
@@ -0,0 +1,41 @@
+import {async, inject, TestBed} from '@angular/core/testing';
+
+import { NullOrEmptyPipe } from './null-or-empty.pipe';
+
+describe('NullOrEmptyPipe', () => {
+    beforeEach(async(() => {
+        TestBed.configureTestingModule({
```
joeskeen: I'd like to avoid using `TestBed` where it's not strictly necessary.

projects/cashmere/src/lib/pipes/null-or-empty/null-or-empty.pipe.ts
```diff
@@ -0,0 +1,15 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+    'name': 'nullOrEmpty'
+})
+export class NullOrEmptyPipe implements PipeTransform {
+
```
joeskeen: Should add JSDoc comments to have documentation generated from them.

projects/cashmere/src/lib/pipes/pipes.module.ts
```diff
@@ -1,9 +1,10 @@
 import {NgModule} from '@angular/core';
 
 import {EllipsisPipe} from './ellipsis.pipe';
+import { NullOrEmptyPipe } from './null-or-empty/null-or-empty.pipe';
 
 @NgModule({
-    declarations: [EllipsisPipe],
-    exports: [EllipsisPipe]
+    declarations: [EllipsisPipe, NullOrEmptyPipe],
```
joeskeen: The more I think about it, the more I am starting to believe that each pipe should be able to be imported independently.  Perhaps we could have a `NullOrEmptyPipeModule` and an `EllipsisPipeModule` then have a `PipesModule` that `exports` all of them for convenience?

projects/cashmere/src/lib/pipes/null-or-empty/null-or-empty.pipe.ts
joeskeen: I would also add test cases for all these cases.

joeskeen: perhaps you could do something like this:

```ts
if (value) {
  value = value.toString().trim();
}
```

joeskeen:
Even though we don't have a place for pipe examples to live yet, I think documentation and examples should be created before this is merged in.

projects/cashmere/src/lib/pipes/null-or-empty/null-or-empty.pipe.ts
```diff
@@ -0,0 +1,15 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+    'name': 'nullOrEmpty'
```
joeskeen: This should be a `pure: true` pipe, right?

corykon:
Added docs, resolved issues from previous comments. Renamed the pipe for clarity, as it seems to have only ever been for use on strings. `ifEmptyString` also reads nicely in the markup:
```
{{ someVar | ifEmptyString: 'alternate value' }}
```

corykon:
Believe I've got the pipe in good working order now. I think the original intention was to have it just be used for strings, so I've tried to make that clearer.

PR #1009, opened 2019-09-26
Additional logic for matching currentRoute in defaultToFirstRoute method

iamboo:
This has been tested locally, but I don't know what the build test process is.  Just trying to help fix a bug we found on the tabs component.  Thanks.


joeskeen:
Cashmere commits must conform to the conventional commit format (so that semantic versioning is objective and correct). You'll need to change your commit message to match that format.  Usually when working on Cashmere, you can just stage your files then run `npm run cm`. That will walk you through the process of creating the commit message in the proper format.

joeskeen:
Hey @iamboo do you need help?

PR #1014, opened 2019-10-01
User/philip.heebner/rework tabs

pheebner:
Allow user to access TabsService if they need to

### DIFF
--- a/projects/cashmere/src/lib/tabs/index.ts
+++ b/projects/cashmere/src/lib/tabs/index.ts
@@ -2,3 +2,4 @@ export {TabComponent} from './tab/tab.component';
 export {TabSetComponent, TabChangeEvent} from './tab-set/tab-set.component';
 export {HcTabTitleComponent} from './tab/tab-title.component';
 export {TabsModule} from './tabs.module';
+export {TabsService} from './tabs.service';


pheebner:
@andrew-frueh do you think we could do a patch release sometime soon? We need this for something we're working on in Ops Console using the tabs.

andrew-frueh:
Sure thing.  @joeskeen - can you queue up the release today or would you like me to take care of that?

projects/cashmere/src/lib/tabs/index.ts
```diff
@@ -2,3 +2,4 @@ export {TabComponent} from './tab/tab.component';
 export {TabSetComponent, TabChangeEvent} from './tab-set/tab-set.component';
 export {HcTabTitleComponent} from './tab/tab-title.component';
 export {TabsModule} from './tabs.module';
+export {TabsService} from './tabs.service';
```
joeskeen: I'm curious what this would be used for. As you said, this TabsService isn't a singleton, so why would anyone want to import it and use it in their code as it's not going to be the same instance?

pheebner: The reason I was adding this is because I am extending the TabSetComponent in Operations Console to replace the template and styling, etc. I couldn't extend it because I couldn't provide the TabsService in the child class because it's not exported. This is kind of annoying and I've been getting a bit of a code smell from this. I think you might be right, this service is probably not a good idea.

I've been thinking about this a bit actually today, maybe it'd be better to just inject the TabSetComponent into the TabComponent's constructor? I remember trying this before and ran into circular dependency issues. I didn't understand why it was happening at the time because I didn't understand barrel index files. I've been researching this a bit during open space though. This article seems promising: https://medium.com/visual-development/how-to-fix-nasty-circular-dependency-issues-once-and-for-all-in-javascript-typescript-a04c987cf0de

I'm hoping I can solve the barrel index file circular dependency issue. What do you think?

joeskeen: I agree that if we can eliminate the need for this service, it will be better.  Why don't you try that out and if you figure out how to do that, we can close this PR in favor of that solution.

pheebner: Hey @joeskeen it seems that the circular dependency was in fact an issue with Angular, not the barrel files. I can use a forwardRef() to make it work though. It will build and run just fine, but unfortunately I still get a "circular dependency detected" warning upon building.

```
export class TabSetComponent implements OnInit, AfterContentInit {
    ...

    @ContentChildren(forwardRef(() => TabComponent))
    _tabs: QueryList<TabComponent>;

    ...
}
```
```
export class TabComponent implements OnInit, AfterContentInit {
    ...
    get direction(): string {
        return this._tabSetComponent.direction;
    }

    constructor(@Inject(forwardRef(() => TabSetComponent)) private _tabSetComponent: TabSetComponent) {}
}

```

Another option I suppose would be to just use setTimeout() to set each tab's direction after the initial content init, but this causes some janky looking view changes.

Yet another option would be to use ng-deep to style the tabs using the tab-set-component's styling, although I know this isn't exactly best practice.

PR #1015, opened 2019-10-04
fix(pagination): initial value of pageSize

### LINKED ISSUE #776: (pagination) page size is set by component before value is able to be bound
**Describe the bug**
If `pageSize` is two-way bound to a value, the pagination component will revert that value to the default value of `20` before it has a chance to bind the initial value (especially if that value isn't available at `ngOnInit`, such as query string subscriptions).  The result is that no matter what you pick, the value will always go back to `20`.

andrew-frueh:
Stops initial value of pageSize from being overwritten to the default.  I starred at this one for hours thinking it had something to do with base class inheritance.  Turns out it just had to do with the two-way binding being broken;  the pageSize setter wasn't emitting on a change.

closes #776

### DIFF
--- a/projects/cashmere/src/lib/pagination/base-pagination.ts
+++ b/projects/cashmere/src/lib/pagination/base-pagination.ts
@@ -52,6 +52,7 @@ export class BasePaginationComponent extends Initializable implements OnInit {
     set pageSize(value: number) {
         this._pageSize = coerceNumberProperty(value);
         this._pageSizeUpdated();
+        this._emitPageEvent(this.pageNumber);
     }
     private _pageSize: number = BasePaginationComponent._DEFAULT_PAGE_SIZE;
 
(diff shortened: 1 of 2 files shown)


projects/cashmere/src/lib/pagination/base-pagination.ts
```diff
@@ -52,6 +52,7 @@ export class BasePaginationComponent extends Initializable implements OnInit {
     set pageSize(value: number) {
         this._pageSize = coerceNumberProperty(value);
         this._pageSizeUpdated();
+        this._emitPageEvent(this.pageNumber);
```
joeskeen: 🤦‍♂ I can't believe it was that simple!

joeskeen: @andrew-frueh looks like you broke some tests though... might want to look into that

andrew-frueh:
Thank goodness for those unit tests @joeskeen!  Turns out that because the two-way binding was broken, we never noticed that the page `EventEmitter` wasn't emitting on pageSize changes.  And `pageNumberChange` and `pageSizeChange` would emit on any page event regardless of whether their specific value changed or not.  So behavior of both of those have been corrected - which gets the unit tests to passing.  

Thanks for being so thorough when you wrote those!!

PR #1017, opened 2019-10-07
feat(tab): remove tabs service

pheebner:
Remove dependency on TabsService. TabSetComponent and TabComponent no longer need it. Instead TabComponents are not rendered until their direction is set by their TabSetComponent.

@joeskeen : I had to add the use of setTimeout when setting the direction. I looked in the console and found that Angular was throwing because its value changed before the change detection cycle completed.


pheebner:
Whoops, need to fix tests...

PR #1019, opened 2019-10-08
fix(toasts): closing multiple toasts

### LINKED ISSUE #1018: Adding toasts cause error
**Summary**

If you add more than one test at a time, then wait for them to disappear, then try it again, it causes an error (instead of launching a toast).  From then on, you can no longer add toasts.  You can reproduce this on the Cashmere Toaster demo page: https://cashmere.healthcatalyst.net/components/toaster/examples

https://healthcatalyst.slack.com/files/U8FU3T1AA/FNYDB0QJV/cashmere_and_3_more_pages_-_microsoft_edge_2019-10-07_14-48-33.mp4

andrew-frueh:
Ensures cleanup functions are called on close with multiple toasts.  The `takeUntil` pipe was preventing the cleanup functions from being called more than once when multiple toasts were open.  

@joeskeen can you verify there aren't any unintended consequences from removing that pipe?  I'm not entirely sure why it was there in the first place, it may have been an artifact of an earlier version.

closes #1018

### DIFF
--- a/projects/cashmere-examples/src/lib/toaster-overview/toaster-overview-example.component.scss
+++ b/projects/cashmere-examples/src/lib/toaster-overview/toaster-overview-example.component.scss
@@ -33,6 +33,7 @@
 
 .progress-row {
     display: flex;
+    align-items: end;
     width: 100%;
 }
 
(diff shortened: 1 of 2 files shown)


joeskeen:
I'm pretty sure that takeUntil was introduced to fix the IE bug... I'll take a look at it and make sure that this doesn't cause a regression.

andrew-frueh:
Interesting.  I tested on IE11 with BrowserStack and the fix looks like it's working.  Let me know if you see anything different.

andrew-frueh:
Thanks for testing that @corykon

PR #1028, opened 2019-10-14
fix(daterange): highlight preset on date change

### LINKED ISSUE #939: Can't pre-select a preset radio button on the date-range  component
When you  add radio buttons for preset ranges on the date range component it doesn't seem to be possible to pre-select one when you bring up the modal.

We are currently replacing our interval picker with the date range component and want to be able to indicate the default range for the dashboard when the dashboard is first displayed. And then as the user chooses different ranges via the buttons I think they would expect them to still be set on subsequent visits to the directive.

andrew-frueh:
check an approproate preset if the currently selected range matches it

closes #939

### DIFF
--- a/projects/cashmere/src/lib/date-range/picker-overlay/picker-overlay.component.ts
+++ b/projects/cashmere/src/lib/date-range/picker-overlay/picker-overlay.component.ts
@@ -7,6 +7,7 @@ import {D} from '../../datepicker/datetime/date-formats';
 import {CalendarWrapperComponent} from '../calendar-wrapper/calendar-wrapper.component';
 import {Observable} from 'rxjs';
 import {map} from 'rxjs/operators';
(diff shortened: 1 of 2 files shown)


projects/cashmere/src/lib/date-range/picker-overlay/picker-overlay.component.ts
```diff
@@ -76,6 +85,19 @@ export class PickerOverlayComponent implements OnInit, AfterViewInit {
         this._setValidity();
     }
 
+    _isRangePreset() {
```
joeskeen: @andrew-frueh it sure would be nice to have some unit tests around this... ;)

andrew-frueh:
Lol - you are so right, Joe. I’ll get one added :-)

andrew-frueh:
Unit test added @joeskeen - let me know how that looks.

PR #1032, opened 2019-10-16
fix(tabs): defaultTab accepts number

### LINKED ISSUE #1021: TabSetComponent.defaultTab should accept a number
**Summary**

The description for TabSetComponent.defaultTab says "Zero-based numerical value..." but the type is `string`, likely because it accepts `none` as a value. This is confusing; the type should probably be `string | number` to allow either kind of input, so we don't have to cast numbers to string for a numerical value.

andrew-frueh:
allow the defaultTab parameter to accept either a string or a number

closes #1021

### DIFF
--- a/projects/cashmere/src/lib/tabs/tab-set/tab-set.component.ts
+++ b/projects/cashmere/src/lib/tabs/tab-set/tab-set.component.ts
@@ -73,7 +73,7 @@ export class TabSetComponent implements AfterContentInit {
     }
 
     set defaultTab(tabValue: string | number) {
-        if (!isNaN(Number(tabValue)) || tabValue === 'none') {
+        if (!isNaN(+tabValue) || tabValue === 'none') {
             this._defaultTab = tabValue;
         } else {
             invalidDefaultTab(tabValue);
--- a/projects/cashmere/src/lib/tabs/tab-set/tab-set.component.ts
+++ b/projects/cashmere/src/lib/tabs/tab-set/tab-set.component.ts
@@ -26,7 +26,7 @@ export function tabComponentMissing(): Error {
     return new Error(`TabSet must contain at least one TabComponent. Make sure to add a hc-tab to the hc-tab-set element.`);
 }
 
-export function invalidDefaultTab(tabVal: string) {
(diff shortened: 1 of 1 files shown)


projects/cashmere/src/lib/tabs/tab-set/tab-set.component.ts
```diff
@@ -68,11 +68,11 @@ export class TabSetComponent implements AfterContentInit {
...
      * will be immediately selected. Defaults to 0 (the first tab). */
     @Input()
-    get defaultTab(): string {
+    get defaultTab(): string | number {
         return this._defaultTab;
     }
 
-    set defaultTab(tabValue: string) {
+    set defaultTab(tabValue: string | number) {
         if (Number(tabValue) !== NaN || tabValue === 'none') {
```
joeskeen: This is actually not correct, since `NaN !== NaN`.  Should use `isNaN()` instead

projects/cashmere/src/lib/tabs/tab-set/tab-set.component.ts
```diff
@@ -38,7 +38,7 @@ export function invalidDefaultTab(tabVal: string) {
 export class TabSetComponent implements AfterContentInit {
     _routerEnabled: boolean = false;
     private _direction: string = 'vertical';
-    private _defaultTab: string = '0';
+    private _defaultTab: string | number = '0';
```
joeskeen: Since we're accepting a number, shouldn't the default just be `0`? That way we can save < 1ms of parsing that every single time.

projects/cashmere/src/lib/tabs/tab-set/tab-set.component.ts
joeskeen:
```ts
if (!isNaN(+tabValue) || tabValue === 'none') {
```

andrew-frueh: Oh I just saw the second post - is there a difference between `isNaN(+tabValue)` and `isNaN(Number(tabValue))`

PR #1035, opened 2019-10-18
Right-click Trigger on HcPop

### LINKED ISSUE #975: Right click context menu
**Describe your idea**
Would be awesome to include a right click context menu like this one:
image

Found in user testing that people love these. Might be nice to encourage app across the suite to use this pattern as much as makes sense.

**Tell us what you wish Cashmere could do. The rest of this form is optional.**
Reuse our popover menu, set up a nice directive like `hc-pop` but for right click context menus.

Introduce a pattern that many users appreciate. Its a nice escape hatch when you're not sure where to go to accomplish a task.

### LINKED ISSUE #1029: `[autoCloseOnContentClick]="true"` won't work if there is a duplicate instance of `<hc-pop>`
**Summary**

`[autoCloseOnContentClick]="true"` won't work if there is a duplicate reference to the same `<hc-pop>`.
An example of this is having a two of `#helpMenu` in the navbar; one for the desktop mode, and one for the mobile collapsed menu (two total references). 

**Additional context**

In this screenshot, there is one `<hc-pop>`, but it's used in two places. `[autoCloseOnContentClick]="true"` won't work unless I remove one of the uses. 

image

In the expanded navbar:

`<hc-icon class="hc-navbar-icon" fontSet="fa" fontIcon="fa-question-circle-o" [hcPop]="helpMenu"></hc-icon>`

In the mobile collapsible menu:

```
<hc-navbar-mobile-menu>
...
   <hc-icon class="hc-navbar-icon mobile-options" fontSet="fa" fontIcon="fa-question-circle-o" [hcPop]="helpMenu"></hc-icon>
...
</hc-navbar-mobile-menu>
```

Will break `autoCloseOnContentClick` inside of this:
```
<hc-pop #helpMenu [autoCloseOnContentClick]="true">
...
</hc-pop>
```
Remove one of the instances, the flag starts working again.

andrew-frueh:
Adds a right-click trigger to HcPop, the ability to position the popover based on the triggering click, and a popover context parameter that let's the developer store info about the popover (@corykon pointed out similar functionality in [this component](https://www.npmjs.com/package/ngx-contextmenu)).  Also added a new example to demonstrate the right-click context menu functionality, and updated the existing overview example with the new options.

While I was in there, I also submitted a fix for the `autoCloseOnContentClick` bug recently reported.  I saw it coming up in my example.  The reason for the bug was the `_overlayRef` in the service was coming back as `undefined` when close() was called from the component.  My fix is to store a reference to the component's overlay in a reference in the component itself, and use that instead in the service's close function - and that seems to work.  But I look to the steering committee to let me know if there are unintended consequences of that approach.

closes #975 
closes #1029

### DIFF
--- a/projects/cashmere-examples/src/lib/popover-overview/popover-overview-example.component.html
+++ b/projects/cashmere-examples/src/lib/popover-overview/popover-overview-example.component.html
@@ -26,6 +26,7 @@
             <option value="center">Center</option>
             <option value="end">End</option>
             <option value="below">Below</option>
+            <option value="mouse">Mouse Position</option>
         </hc-select>
     </hc-form-field>
 
@@ -37,6 +38,7 @@
             <option value="center">Center</option>
             <option value="end">End</option>
             <option value="after">After</option>
+            <option value="mouse">Mouse Position</option>
         </hc-select>
     </hc-form-field>
 
@@ -56,6 +58,7 @@
             <option value="click">Click</option>
             <option value="mousedown">Mousedown</option>
             <option value="hover">Hover</option>
+            <option value="rightclick">Right-Click</option>
             <option value="none">None</option>
         </hc-select>
     </hc-form-field>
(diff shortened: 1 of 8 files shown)


andrew-frueh:
I wanted to do unit tests - but it looks like we any have _any_ for HcPop right now :-(  So that felt like a separate PR...

andrew-frueh:
Also noticed that we have an issue already logged for unit tests on the popover component - #619

PR #1041, opened 2019-10-29
fix(chips): improved examples

### LINKED ISSUE #1040: Chip with Action Icon Registers Separate Click Events
**Summary**

When a chip is displayed with `[action]=true` and the resulting "X" is clicked, only the icon disappears. However, if someone clicks the chip body, then it behaves as expected.

**Reproduction**

Steps to reproduce:

1. Open StackBlitz example for chips with action
2. Click "X" on chips to make only the icon disappear
3. Click body of the chip to make the chip disappear

**Additional context**

Cashmere v6.2.0
Seen in both Angular 6 and Angular 8

andrew-frueh:
When I looked into #1040, I realized our chip examples were pretty hard-coded.  Seemed like there was a better way to do those.  This PR is my attempt to improve them - but I suspect you all might have ideas to improve what I have here.  All suggestions welcome. 

closes #1040

### DIFF
--- a/projects/cashmere-examples/src/lib/chip-action/chip-action-example.component.html
+++ b/projects/cashmere-examples/src/lib/chip-action/chip-action-example.component.html
@@ -1,2 +1,3 @@
-<hc-chip action="true" (click)="hideChip($event);">Action One</hc-chip>
-<hc-chip color="red" action="true" (click)="hideChip($event);">Color Action</hc-chip>
+<hc-chip *ngFor="let chip of chipset" [color]="chip.color" action="true" [hidden]="chip.hidden" (click)="chip.hidden = true">
+    {{ chip.name }}
+</hc-chip>
--- a/projects/cashmere-examples/src/lib/chip-action/chip-action-example.component.html
+++ b/projects/cashmere-examples/src/lib/chip-action/chip-action-example.component.html
@@ -1,3 +1,2 @@
-<hc-chip *ngFor="let chip of chipset" [color]="chip.color" action="true" [hidden]="chip.hidden" (click)="chip.hidden = true">
-    {{ chip.name }}
-</hc-chip>
+<hc-chip *ngIf="chips[0]" (click)="chips[0] = false" hasCloseButton="true" class="chip-pointer">Click Anywhere to Close</hc-chip>
+<hc-chip *ngIf="chips[1]" (closeClick)="chips[1] = false" hasCloseButton="true" color="red">Click the Close Button</hc-chip>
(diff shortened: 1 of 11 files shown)


corykon:
This looks like good code that will fix the faulty logic reported in the bug, but I feel like the examples are technically a bit more complicated now. That might be ok since the basic chip example is still there.

About the chip component in general - I find it odd that `action = true` adds a close button, and that clicking on the "x" and clicking on the body of the chip fire the same click event... Feels awkward for how I'd really want to use this component.

andrew-frueh:
I had a similar thought about the ‘action’ parameter, Cory.  Really all it does is add an X right now.  Seems like maybe we’d expect more - like having an Emitter for chip events, etc.

So maybe I can start you you.  What would be on your wishlist for the action parameter functionality?

corykon:
I might have better ideas when I actually need to implement the chip component in popbuilder...

My first thought we be to totally deprecate the action param. I think the `(click)` event should be emitted and work regardless whenever the chip is clicked on - that may already be the case. Than either allow users to project their own close icon button into the chip, or preferably, have a `[hasCloseButton]` input, which adds the button (basically does what the action input does today) and then set up a new event emitter called `(closeClick)` or something.

andrew-frueh:
Great suggestions. Would your expectation be that only clicking the X would close the chip? Or clicking anywhere on it would close?  Is that an option we should allow the dev to specify?  What would you call that - onlyIconCloses or something like that?

corykon:
If clicking x or the chip body should close/remove it, then I'd prob just
attach the same function to `(click)` and `(closeClick)`

andrew-frueh:
Ah I like it.  So add one close() function to the component.  And allow the dev to bind that to (click) for click to close anywhere; or (closeClick) which is a new emitter just for clicks on the close icon.

I think that makes good sense.  Did I understand you correctly?

corykon:
Yup that sounds right

andrew-frueh:
Regarding the `close()` piece of this.  Seems like the only way to really do that effectively is with a service.  Does it make sense to add a ChipService?  Or is that usually something that an app would want to do on their own?

andrew-frueh:
On second thought, it looks like Angular Material isn't doing anything like a service for their chips.  They're just handling the show/hide with an `ngFor` on an array.  So I'll probably just stick with something like that.  And as per your previous suggestion Cory - I'll keep the second example simpler.

andrew-frueh:
Thanks for the discussion on this one @corykon - pushed an update that feels like it makes the component more intuitive, and simplifies the examples too. Let me know what you think.

andrew-frueh:
Thanks @corykon - actually there shouldn't be anything breaking on this.  Before - the click was binding to the whole button, it's just the example I had where it was hiding the close icon because I had bad code.  So no breaks here - just an additional option now to bind events to just the button or the whole chip (with `click`)

PR #1052, opened 2019-11-08
docs(app): applications insights

andrew-frueh:
adds Azure Application Insights page view tracking

### DIFF
--- a/src/environments/environment.prod.ts
+++ b/src/environments/environment.prod.ts
@@ -1,3 +1,4 @@
 export const environment = {
-    production: true
+    production: true,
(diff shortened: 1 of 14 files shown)


src/environments/environment.prod.ts
```diff
@@ -1,3 +1,4 @@
 export const environment = {
-    production: true
+    production: true,
+    instrumentationKey: '7ce7e52a-addf-427b-91cf-094dd0ea1df0'
```
andrew-frueh: Any concerns @joeskeen with including our key in the GitHub repo?  Any other way to do this that keeps it hidden?  I think the worst someone could do is post bad data to our AppInsights instance.

PR #1057, opened 2019-11-18
feat(pipes): add pipes to Cashmere documentation

### LINKED ISSUE #1010: Cashmere Pipes
**Describe your idea**

There are a handful of common pipes we'd like to include in Cashmere.

@benjanderson already contributed an Ellipsis pipe, which has been an undocumented feature of Cashmere for a while.  We would like to add additional pipes and feel this is a good time to surface them in the documentation as well.

As per Slack conversation, we would like to just add an additional section to the Components page for pipes.  If we end up with too many pipes and feel like we'd rather have them on a separate page, we can later separate that out.

Also, we would like to separate each pipe out into their own module.  This will allow you to import as few or many pipes as you need (and works well with #960 in mind).  We will also provide a `PipesModule` that will export all of the pipes if you just want them all.

joeskeen:
also revise the existing (undocumented) ellipsis pipe

fix #1010


src/app/app.component.html
```diff
@@ -2,7 +2,7 @@
     <hc-navbar-link uri="/home" linkText="Home"></hc-navbar-link>
     <hc-navbar-link uri="/styles" linkText="Styles"></hc-navbar-link>
     <hc-navbar-link uri="/components" linkText="Components"></hc-navbar-link>
-    <hc-navbar-link uri="/guides/getting-started" linkText="Guides"></hc-navbar-link>
```
joeskeen: This was causing a routing bug when navigating to any guides other than getting started.  The navbar didn't highlight guides as the active route.

src/app/components/component-viewer/component-viewer.component.ts
```diff
@@ -22,12 +22,15 @@ export class ComponentViewerComponent implements OnInit {
...
+                let availableSections: string[] = [];
 
                 if (this.docItem) {
                     const examples = this.docItem.examples;
                     if (examples && examples.length > 0) {
-                        availableSections = ['Examples', 'API'];
+                        availableSections = ['Examples'];
+                    }
+                    if (!this.docItem.hideApi) {
+                        availableSections.push('API');
```
joeskeen: Because of the folder structure for the pipes, the dgeni process doesn't generate API documentation for pipes.  We may want to change this in the future, but the API surface for pipes is very small and pretty consistent (just a transform function that accepts a value).  I instead document the API in the usage markdown file.

src/app/core/document-items.service.ts
```diff
@@ -3,10 +3,10 @@ import {Injectable} from '@angular/core';
 export interface DocItem {
     id: string;
     name: string;
-    // Category options are: 'forms', 'nav', 'layout', 'buttons', 'popups', 'table'
-    category: string;
+    category: 'forms' | 'nav' | 'layout' | 'buttons' | 'popups' | 'table' | 'pipes';
```
joeskeen: I thought that the string literal union was better than a comment here ;)

andrew-frueh:
Woohoo!  Thanks for getting this in @joeskeen!  As per our discussion in the last steering committee, for big PRs like this one - I'll schedule a short meeting on all our calendars for anyone who is available to make.  And then we can review it live together and discuss (rather than asking everyone to do it independently).

joeskeen:
Here is a screenshot of the Ellipsis pipe in the documentation:
image

joeskeen:
I've also nested the Ellipsis pipe (and future pipes) under their own folder within the `pipes` folder under `projects/cashmere/src/lib`:
image

Issue #1044, opened 2019-10-31
App Switcher Issues


corykon:
Few issues I ran into setting up the app switcher. Couple are nitpicky but figured they're thoughts worth capturing just in case. Feel free to triage and split these up as necessary.

- **Cleanup loading styles.** Just so the app switcher is more style compliant, would be nice to clean up the loading state by either using a cashmere loader, or a skeleton screen. Frankly it's unlikely for a user to see this screen, but it'd be nice to get it looking cleaner nonetheless. Some images of current and suggested styles are below.
- **Firmer guidance for sizing in the component** Default is 100px, but this feels a little big. Somewhere around 75px might be nice. I assumed we want webapp apps to use the default, but I wasn't sure. Probably want to make sure we're all aligned here, so I'd appreciate firmer guidance in the docs. 
- **Better support (or documentation) for runtime initialization** Common usage (or at least for both popbuilder and LW) is to get the DiscoveryServiceUrl at runtime from an API call. In order for this to work, you either have to avoid initializing the app switcher by not adding it to the DOM (with an *ngIf or something), or you have to override the AppSwitcherService in a way that prevents the component from breaking on initialization, since it tries to fire a service request immediately. I found it was more complicated to set up for my use case than is detailed in the usage guide. We should either shore up the docs some more, or provide a way to pass in the url after the app switcher is initialized.
- **Document the required sizing for app registration.** The icons in the app switcher have to be at a ratio of 177.7 x 100. I passed up an icon that was a square not knowing the image needed to have padding built in so it looked a little goofy at first. We should document this.

*Current styles*
![image](https://user-images.githubusercontent.com/12416432/67967349-16258000-fbcb-11e9-8828-6928e2f47ba5.png)

*Suggested (Placeholders would have a pulsing opacity animation going on)* 
![image](https://user-images.githubusercontent.com/12416432/67967380-22114200-fbcb-11e9-882c-8259f3af7f56.png)

andrew-frueh:
**Regarding item #1 -** 
Steering committee agrees that the style on "loading applications" should be a spinner.  And if fails to load, should display that as a warning or message...not just spin forever.

andrew-frueh:
**Regarding item #3 -** 
Have the app switcher load when the popover is accessed, not on page load.  Currently discoveryServiceURL most be known at module init, and cannot be loaded async; need a way to get that async at runtime.

andrew-frueh:
[...]

PR #1064, opened 2019-12-05
feat(app switcher): add loading, failure, and empty states

corykon:
re #1044

@joeskeen This'll likely cause some conflicts with your work! Sorry! Feel free to grab me to work on merging if needed.

**Empty State:**
appSwitcher-empty

**Loading and Fail State:**
appSwitcher-fail

For testing, you can make tweaks in `...Fabric.Cashmere\projects\cashmere-examples\src\lib\navbar-app-switcher\custom-app-switcher.service.ts`

```
getApplications(): Observable<IDiscoveryRequest> {
    return of({
        value: []
    }).pipe(delay(5000)).pipe(tap(() => { throw Error("hi"); }));
}
```


joeskeen:
@corykon Do we no longer have a View All My Applications URL?  I would think that should be available even on error or empty states if configured.

projects/cashmere/src/lib/app-switcher/app-switcher.module.ts
```diff
@@ -4,22 +4,25 @@ import {HttpClientModule} from '@angular/common/http';
 
 import {AppSwitcherComponent} from './app-switcher.component';
 import {PopModule} from '../pop/popover.module';
+import {ProgressIndicatorsModule} from '../progress-indicators/progress-indicators.module';
 import {AppSwitcherService} from './app-switcher.service';
+import { WorkTrackerService } from '../picklist/services/work-tracker.service';
```
joeskeen: ...so app switcher is now dependent on picklist?  Seems strange to me...

corykon: Just a helper service that I didn't realize it was sitting within the picklist directory. I can move it to a more general location if you'd like.

joeskeen: Maybe this service should belong someplace else if it is used in multiple places

corykon:
> @corykon Do we no longer have a View All My Applications URL? I would think that should be available even on error or empty states if configured.

@andrew-frueh would you agree that I ought to add the link to the empty and error states? While I'm at it, could I move the link up top? Sometimes that right aligned at the bottom feels a little off, I think because of the center aligned feeling of the app icons.

This would be the tweaked version:
image

projects/cashmere/src/lib/app-switcher/app-switcher.module.ts
corykon: @joeskeen do you think `...\Fabric.Cashmere\projects\cashmere\src\lib\shared` would be an ok place to move `WorkTrackerService`?

andrew-frueh:
I worry that the View All at the top might be a little less findable, but aesthetically I agree it's a big improvement.   So let's do it.  Just double check your vertical alignment on the view all button.  Right now it's aligned with the descender on the 'p' of Applications, so it's a little off.  Could either nudge it up 3px or so, or maybe `align-items: center` if that looks better

andrew-frueh:
Also @corykon - you could probably tighten up the padding-bottom on that version with View All at the top.  There's more white-space than we need below the thumbnails in that example.

corykon:
@andrew-frueh Agreed on all points. That was just an image from adobe where I covered up the bottom link and quickly threw a new one up top. When I do it for real I'll be sure it's tight.

corykon:
@andrew-frueh here's the real deal. Feedback before I merge?
image

andrew-frueh:
Looking great @corykon.  Hey can you add one css line while you're in there, below line 13 in app-switcher.component.scss - can you add `padding-left: 2px`?  For some reason "Applications" looks like it's falling off the left edge of the line below it.  That's not something new - it's been that way, but I just noticed it now.

andrew-frueh:
Thanks @corykon - looks like you're good to merge your branch

PR #1065, opened 2019-12-06
Open space/app switcher current app

jacquie:
My mission was to disable and somehow distinguish the icon for the current app in the app switcher.
Added two `@Input()`s to the component for the app name and version which should correspond to the serviceName and version registered with the discovery service.
Added functionality to determine the current app and then remove the href attribute for it.
Added a class for styling the icon - just made it 0.5 opacity for now.
Edited markdown file to document the added input parameters.


joeskeen:
Could you provide a screenshot @jacquie ?

projects/cashmere/src/lib/app-switcher/app-switcher.component.html
```diff
@@ -12,9 +12,9 @@ <h4>Applications
     <ng-container *ngIf="applications?.length > 0">
         <div class="apps">
             <div class="app" *ngFor="let app of applications">
-                <a class="thumbnail" [title]="app.Description" [href]="app.ServiceUrl" target="_blank">
-                    <img class="thumbnail-img" [height]="iconHeight" [src]="app.Icon" [ngStyle]="{'max-width': iconHeight + 'px'}">
```
andrew-frueh: Hey @jacquie - looks like the `ngStyle` block I added for max-width got lost in the merge.  That's important because it keeps the icons from overflowing their square.

jacquie: Oh no - sorry!  I'll add it back in...

projects/cashmere/src/lib/app-switcher/app-switcher.component.scss
```diff
@@ -37,6 +37,10 @@
 
                 & > .thumbnail-img {
                     width: auto;
+
+                    &.highlight-current {
+                        opacity:0.5;
```
andrew-frueh: I think dropping the opacity of the current app works for now.  If we get to the point of distinguishing what apps the user has access to, then we'll need to differentiate those from the current app.  But we can cross that bridge when we get to it.  @corykon - any other suggestions?

projects/cashmere/src/lib/app-switcher/app-switcher.component.html
andrew-frueh: hooray!  thank you :-)

projects/cashmere/src/lib/app-switcher/app-switcher.component.html
```diff
@@ -12,9 +12,9 @@ <h4>Applications
...
-                    <div class="title">{{app?.FriendlyName | ellipsis:18}}</div>
+                <a class="thumbnail"  [title]="app.Description" [attr.href]="linkIfNotMe(app)" target="_blank">
```
andrew-frueh: `linkIfNotMe` is nice!  I was just going to ask if we should disable the link for the current app, but I see you're way ahead of me 👍

andrew-frueh:
Ugh - sorry @jacquie, I merged your branch in locally, didn't mean to merge it for everyone.  Oh well, if you have changes - feel free to push them right into the main branch.

And @joeskeen - here's a screenshot:
Screen Shot 2019-12-06 at 4 07 16 PM

jacquie:
It looks like my merge messed up an import in `app-switcher.component.spec.ts` and I just updated it directly in here...

projects/cashmere/src/lib/app-switcher/app-switcher.component.scss
corykon: @andrew-frueh I'd think if an app isn't available (when we get there), we may not show it in the list at all. If we do show disabled apps (perhaps cause we want people to see the apps and ask for them?), we might want opacity for those.

Opacity is probably alright for now for the current app, but I'd add a tooltip (`[title]` attribute) that says something like "current application" to clarify why its disabled just in case it's not obvious to some.

I played around with some other ideas while listening to a class lecture, but I don't know that anything is really any better.
Artboard – 2

andrew-frueh: Ugh - you have a lecture on Friday nights!  That's rough.  But thanks for the suggestions, I took two of your ideas and just merged them into the branch.  I have the opacity dropping on both the icon and the text now as per your first example.  And I also added logic to have the title say "Current Application" if it's grayed out.

PR #1068, opened 2019-12-10
style(app-switcher): css renaming

andrew-frueh:
Adjustments to the css naming on the app switcher to match our current conventions.


andrew-frueh:
Hey that's a really good question @jacquie and remind me to bring it up in the steering committee meeting tomorrow.  The way I've thought about it is that any root level class, one that isn't nested in another class via {} or & should have the hc- prefix, because it eliminates any chance of collisions.  For the ones that are nested, I often still do it when I'm writing things from scratch, but it's probably less important.  But like I said, let's discuss tomorrow

PR #1093, opened 2019-12-17
feat(dynamic input support): richer errors & better select change event

c0d3r1010:
support better error messages with a rich error object for hcFormControlComponent and better change
event support for select component

### DIFF
--- a/projects/cashmere/src/lib/form-field/hc-form-control.component.ts
+++ b/projects/cashmere/src/lib/form-field/hc-form-control.component.ts
@@ -3,6 +3,14 @@ export abstract class HcFormControlComponent {
     /** Whether the control should be displaying an associated error */
     _errorState: boolean = false;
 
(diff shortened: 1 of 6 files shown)


joeskeen:
Just a note on the motivation for these changes (for posterity's sake): Currently the error messages are pretty static (you have to include them in `hc-error` inside `hc-form-field`, but for the dynamic input control for IDEA, we needed a way for the control itself to declare not only when it was in an error state, but what the error is.

For example, the dynamic input can be a text field, a date picker, a select box, a checkbox, etc.  For text fields, it can represent text, number, etc.  Each of these have their own validation rules, and as such they can produce a variety of validation errors.  With these changes, the dynamic input can report its own validation errors, which will be displayed above the other `hc-error`s in the form field template (if any).

This isn't a breaking change, but does make it possible for richer form control-compatible components to be built.

The changes for the select component improve the emitting of the change event and thus help trigger validation quicker.

PR #1097, opened 2019-12-20
feat(daterange): selected preset emitter

### LINKED ISSUE #1095: Get identifier or object of selected preset from date-range
**Description**

It would be great if the date-range would uniquely identify which preset was used.

Imagine your application supplies a date-range component with user-configured presets that are dynamic with respect to today. Here is an example. A user can decide that their fiscal year starts on July 1st. They could also could add two presets of "2 Quarters to Date" and "Fiscal Year to Date". On any given day, a component outside the date range calculates the start and end dates of these presets and supplies them, with appropriate labels, to the date-range component.

image

However, sometimes, these presets have the same start and end dates. Suppose it is December 17th, 2019. "2 Quarters to Date" is 7/1/2019 to 12/17/2019. Likewise, "Fiscal Year to Date" is also 7/1/2019 to 12/17/2019.

When a user clicks the Apply button, the date-range emits only the start and end dates. Components outside of the date-range component do not know which of the two presets were selected. Of course, those two presets have the same start and end dates, but presentationally, it's impossible to know which preset the user actually used within the date-range.

Notice in the following screenshot that the selected preset is shown to the user:

image

If the "Fiscal Year to Date" preset is selected from the date-range, the external components only get  the start and end dates and don't know which preset to display:

0D91D903-8F29-401A-86AD-4ED5402BA573

If the date-range required unique labels for the presets or allowed for a identifier to be supplied with a preset and then emitted this identifier with the selected start and end date, it would be possible to distinguish between presets that have different labels, but the same start / end dates.

andrew-frueh:
adds alternate emitter of preset index or daterange (if no preset)

closes #1095

### DIFF
--- a/projects/cashmere/src/lib/date-range/picker-overlay/picker-overlay.component.html
+++ b/projects/cashmere/src/lib/date-range/picker-overlay/picker-overlay.component.html
@@ -32,7 +32,7 @@
     <div class="hc-date-range-calendar-item">
         <div class="hc-date-range-menu">
             <hc-radio-group class="presets" [(ngModel)]="_selectedPreset">
-                <hc-radio-button *ngFor="let p of options.presets; let i = index" [value]="p.range" [id]="i" (change)="_updateRangeByPreset(p.range)">{{ p.presetLabel }}</hc-radio-button>
+                <hc-radio-button *ngFor="let p of options.presets; let i = index" [value]="p.range" [id]="'hc-date-range-preset-'+i" (change)="_updateRangeByPreset(p.range)">{{ p.presetLabel }}</hc-radio-button>
             </hc-radio-group>
             <div class="hc-date-range-controls">
                 <button hc-button buttonStyle="link" type="button" (click)="_discardNewDates()">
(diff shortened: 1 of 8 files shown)


projects/cashmere/src/lib/date-range/picker-overlay/picker-overlay.component.html
```diff
@@ -32,7 +32,7 @@
     <div class="hc-date-range-calendar-item">
         <div class="hc-date-range-menu">
             <hc-radio-group class="presets" [(ngModel)]="_selectedPreset">
-                <hc-radio-button *ngFor="let p of options.presets" [value]="p.range" (change)="_updateRangeByPreset(p.range)">{{ p.presetLabel }}</hc-radio-button>
+                <hc-radio-button *ngFor="let p of options.presets; let i = index" [value]="p.range" [id]="i" (change)="_updateRangeByPreset(p.range)">{{ p.presetLabel }}</hc-radio-button>
```
njwhc: This is a rather minor suggestion, but what do you think about adding a prefix or suffix to the assigned id? As is, it seems like these ids have a somewhat high risk of violating the id uniqueness constraint. Perhaps something like `[id]="'hc-date-range-preset-' + i"`

andrew-frueh: Great suggestion!  Thanks for the review Nick - just pushed a commit with your suggestion.  @jacquie - how does this all look to you.  I'll need your approval before I can merge :-)

andrew-frueh:
Thanks again @jacquie - looking forward to hearing how this works for you guys.  I'll release a new version of the library tomorrow that will have this included.

PR #1107, opened 2020-01-11
refactor(examples): convert colors to variables

### LINKED ISSUE #920: Convert Example's scss Colors to Variables
**Describe your idea**

As per @corykon 's suggestion, it would be better if our color values in our Cashmere examples don't use hard-coded hex values.  They should really be using our color variables instead.  The problem has been figuring how to import the color variables sheet.

Using hard-coded hex values sets a bad precedent, when we're trying to have developers use color variables instead

andrew-frueh:
Finally figured out how to allow stylesheets in examples to import files from the library. Updated all the instances of hard-coded color values to variables wherever possible (includes a few components that slipped through the cracks).

Please do me a favor and pull this branch down to validate the imports are working in your environment as well.

closes #920


andrew-frueh:
Shoot - too good to be true.  And yet somehow it's working in my environment now but not on Travis.  Gonna have to try and sort this out some more.

andrew-frueh:
Okay @joeskeen and @corykon - I think I've got it working on Travis now (as well as locally).  I created a cashmere dependency for the examples and updated the Travis build script to make sure and run `npm i` in the examples directory.  For this to work on your env, you'll probably need to do an install in `projects/cashmere-examples` as well.  I added this step to the contribution guide.

This does add another step to the environment setup, and it feels a little convoluted.  Open to any other simpler ideas you might have...

andrew-frueh:
Ahhh - that's so much better than having to do a separate `npm i`.  Thanks for the suggestion @corykon - I'll make that change.

andrew-frueh:
Worked like a charm @corykon.  Brilliant!  Let me know the update works for you.  I rolled back a few commits and replaced with the current one

PR #1112, opened 2020-01-14
fix(datepicker): support preset overlapping values

### LINKED ISSUE #1111: More date range preset picker woes
**Summary**

Although the chosen preset is being communicated back out of the date range picker there are still some odd things happening when the presets coincidentally happen to describe the same date range.

**Reproduction**
In this example the current date is Jan 13 and the last 3 presets are 'month to date', 'quarter to date' and 'year to date' which all happen to map onto Jan 1 - 13
When I select the 'month to date' button it actually selects the 'year to date' button instead
Interestingly if I then select the 'month to date' button again it does select it correctly.

daterangepicker

Also if I select 'month to date' apply it, and then reopen the dialog the 'year to date' option is preselected as shown here:
wrongpreset

There is no way of passing in the preselected preset item and it is calculating it from the passed in dates and not always getting it right.  Would it be possible to add something like an optional 'currentlySelected' property to the `PresetItem` interface so it doesn't need to guess

andrew-frueh:
Allow multiple presets with same date ranges & selectedDate may be number.  @jacquie the DatePicker definitely wasn't set up to support your use case, so it required some well needed house cleaning.  The good news is that it's leaner and meaner than before - and I believe it now should do what you need it to.  I've updated our example a prove out the functionality (it has two presets with the same value).

Please give it a good look over and ensure that the functionality looks like it's got you covered.

closes #1111

### DIFF
--- a/projects/cashmere-examples/src/lib/date-range/date-range-example.component.html
+++ b/projects/cashmere-examples/src/lib/date-range/date-range-example.component.html
@@ -7,7 +7,7 @@
         popperPlacement="bottom"
         (selectedDateRangeChanged)="updateRange($event)"
         (selectedPresetChanged)="updatePreset($event)"
-        [selectedDate]="range"
+        [selectedDate]="selected"
         [options]="options"
         #pickerOne
     >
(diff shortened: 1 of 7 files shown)


andrew-frueh:
Great find on this one @jacquie and thanks for the review!  I checked the current site and it seems to be an existing issue, but I'll see if I can track it down.

jacquie:
Meant to add thanks for jumping on this one so quickly!

andrew-frueh:
Phew!  Good catch on that one @jacquie - what was happening was when a preset was selected, the activeDate (the value displayed on the calendar) was being blocked because the start date can't be greater than the end date and visa versa.  But in the case you mentioned, the preset was trying to be set based on the previous min/max.  So the fix came down to updating the activeDate after the preset has been recorded.

PR #1113, opened 2020-01-14
fix(navbar): dynamic links & switcher styles

### LINKED ISSUE #1110: Navbar + AppSwitcher Issues
**1. More menu triggering too early and then getting stuck.** 
navbarMoreBug

Was going to create a stackblitz to repro but noticed the navbar stackblitzes don’t work:
- Doesn't load: https://stackblitz.com/run?file=src/app/navbar-app-switcher/navbar-app-switcher-example.component.html&hidedevtools=1
- Loads but the more menu isn't working, and the images don't load in: https://stackblitz.com/run?file=src/app/navbar-overview/navbar-overview-example.component.html&hidedevtools=1 

-------------------------------------------

**2. App switcher icons are squished**
Probably should be able to handle icons regardless of the size it is given. (Tricky but should be doable in css)
image

andrew-frueh:
navbar supports links added after load & non-square app switcher icons

@corykon this should knock out two of the three issues you reported.  Can we put the table styles one into a separate issue?

closes #1110

### DIFF
--- a/projects/cashmere/src/lib/app-switcher/app-switcher.component.html
+++ b/projects/cashmere/src/lib/app-switcher/app-switcher.component.html
@@ -18,7 +18,11 @@
                     [title]="appIsMe(app) ? 'Current Application' : app.Description"
                     [attr.href]="linkIfNotMe(app)"
                     target="_blank">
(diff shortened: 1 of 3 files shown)


corykon:
Split out the table stuff into #1114

PR #1115, opened 2020-01-15
fix(datepicker): two digit year conversion

### LINKED ISSUE #1042: Two-digit year input to datepicker yields 1900-based in IE
**Summary**

If a date with a two-digit year is manually typed into a datepicker, IE will produce a date with the year in the 1900s. Chrome and Firefox produce dates with the year in the 2000s. My best guess is that this is the result of Cashmere using `new Date(inputString)` or similar to produce the date, which will always give a year in the 1900s in IE.

### IE, input 2/2/19 becomes 2/2/1919:
IE-datepicker-with-two-digit-year

### Chrome, input 2/2/19 becomes 2/2/2019:
Chrome-datepicker-with-two-digit-year

**Reproduction**

Steps to reproduce:

1. Type a date with a two-digit year, say `2/2/19` into a datepicker.
1. Tab or click out of the input

~~Stackblitz reproduction~~ Stackblitz does not support IE.

**Additional context**

A partial workaround is to use Angular Material's moment date adapter with Cashmere datepickers. However, this workaround appears to **not** work with Cashmere daterange components.

```js
import {
  MAT_MOMENT_DATE_FORMATS,
  MomentDateAdapter
} from '@angular/material-moment-adapter';
import {
  DateAdapter,
  HC_DATE_FORMATS,
  HC_DATE_LOCALE,
} from '@healthcatalyst/cashmere';

@NgModule({
  ...
  providers: [
    { provide: DateAdapter, useClass: MomentDateAdapter, deps: [HC_DATE_LOCALE] },
    { provide: HC_DATE_FORMATS, useValue: MAT_MOMENT_DATE_FORMATS },
  ]
```

andrew-frueh:
Fixes two digit year conversion in IE to match other browsers.  This correction applies to both the datepicker and daterange components.

closes #1042

### DIFF
--- a/projects/cashmere/src/lib/datepicker/datepicker-input/datepicker-input.directive.ts
+++ b/projects/cashmere/src/lib/datepicker/datepicker-input/datepicker-input.directive.ts
@@ -329,7 +329,7 @@ export class DatepickerInputDirective implements ControlValueAccessor, OnDestroy
         if (date) {
             let inputString: string = this._elementRef.nativeElement.value;
             /** Skip this check if the input string contains any 3+ digit numerical values - assumed to be a year */
-            if (!inputString.match('[1-9][0-9][0-9]')) {
+            if (!inputString.match(/[1-9][0-9][0-9]/g)) {
                 let currentDate = new Date();
                 if (date.getFullYear() >= currentDate.getFullYear() + 30) {
                     date.setFullYear(date.getFullYear() - 100);
--- a/projects/cashmere/src/lib/datepicker/datepicker-input/datepicker-input.directive.ts
+++ b/projects/cashmere/src/lib/datepicker/datepicker-input/datepicker-input.directive.ts
@@ -323,6 +323,22 @@ export class DatepickerInputDirective implements ControlValueAccessor, OnDestroy
 
     _onInput(value: string) {
         let date = this._dateAdapter.parse(value, this._dateFormats.parse.dateInput);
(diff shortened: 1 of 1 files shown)


projects/cashmere/src/lib/datepicker/datepicker-input/datepicker-input.directive.ts
```diff
@@ -323,6 +323,22 @@ export class DatepickerInputDirective implements ControlValueAccessor, OnDestroy
 
     _onInput(value: string) {
         let date = this._dateAdapter.parse(value, this._dateFormats.parse.dateInput);
+        /** Two-digit year input conversion method for IE
+         * Based on the current year, assume that the four-digit year date should be in
+         * either the next 30 years, or the preceding 70 years */
+        if (date) {
+            let inputString: string = this._elementRef.nativeElement.value;
+            /** Skip this check if the input string contains any 3+ digit numerical values - assumed to be a year */
+            if (!inputString.match('[1-9][0-9][0-9]')) {
```
isaaclyman: This works, but it might be clearer to write it as a regular expression. `/[1-9][0-9][0-9]/`

andrew-frueh: Gah - good catch.  I screwed up my syntax there, meant it to be a regular expression but wrote it wrong.  Just pushed a commit to correct that.  Thanks @isaaclyman

Issue #1034, opened 2019-10-17
RFC: Cashmere documentation rewrite strategy


isaaclyman:
@andrew-frueh has asked me to spearhead improvements to the Cashmere documentation. Please read through this RFC and leave comments below, everything is open to change. I'll incorporate your suggestions as we go along.

-----

## Introduction

Cashmere will be eminently more accessible and easier to adopt if our documentation is good. I propose a few guiding principles for this project:

1. The quality of documentation can be measured by how quickly the reader finds what they need and leaves.
1. A code snippet is worth a thousand words.
1. Library code is not documentation. A component, public class, method, or attribute is "documented" if the user can easily find an example of it in action and confidently modify it for inclusion in their own project. Otherwise, it is undocumented.
1. If something is undocumented, it doesn't exist for our users.

## Structure: Components

We'll structure documentation for the Components pages by making sure the following questions are answered.

<sup>In the first few sections of a Components page, raw code snippets will be preferred over Stackblitz frames. Code snippets are easier to scan through quickly, require less interaction, and are easier to maintain.</sup>

_How do I use this component?_

Each page will begin with a code snippet (usually only HTML, but HTML and TypeScript in special cases) representing the most common usage of the component. Most users will copy and paste this code and then leave.

If there are multiple common ways to use the component (e.g. using the `hc-table` class on a standard HTML table, or using the `hc-table` attribute on a table element structured as a CdkTable) we may include more than one code sample in this section.

_How do I use this component's additional template features, like attributes, events, related components, and content projection?_

Here we'll address each potential modification that can be made to the component as part of the HTML template where it's used. Each Input `[attribute]` will have a type signature, an HTML code snippet, and a list of possible values that the user can provide. Each Output `(event)` will have a function signature and a TypeScript code snippet. Each related component (e.g. `hc-label` is a related component for `hc-form-field`; `hc-modal-body` is a related component for `hc-modal`) will have an HTML code snippet. Each component that allows content projection will have an HTML code snippet demonstrating this, if not already demonstrated by previous examples. A brief description of each of the above will also be present.

_How do I call methods or access fields on this component's public class?_

Here we'll demonstrate how to access each component's public class from your own code and use any public methods or fields it provides (e.g. HcPopComponent provides a `close()` method, among others). Each method and field will be represented in a TypeScript code snippet (multiple features can share a code snippet where appropriate, such as `isOpen()`, `close()`, `toggle()` and `open()`).

[...]

PR #1117, opened 2020-01-15
1034 documentation rewrite example

isaaclyman:
This is a draft of how we might want the documentation for the Button component to look, based on #1034. Feel free to poke about and offer suggestions.

I anticipate the following concerns:

_This is one of the simplest components in Cashmere but it required ~150 lines of handwritten documentation, a non-trivial amount of research, and a new example. And it didn't even need an FAQ, like some components will. That portends a **lot** of work to get all the documentation finished._

Yeah, good documentation is no joke. 😅

_How will we keep all this handwritten documentation up to date as components change?_

Dedicated effort; see above.

_The Usage page is a little hard to navigate without a table of contents._

True. @joeskeen and I have discussed that a TOC would be great to have. I'm not sure how to pull that off yet, so I'm saving the effort for another time.

_Could we use a more declarative style on the Usage page to reduce duplication?_

Potentially, yes, but I don't know if Markdown will play nice. Maybe it can give us partials, at least? Something to look into.

### DIFF
--- a/projects/cashmere-examples/src/lib/button/button-example.component.html
+++ b/projects/cashmere-examples/src/lib/button/button-example.component.html
@@ -1,0 +1,3 @@
+<button hc-button buttonStyle="primary" size="md">
+    Primary, Medium
+</button>
--- a/projects/cashmere-examples/src/lib/button/button-example.component.scss
+++ b/projects/cashmere-examples/src/lib/button/button-example.component.scss
@@ -1,0 +1,3 @@
+button {
+    margin: 10px;
+}
--- a/projects/cashmere-examples/src/lib/button/button-example.component.scss
+++ b/projects/cashmere-examples/src/lib/button/button-example.component.scss
@@ -1,3 +1,0 @@
-button {
-    margin: 10px;
-}
--- a/projects/cashmere-examples/src/lib/button/button-example.component.ts
+++ b/projects/cashmere-examples/src/lib/button/button-example.component.ts
@@ -1,0 +1,11 @@
+import {Component} from '@angular/core';
+
+/**
+ * @title Button (basic example)
+ */
+@Component({
+    selector: 'hc-button-example',
+    templateUrl: 'button-example.component.html',
+    styleUrls: ['button-example.component.scss']
+})
+export class ButtonExampleComponent {}
--- a/projects/cashmere/src/lib/button/button.md
+++ b/projects/cashmere/src/lib/button/button.md
@@ -106,6 +106,8 @@ For a split button (a button with a caret and a dropdown menu), replace `<button
 
 ##### Methods
 
+First, get a reference to the element.
+
 ```html
 <button hc-button #buttonElement></button>
 
--- a/projects/cashmere/src/lib/button/button.md
+++ b/projects/cashmere/src/lib/button/button.md
@@ -1,9 +1,148 @@
-##### Split Buttons
+##### Attributes
 
-**hc-split-button**
+**[buttonStyle]**
 
-To use `hc-split-button` you encapsulate the content you want to be shown on the button. Anything marked with the directive `hcButtonItem` will get transferred into the dropdown menu. Everything else will be shown as the main button's content
+`buttonStyle: 'primary' | 'primary-alt' | 'secondary' | 'destructive' | 'neutral' | 'minimal' | 'link' | 'link-inline' = 'primary'`
 
-**hcButtonItem**
+Sets the color and style of the button.
 
-Marks an element for being put into the dropdown menu of the button. Multiple things can be marked with this directive
+```html
+<button hc-button buttonStyle="primary">button</button>
+```
+
+**[size]**
+
+`size: 'sm' | 'md' | 'lg' = 'md'`
+
+Sets the size of the button.
+
+```html
+<button hc-button size="md">button</button>
+```
+
+**[disabled]**
+
+`disabled: boolean = false`
+
+Puts the button in a disabled state.
+
+```html
+<button [disabled]="true">button</button>
+```
+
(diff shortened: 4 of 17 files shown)


corykon:
Awesome work, Isaac. There's a lot of stuff in here I really like. Feels more approachable.

**Stuff I like:**
- Super stripped down example front and center
- Code examples for each attribute, method, event, etc is awesome
- Related components section is an excellent idea
- Like how the methods sections first shows and example of how to get a reference to the component
- Overall the tone is great, descriptions clear and concise. You are a good writer :+1:

**Concerns/Critiques:**
- I agree that a TOC would be very helpful
- On the usage page, towards the bottom of the attributes section, things feel a little muddy because we've got split-button attributes mashed right in with the regular hc-button attributes. Feels like this could be clearer.
- A lot of the handwritten stuff that is duplicative to what's on the API page will become a nightmare to maintain. I bet we could find a nice way to merge the usage/api tabs and avoid the duplication. Perhaps some of the awesome new elements of your usage tab like the formatting, the code examples, etc could be included in the source files and then pulled in with dgeni? (Example, here in the old angularJS library code examples are right in the source code: https://github.com/angular/material/blob/master/src/components/button/button.js) It'll take some thought and effort to rework dgeni, but I think we're way more likely to keep things up to date if we can find a way to take that route. And we'd prob still land on a mix between dgeni and the very custom handwritten stuff, but I think we can fine tune that balance some more.

isaaclyman:
@corykon critique #2 was an easy fix. I have some ideas in mind for getting a TOC to work, so I'll jump on that soon.

I'm completely unfamiliar with the autogenerated documentation. It has some major issues at present (tons of duplicated docs--see how many times identical Button attributes are listed), but I'm open to incorporating it more if it can be configured to work better. I'm a little concerned it will be more work to get the automated docs up to an acceptable standard than to just write everything by hand.

andrew-frueh:
Wow - amazing start @isaaclyman!  Thanks so much for taking this on.  I'll echo the compliments from @corykon, you've got quite a talent as a technical writer.  The content on the new usage page is really well organized and readable.

A few thoughts about how we can move this forward:

1. For now, let's keep Examples first and Usage second in the tab order.  Until we have the majority of our pages converted, we don't want some of our less than stellar usage pages to be the first thing folks see when they click on a component.  That will give us the time to make these conversions without compromising the existing usability.
2. I think our header and paragraph styles are working against the overall readability on the usage page.  So I'll sign up for cleaning that up and pushing to your branch tomorrow.  The good news is that we have a really strong precedent to follow from the work we've done on the [Catalyst docs site](http://docs.healthcatalyst.com).  We spent a good bit of time designing those styles for readability, so I'll probably use that as my template.
3. And last - once we've got those styles updated, and we're ready to merge - let's be data-driven in our design.  The UX team has access to a platform called usertesting.com that we can use to do some simple user tests with a panel of angular developers.  So we can take a group that's never seen Cashmere before, and ask them whether our instructions are clear enough that they could get started.
4. As an aside from the user testing, I'd also like to set up a chat with @cedar-ave about our updated Cashmere documentation approach.  Marla is about the most knowledgeable person I know when it comes to tech docs, so I'd love to get her thoughts as well.

I think Cory raises a great point about the sustainability of handwritten docs - but I also don't want to stress about it too much until we do some testing and have data to support the optimal content developers want.  Once we have that locked in, then we can spend some more time seeing how much of it can be auto-generated.

Bottom line - you've done an amazing job getting the ball rolling here.  Let me work on those styles tomorrow, and then next week we can start talking about a test plan.  Thanks again!

isaaclyman:
@andrew-frueh I've set the default tab back to Examples. Keep me in the loop on the other things you mentioned.

andrew-frueh:
thanks @isaaclyman - I haven't lost track of this, just had a few things get bumped up on the priority list.  Should be able to push a commit to this branch by Monday with those style updates I was talking about.

andrew-frueh:
@isaaclyman - sorry it took me so long to get back to this.  Had some time to look at styles today, and give some more thought to this initiative.  The concern @corykon raised about keeping the manually written docs current is one that I agree with.  But I wonder if there is a middle ground we can take with the written docs, so they complement the generated API docs, rather than replace them.

As with many things, I think Angular Material gives us a pretty good template in how they divide their sections:

https://material.angular.io/components/radio/overview

Their "Overview" section is what we're lacking, and represents the niche that I think your work could fill.  As with the example section you wrote for Button, the Overview page could represent the quickstart guide - essentially providing all the real-world, practical information a person would realistically need to keep up and running with a component.  It doesn't have to account for every last parameter, only the ones that are most useful.  And for devs who really need to get into the weeds and check for specific info, that's what we have the generated API docs for.

In my mind, I could see it going like this:

1. A seasoned angular developer who is familiar with Cashmere just needs to grab a code snippet to insert, say, a popover menu in to their app.  So they just hit the examples page, flip to the HTML, copy/paste and they're done.
2. A developer who is newer to Angular, or to the Cashmere library - is just getting started on a new web app, and although they can copy/paste from examples - they don't really have a good understanding of what's happening in that example code.  They would be our primary persona for an Overview doc - which they can read in a few minutes and walk away feeling armed with a solid understanding of how it works.
3. For a developer who is trying to troubleshoot and issue, or needs to do some heavy customization of a component - they want to know specific technical information about the guts of a component.  They should be able to call up the API section and find what they're looking for.

I think those three use cases all demonstrate unique instances where each section serves a valuable role - and I'd argue that as it stands now, we don't have a good answer to the "Overview" need.

I'll schedule a call with @cedar-ave for us to brainstorm on this live and talk through the layout of the content but also the overall strategy.

Thanks again for leading the charge on this @isaaclyman

PR #1129, opened 2020-01-17
feat(pop): submenus

### LINKED ISSUE #1089: Nested Popovers
**Describe your idea**

From the Touchstone product team:

> I think the biggest improvement that we could make to our app is to switch out our custom popover menus for Cashmere popovers. However I wasn't able to find functionality that allowed a nested menu, which we currently need in our app.

**Additional context**

Screen Shot 2019-12-12 at 1 49 13 PM

andrew-frueh:
allows hcMenus in hcPop to support submenus

closes #1089


andrew-frueh:
Thanks again for the review today @corykon and @Br00t4L17y!  Great suggestions, and I think the implementation is much better now.  Markup is a lot cleaner.  And no surprise, there was definitely some work to be done to support tertiary menus, or multiple submenus at the same level.  I've updated the example to prove out both of those cases - and I think it's pretty solid now.  But I'd love for you both to give it a good look over before anything gets merged.  Thanks!

PR #1132, opened 2020-01-23
style(input): thinner borders

### LINKED ISSUE #643: Change input border styles
Issue being created based on a convo with @andrew-frueh. 
The current border width for inputs is set to 1.5px. In certain browsers (typically on a Mac), 1.5px renders correctly and looks great. On others (often on Windows) 1.5px just renders as 2px, and looks too thick. 

Change border width to 1px, and make the border color a tad darker.

None, but open to other suggestions.

Styles for all components with half pixel borders. (Inputs, Selects, text areas, perhaps secondary button style) No breaking changes, just cosmetic.

**Additional context**
Will be important to look at radio buttons and checkboxes which have 2px borders, make sure 1px bordered form controls still look good next to them

image

andrew-frueh:
adjusts border width to 1.5px on all modern browsers

closes #643

### DIFF
--- a/projects/cashmere/src/lib/sass/select.scss
+++ b/projects/cashmere/src/lib/sass/select.scss
@@ -56,7 +56,9 @@
     -webkit-appearance: none;
     -moz-appearance: none;
     background-color: $white;
-    border: 1.5px solid $gray-300;
+    border: 1px solid $gray-300;
+    /* box-shadow used to add 0.5px to the border */
+    box-shadow: inset 0.5px 0.5px 0px $gray-300, inset -0.5px -0.5px 0 $gray-300;
     border-radius: 0;
     color: $text;
     cursor: pointer;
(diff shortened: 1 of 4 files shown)


projects/cashmere/src/lib/sass/select.scss
```diff
@@ -56,15 +56,17 @@
...
+    box-shadow: inset 0.5px 0.5px 0px $gray-300, inset -0.5px -0.5px 0 $gray-300;
     border-radius: 0;
     color: $text;
     cursor: pointer;
     // Font needs to be inherited, because by default <select> uses a system font
     font: inherit;
     height: 35px;
     line-height: 1.3; // avoid clipping character descenders in some browsers
-    padding-left: 7px;
+    padding-left: 9px;
```
andrew-frueh: @corykon - this one is dumb.  On every single browser, the about of padding left added to an option in a select element is different (even with appearance: none).  In Chrome there is none, in Firefox there is a lot, and IE/Safari somewhere in between.  This previous value was based on Firefox, and I realize in Chrome it looked off.  So I'm trying to split the difference here and bump it up slightly.

Issue #654, opened 2018-12-06
fixedTop directive


andrew-frueh:
**Describe the problem this feature request is solving**
We currently have a `fixedTop` parameter on both our navbar and subnav components which pins them to the top of the page on scroll.  The parameter on subnav had to be deprecated when the notification banner was added because it isn't compatible with additional pinned elements in between the navbar and subnav.  There is not currently a way to elegantly pin additional items to the top of the page beyond those two components.

**Describe the solution you'd like**
Create a `fixedTop` directive that is separate from individual components, but can be added to any element on the page.  This will allow anything to be pinned rather than only specific components, and will handle the ordering and z-index even if pinned items are removed dynamically.

**Describe alternatives you've considered**
Remove fixedTop from all components and have the apps handle their pinning themselves.

**If any, which components will have to change, and how**
-   Components changed:  **navBar, subNav**
-   Are these breaking changes to public interfaces:  **YES**

PR #1137, opened 2020-01-28
refactor(banner): convert to mixins

### LINKED ISSUE #1081: (Opt-in) Convert the Banner

isaaclyman:
Fixes #1081.

### DIFF
--- a/projects/cashmere/src/lib/banner/hc-banner.component.ts
+++ b/projects/cashmere/src/lib/banner/hc-banner.component.ts
@@ -11,7 +11,6 @@ import {parseBooleanAttribute} from '../util';
 })
 export class HcBannerComponent {
     private _type: string = 'info';
-    private _fixedTop: boolean = false;
     private _clickDismiss: boolean = false;
 
     /** If the banner can be dismissed, emits when the banner is clicked to close */
--- a/projects/cashmere/src/lib/banner/hc-banner.scss
+++ b/projects/cashmere/src/lib/banner/hc-banner.scss
@@ -1,110 +1,57 @@
-@import '../sass/colors';
-@import '../sass/functions.scss';
-@import '../sass/mixins.scss';
+@import '../sass/banner.scss';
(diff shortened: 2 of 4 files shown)


projects/cashmere/src/lib/banner/hc-banner.component.ts
```diff
@@ -11,7 +11,6 @@ import {parseBooleanAttribute} from '../util';
 })
 export class HcBannerComponent {
     private _type: string = 'info';
-    private _fixedTop: boolean = false;
```
isaaclyman: VS Code indicated that this property wasn't being used anywhere, so I deleted it. See also the `fixed-top` class below.

projects/cashmere/src/lib/banner/hc-banner.scss
```diff
@@ -1,110 +1,57 @@
...
-    display: flex;
-    align-items: center;
-    justify-content: flex-start;
-    width: 100%;
-    padding: 10px;
-    z-index: $zindex-subnav;
+    @include hc-banner();
+}
 
-    &.fixed-top {
```
isaaclyman: This `fixed-top` class isn't referenced anywhere in Cashmere (documentation, examples, etc.). It appears to be orphaned. I deleted it, but I can add it back if we believe it's in use.

projects/cashmere/src/lib/banner/hc-banner.component.ts
andrew-frueh: hey @isaaclyman - you're right about this one.  I believe it was added but then we realized it didn't play nice with subnav.  There's an issue logged to make `fixedTop` a directive that can be used across components.  I'll bump that one up on the priority list.  Bump for this component, we're good to remove that orphaned variable

andrew-frueh: This is the issue #654

PR #1145, opened 2020-01-30
1084 convert dialog

### LINKED ISSUE #1084: (Opt-in) Convert the Dialog

isaaclyman:
Fixes #1084.

I added an example that uses the `[touchUi]` Input field. It's the only thing in Cashmere that uses the dialog element.

### DIFF
--- a/projects/cashmere/src/lib/sass/dialog.scss
+++ b/projects/cashmere/src/lib/sass/dialog.scss
@@ -1,0 +1,64 @@
+$hc-dialog-padding: 24px !default;
+$hc-dialog-border-radius: 4px !default;
+$hc-dialog-max-height: 65vh !default;
+$hc-dialog-button-margin: 8px !default;
+
+@mixin hc-dialog-container() {
+    border-radius: $hc-dialog-border-radius;
+    box-sizing: border-box;
+    display: block;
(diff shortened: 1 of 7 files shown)


projects/cashmere/src/lib/sass/dialog.scss
```diff
@@ -0,0 +1,64 @@
+$hc-dialog-padding: 24px !default;
+$hc-dialog-border-radius: 4px !default;
+$hc-dialog-max-height: 65vh !default;
+$hc-dialog-button-margin: 8px !default;
+
+@mixin hc-dialog-container() {
+    border-radius: $hc-dialog-border-radius;
+    box-sizing: border-box;
+    display: block;
+    outline: solid 1px;
```
isaaclyman: We previously had two `outline` rules on this element. I kept the latter one. @benjanderson I seem to remember you were the last one to work on this, are you aware of any reason we might want to keep both?

andrew-frueh:
Well @isaaclyman - the good news is that doing this PR surfaced an issue that had flown under the radar.  The bad news is that we'll end up scrapping your work on this one because we'll yank the component out.  I'll submit a PR later today to remove the Dialog component.  Thanks!

PR #1151, opened 2020-01-31
feat(modal): tight parameter

### LINKED ISSUE #954: Allow banner (without padding) at top of modal body
**Describe your idea**

Much like modal header, body, and footer, it would be nice to be able to specify a banner that would display without the body padding.

A couple pictures say 1K words:

This is what you can do today out of the box:

image

This is what I would like to do:

image

Currently this is only possible by overriding Cashmere styles, which would be great to not have to do.

andrew-frueh:
adds parameter to remove default padding

closes #954

### DIFF
--- a/projects/cashmere-examples/src/lib/modal-overview/modal-overview-example-modal.component.html
+++ b/projects/cashmere-examples/src/lib/modal-overview/modal-overview-example-modal.component.html
@@ -1,6 +1,12 @@
 <hc-modal>
     <hc-modal-header>Modal Header Title</hc-modal-header>
-    <hc-modal-body>Data: {{ activeModal.data }}</hc-modal-body>
+    <hc-modal-body tight="true">
+        <hc-banner type="warning">
+            <span hcBannerStamp>WARNING</span>
+            You are about to export data which may include Protected Health Information (PHI)
+        </hc-banner>
+        <div class="example-content">Data: {{ activeModal.data }}</div>
+    </hc-modal-body>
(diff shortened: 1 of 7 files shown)


projects/cashmere-examples/src/lib/modal-overview/modal-overview-example-modal.component.html
```diff
@@ -1,6 +1,12 @@
 <hc-modal>
     <hc-modal-header>Modal Header Title</hc-modal-header>
-    <hc-modal-body>Data: {{ activeModal.data }}</hc-modal-body>
+    <hc-modal-body tight="true">
```
corykon: `<hc-modal-body tight="true">` will work because `"true" == true`, but technically this should probably be `<hc-modal-body [tight]="true">`

andrew-frueh: Ah that's a really interesting point.  In most cases I try to add the `parseBooleanAttribute` utility function to params that are setting boolean values.  But I hadn't considered that this is an example of why it's necessary.  I guess without that function, this might not evaluate correctly (if tight wasn't in brackets)?

andrew-frueh:
@corykon I updated this PR to follow the revised strategy we discussed in the other PR.  For the modal - I think the tight class should remove all padding in the content area its applied to, so this one is pretty straight forward.  Let me know how the change looks.

PR #1154, opened 2020-02-03
feat(multiple): tight parameter

### LINKED ISSUE #944: Add "tight" option for components (especially form controls)
**Describe your idea**

Trying to implement the Care Management app according to the mockups developed by Karri Quintana, our UX Designer--and especially our inline forms or "edit cards"--we find that the elements have a lot of built-in padding, large line-heights, etc., which make the forms much more spread out than we need. It would be nice to have an option like a "tight" directive we could include on the components or something similar that would allow us to achieve a tighter design without breaking existing apps that already use the controls. This option was discussed with Andrew Frueh several months ago and he seemed to be very supportive and enthusiastic for this idea. The current controls, with their included padding, etc. are a bit too "opinionated" for general use. They look fine on a large stand-alone form in a modal with lots of room, but look a bit too spread out when trying to use them in something like an inline edit card that appears within the context of a page displaying lots of other data. I would suggest working with Karri Quintana and Andrew Frueh on the details of spacing and sizing when implementing some type of "tight" option.
 
Many of the standard cashmere components (hc-tile, form controls, etc.) have too much built-in padding/margins/large line-height, etc. to make them general purpose enough for use throughout the application.

**Additional context**

andrew-frueh:
`tight` param removes default padding or margin on components

Lots of files affected here - but pretty much the same pattern repeating.  Tried to hit every component that had some kind of default margin or padding and give the option to omit that.

@jonest - for the `tight` param on table, I used it to apply the `hc-table-small` class that @corykon had already created in the stylesheet (but was never surfaced in the component).  I updated the second table example in this branch to display that.  Let me know if you think that trims the table enough for what your users are requesting.

closes #944

### DIFF
--- a/projects/cashmere/src/lib/form-field/hc-form-field.component.html
+++ b/projects/cashmere/src/lib/form-field/hc-form-field.component.html
@@ -1,4 +1,9 @@
-<div [class.hc-form-field-wrapper-inline]="inline" [class.hc-form-field-wrapper]="!inline" [class.hc-form-field-no-label]="!hasLabel">
+<div
+    [class.hc-form-field-wrapper-inline]="inline"
+    [class.hc-form-field-wrapper]="!inline"
+    [class.hc-form-field-no-label]="!hasLabel"
+    [class.hc-form-field-padding]="!tight"
+>
     <span *ngIf="hasLabel" [ngClass]="inline ? 'hc-form-field-label-wrapper-inline' : 'hc-form-field-label-wrapper'">
         <label [ngClass]="inline ? 'hc-form-field-label-inline' : 'hc-form-field-label'" [attr.for]="_control._componentId">
             <ng-content select="hc-label"></ng-content>
--- a/projects/cashmere-examples/src/lib/drawer-menu/drawer-menu-example.component.scss
+++ b/projects/cashmere-examples/src/lib/drawer-menu/drawer-menu-example.component.scss
@@ -1,24 +1,19 @@
 @import "~@healthcatalyst/cashmere/scss/colors";
 
 .drawer-content {
-    height: 200px; /* used to show menu scrolling effect*/
+    height: 200px; /* used to show menu scrolling effect */
 }
 
-.drawer-label {
-    margin: 0 50px 0 0;
-}
+/* positions a plus button to the right edge of a drawer item */
+.hc-drawer-item {
+    .drawer-label {
+        margin: 0 50px 0 0;
+    }
 
-.fa-plus {
-    color: $blue;
-    margin: 0 0 0 auto;
-}
-
-.hc-drawer-right .drawer-label {
-    margin: 0 0 0 50px;
-}
-
-.hc-drawer-right .fa-plus {
-    margin: 0 auto 0 0;
+    .fa-plus {
+        color: $blue;
+        margin: 0 0 0 auto;
+    }
 }
 
 .icon-left {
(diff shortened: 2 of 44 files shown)


projects/cashmere/src/lib/form-field/hc-form-field.component.html
```diff
@@ -1,4 +1,9 @@
-<div [class.hc-form-field-wrapper-inline]="inline" [class.hc-form-field-wrapper]="!inline" [class.hc-form-field-no-label]="!hasLabel">
+<div
+    [class.hc-form-field-wrapper-inline]="inline"
```
jonest: I'm actually not sure what this binding does. Mind explaining that to me?

jonest: I think it connects to that `@HostBinding` that sets a field based on whether the class is present and this is a way to then communicate that or something?

projects/cashmere/src/lib/form-field/hc-form-field.component.html
```diff
@@ -1,4 +1,9 @@
...
+    [class.hc-form-field-wrapper]="!inline"
+    [class.hc-form-field-no-label]="!hasLabel"
+    [class.hc-form-field-padding]="!tight"
```
andrew-frueh: Hey @jonest - line 5 here is actually the one to look at (the others were already there before).  Basically what I did is pull all the padding or margin values in components out into their own separate classes.  And a developer can now set `tight="true"` on any of these components to omit the default padding - ie. omit that padding class.  So what you've got here is binding that padding class to the inverse value of of the tight param.  In other words, if tight is true, then don't include the padding class.

jonest: OK got it, so it's like a global binding to change it everywhere (a step toward a "tight" UI across the board then?)

andrew-frueh: @jonest - that's mostly correct, I've added a `tight` parameter to every component that has default padding or margin.  But it's not global in the sense that you set a value once and it applies it everywhere.  It's still applied on an instance by instance basis.  So for example I could have `<table hc-table tight="true">` for one table that is using the compressed style, but another `<table hc-table>` that still has the default padding.

andrew-frueh:
You bring up a great point @corykon - that was a concern of mine too.  That if folks are using the styles on their own (for say, hc-tile) and not the component, then this will be a breaking change.  But this is why I approached it the way I did, let me know what you think:

It seems like the problem we're trying to fix here is that the default padding is more than some developers would like.  But it's not that they don't want *any* padding, it just that they want to be able to set their own.  So if I made it so when `tight` is true, it adds a class to override the default padding, then the developers is still going to have to add _another_ class to override the override - maybe having to use `!important`.  And it felt like at that point, `tight` isn't doing much more than overriding the defaults would have anyway.  At least with the approach proposed in this PR, it omits the padding/margins entirely - so the developer is just adding padding in, not overriding anything.

Thoughts on that?

corykon:
@andrew-frueh that's a good argument. Maybe we reach out and see whose using the mixins and validate that they're good with the change?

Other things to consider - Since devs are likely going to add in their own amount of padding after the fact when using `tight`, I would argue that writing styles to override the default padding without using `tight` versus writing styles with `tight` applied is really not very different... both take some work, just a minor difference in the specificity of selectors, maybe. Perhaps `tight` should really be a specifically thought out and recommended (lesser) amount of padding/margin/height/width.

andrew-frueh:
Yes - I like where you're going with that.  My mind was starting to go that direction too, that maybe tight shouldn't be entirely removing padding/margin, it should just be cutting it to the minimum amount that we feel is still acceptable (very much like what you did on `.hc-table-small`).  That's what I have `tight` doing on the table component - is applying that class.  

Let's use our Cashmere programming hour tomorrow to adjust this PR with that strategy.

PR #1160, opened 2020-02-07
Table tight

### LINKED ISSUE #944: Add "tight" option for components (especially form controls)

andrew-frueh:
Adds a tight parameter to the following:

**Table**: applies the hc-table-small class to the table component

**Tabs**: compresses the padding on vertical or horizontal tabs

**Tile**: reduces the padding to 10px 15px

**FormField**: for any components that leverage HcFormField (checkbox, radio, select, input), the tight param compresses the padding and margin on the element.  A new directive was added, `hc-form` which can be added to a `<form>` and the tight param set.  If so, it will apply tight styling to all contained hc-form-fields.  A new example was added to FormField to demonstrate the difference between a tight form and default one.  The tight version saves about 120px vertically (see screenshot).

Screen Shot 2020-02-07 at 10 15 26 AM

closes #944

### DIFF
--- a/projects/cashmere-examples/src/lib/checkbox-standard/checkbox-standard-example.component.html
+++ b/projects/cashmere-examples/src/lib/checkbox-standard/checkbox-standard-example.component.html
@@ -1,5 +1,7 @@
-<hc-checkbox id="light">Standard Checkbox</hc-checkbox>
+<hc-checkbox>Standard Checkbox</hc-checkbox>
 
-<hc-checkbox id="disabled" disabled="true">Disabled Checkbox</hc-checkbox>
+<hc-checkbox disabled="true">Disabled Checkbox</hc-checkbox>
 
-<hc-checkbox id="disabled" [checked]="true" disabled="true">Disabled Checkbox (Checked)</hc-checkbox>
+<hc-checkbox [checked]="true" disabled="true">Disabled Checkbox (Checked)</hc-checkbox>
+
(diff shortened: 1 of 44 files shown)


andrew-frueh:
Great suggestion @corykon.  As @isaaclyman has said - if there isn't an example for it, in the mind of most developers it doesn't exist.  I've updated all components with this parameter to have at least one example demonstrating the parameter.  I didn't want to add new examples just for this though, so in all cases I just modified existing examples slightly.

PR #1165, opened 2020-02-13
feat(stepper): hc-stepper component

### LINKED ISSUE #1131: Progress Stepper Component
**Describe your idea**

The CORUS Budgeting team has a need for a progress stepper to show users where they are in a process.  We had a style target for this in an older version of the style guide, but it's never been turned into a component.

Screen Shot 2020-01-23 at 12 25 26 PM

andrew-frueh:
adds a new progress stepper component

closes #1131


projects/cashmere/src/lib/sass/stepper.scss
```diff
@@ -0,0 +1,311 @@
...
+        &.hc-step-active {
+            background-color: $active;
+
+            &:not(.hc-step-last) {
+                .hc-step-arrow {
+                    @include step-arrow($track-bg, null);
+                }
+            }
+
+            .hc-step-disk {
```
isaaclyman: A lot of this CSS isn't up to standard, per step 6 of https://cashmere.healthcatalyst.net/guides/using-customizing-components.

Quoted:

- The mixin stylesheet should not include any nested CSS selectors that would make confusing assumptions about the user's HTML structure. That is, the [disabled] attribute selector and pseudo-selectors like :hover or :after are okay, but class or tag names like .hc-checked and input should be avoided.
- Both stylesheets should be as flat as possible. Nesting CSS selectors in SCSS files creates extremely specific selectors, which are hard to override. Cashmere should make it easy to override a style rule if needed. CSS classes can be added to component markup to avoid deep nesting.

projects/cashmere/src/lib/stepper/stepper.component.scss
```diff
@@ -0,0 +1,102 @@
...
+    .hc-stepper-arrow {
+        @include arrow-theme($blue, lighten(tint($blue, 20%), 3%), $slate-gray-100, $slate-gray-200, $white, $white);
+    }
+    .hc-stepper-isolated {
+        @include isolated-theme($blue, $slate-gray-200, $white, $text, $isolated-disk-size);
+    }
+}
+
+.hc-stepper-purple {
+    .hc-stepper-arrow {
```
isaaclyman: The pattern I've been following so far is:

```scss
.hc-stepper-arrow {
...
  .hc-stepper-purple & { ... }
  .hc-stepper-blue & { ... }
...
}
```

Since `.hc-stepper-arrow` is the element being styled, the above pattern keeps all of its potential style variations in one place, making it easier to read IMO.

andrew-frueh: Yup - I agree that would better.

projects/cashmere/src/lib/stepper/stepper.component.ts
```diff
@@ -0,0 +1,135 @@
...
+    }
+    private _useRouterOutlet: boolean = true;
+
+    /** Get or set the currently selected zero-based index of the stepper */
+    @Input()
+    get activeIndex(): number {
+        return this._activeIndex;
+    }
+    set activeIndex(value: number) {
+        if (this.steps) {
```
isaaclyman: A nitpick, but let's invert this `if` statement with an early return.

projects/cashmere/src/lib/stepper/stepper.component.ts
```diff
@@ -0,0 +1,135 @@
...
+            if (value >= 0 && value < this.steps.length) {
```
isaaclyman: This one could be inverted as well, since the `else` statement is a `throw`.

projects/cashmere/src/lib/stepper/stepper.component.ts
andrew-frueh: Got it - so like this?
```
set activeIndex(value: number) {
        if (!this.steps) {
            return;
        } else if (value < 0 || value >= this.steps.length) {
            throw Error('The hc-stepper activeIndex value of ' + value + ' is out of bounds');
        } else {
            if (this._routerEnabled) {
                this.router.navigate([this.steps[value].routerLink]);
            } else {
                this._activeIndex = value;
                this.activeIndexChange.emit(this._activeIndex);
            }
        }
    }
```

isaaclyman: Almost. More like this:

```javascript
if (!this.steps) {
  return;
}

if (value < 0 || value >= this.steps.length) {
  throw Error(`...`);
}

if (this._routerEnabled) {
  this.router.navigate([this.steps[value].routerLink]);
} else {
  this._activeIndex = value;
  this.activeIndexChange.emit(this._activeIndex);
}
```

The goal is to reduce logical nesting and therefore cyclomatic complexity.

andrew-frueh: Fascinating.  I confess that's the first time I've heard the term "cyclomatic complexity".  But I learned something new today :-) thanks Issac

isaaclyman: ha, no problem. It's probably the most pretentious term I know

projects/cashmere/src/lib/stepper/stepper.component.scss
andrew-frueh: Fixed in the most recent commit.

projects/cashmere/src/lib/stepper/stepper.component.ts
andrew-frueh: Haha - it's good one! Fixed in the most recent commit.

projects/cashmere/src/lib/sass/stepper.scss
andrew-frueh: @isaaclyman I've got all the base classes following the standard (I think). But the big exception are the mixins to theme the arrow or isolated types: `arrow-theme()` and `isolated-theme()`.  Because the point of those functions is to override the base classes with specific color values - I'm really struggling to see how to do that without nested selectors.  

Part of the problem is that within those those mixins - the trailing & operator doesn't work.  Because the browser assumes the parent to be the class that called the mixin - not the class in the mixin.  So the hierarchy gets pretty wonky.

I could flatten everything out to not be nested but use > selectors - but it seems like that's just semantics, and it would actually make the hierarchy assumptions even worse. 

Open to suggestions...

andrew-frueh: I take it back - it was doable, just required a good bit of rework.  It makes the css a little crazy, but I think now everything is as flat as it can be.  Give it a look and let me know what you think.

projects/cashmere/src/lib/sass/stepper.scss
```diff
@@ -0,0 +1,185 @@
...
+    display: block;
+    height: 0px;
+    position: absolute;
+    right: 0;
+    top: 0;
+    z-index: 2;
+}
+
+@mixin hc-stepper-bg-color($color) {
+    background-color: $color;
```
isaaclyman: Getting philosophical for a moment. I have to wonder, is this a useful mixin? If the purpose of the mixins is to allow people to apply Cashmere styles without using the full Angular component, then a mixin like this doesn't help much--if they want a "green" stepper, for example, this still requires them to dig around in source code to find the appropriate color variable. I'll approve for now, but maybe this is something we should mull over.

projects/cashmere/src/lib/stepper/stepper.component.scss
```diff
@@ -0,0 +1,324 @@
...
+    .hc-stepper-red .hc-stepper-arrow .hc-step-completed & {
+        @include hc-disk-border($red-complete);
+    }
+}
+
+.hc-step-arrow {
+    .hc-stepper-arrow .hc-step:not(.hc-step-last) & {
+        @include hc-step-arrow($track-bg, null);
+    }
+    .hc-stepper-green .hc-stepper-arrow .hc-step-completed & {
```
isaaclyman: We're getting some pretty high specificity here. Can some of these intermediate selectors be omitted? For example, in `.hc-stepper-green .hc-stepper-arrow .hc-step-completed .hc-step-arrow`, could we safely remove the `.hc-stepper-arrow`? Maybe we could further condense things by adding a conditional class e.g. `hc-step-arrow-completed` in the markup so we wouldn't need the `.hc-step-completed` class. I think the extra effort is worth it to make Cashmere styles as overrideable as possible.

projects/cashmere/src/lib/sass/stepper.scss
andrew-frueh: Yep - I totally hear what you're saying.  And after this sprint, I'd love for you to give this one a look over.  Because I think I've followed the letter of the law on this mixin, but not really the spirit of it.  My brain was hurting last night just getting it this far though.  But now that it's at least up to the minimum standards - I agree we could do better.

projects/cashmere/src/lib/stepper/stepper.component.scss
isaaclyman: In fact, throughout this file I think it would be good to achieve a specificity goal of <3.

andrew-frueh: yep - `hc-step-arrow-completed` could work

andrew-frueh: @isaaclyman I've pushed a commit the implements this suggestion

isaaclyman:
@andrew-frueh looks great - thanks for making those refactors.

andrew-frueh:
Awesome - thanks for all your input and the multiple reviews @isaaclyman.   The css work has definitely been education on this one.

PR #1190, opened 2020-03-05
fix(select): support for optgroup

### LINKED ISSUE #1188: hc-select doesn't support optgroup elements.
**Summary**

Trying to use optgroup elements to group options within an hc-select and it doesn't appear to work - just renders an empty list.

**Reproduction**

https://stackblitz.com/edit/cashmere-optgroup-select?file=src/app/select-standard/select-standard-example.component.html

**Additional context**

Cashmere v6.7.1

andrew-frueh:
allow hc-select to support optgroup markup

closes #1188


andrew-frueh:
@corykon as best as I can tell that would require multiple `ng-content` tags with different selects.  But I'm no authority so maybe there's another way.  In any case, it seemed like there wasn't a whole lot of need for the filter anyway - since folks should assume their using the hc-select tag like they would a DOM select tag.

PR #1200, opened 2020-03-16
fix(hc-stepper): stepper bug fix and enhancements

iamboo:
ActiveIndex can now be null
Added literal types for color and type options


projects/cashmere-examples/src/lib/stepper-overview/stepper-overview-example.component.ts
```diff
@@ -29,4 +28,11 @@ export class StepperOverviewExampleComponent implements OnInit {
             {label: 'Year Round', iconSet: 'fa', icon: 'fa-lock', disabled: true}
         ];
     }
+
+    ngAfterViewInit() {
```
andrew-frueh: This is interesting. From a code style perspective, most of the other Cashmere examples use parameters on the component tag, but I don't think either way is necessarily right or wrong. I'm assuming this change was just a refactor - or was their a functionality reason to define these params after view init?

projects/cashmere/src/lib/stepper/stepper.component.ts
```diff
@@ -80,38 +83,39 @@ export class StepperComponent implements AfterContentInit {
         if (!this.steps) {
             return;
         }
-        if (value < 0 || value >= this.steps.length) {
-            throw Error('The hc-stepper activeIndex value of ' + value + ' is out of bounds');
-        }
-        if (this._routerEnabled) {
+        // if (value !== undefined && (value < 0 || value >= this.steps.length)) {
```
andrew-frueh: Can we get this check back in?  There is one unit test failing because this isn't included.  Was there a reason to remove it?

projects/cashmere-examples/src/lib/stepper-overview/stepper-overview-example.component.ts
iamboo: There are many times when a variable isn't needed for setting these properties and there is no need to make a property public. It works the same either way.  This is a refactor that I wanted to validate by using the routes to set the active step instead of a click or variable.

projects/cashmere/src/lib/stepper/stepper.component.ts
iamboo: One of the reasons for these changes was to allow the activeIndex to be "null".  I am using -1 as the vehicle to not set an activeIndex. I did go down the path of using the type of `number | undefined` to manage the "null" feature. Should I switch over to that option and restore this `out of bounds` error?

projects/cashmere/src/lib/stepper/stepper.component.ts
```diff
@@ -80,38 +83,39 @@ export class StepperComponent implements AfterContentInit {
...
-    constructor(private router: Router) {}
-
-    ngAfterContentInit() {
-        this._checkForRouterUse();
+    constructor(private router: Router) {
         this.router.events.forEach(event => {
             if (event instanceof NavigationEnd) {
                 const url = event && event.url ? event.url : '';
                 this._findCurrentStep(url);
-                this.activeIndexChange.emit(this._activeIndex);
```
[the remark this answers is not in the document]
iamboo: I think it should. I didn't test that. Is it not triggering on a route change?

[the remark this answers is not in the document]
andrew-frueh: Oh - well if we want the emitter to fire on route change, than we probably need to leave line 109 in (looks like it was removed).  Because that should emit any time the router routes.

projects/cashmere-examples/src/lib/stepper-overview/stepper-overview-example.component.ts
andrew-frueh: Gotcha - that's fine by me

projects/cashmere/src/lib/stepper/stepper.component.ts
andrew-frueh: Ah okay, I thought it might be something like that.  Yes, if we could go that route and restore the error, I think that just makes life a little easier for our users to debug.  And just make sure the unit test is passing again (you can adjust the test if needed)

projects/cashmere/src/lib/stepper/stepper.component.ts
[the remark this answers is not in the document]
andrew-frueh: Although I just realized for this to work, `_findCurrentStep()` would have to be allowed to work when activeIndex is anything (not just -1).  Because otherwise line 109 will just emit whatever the last activeIndex was (not what the index is for the new route)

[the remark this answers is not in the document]
iamboo: This is now working correctly with route changes.

projects/cashmere/src/lib/stepper/stepper.component.ts
iamboo: Error is now restored.

PR #1212, opened 2020-03-26
convert scroll nav to mixins

### LINKED ISSUE #1086: (Opt-in) Convert the Scroll Nav

isaaclyman:
Fixes #1086

### DIFF
--- a/projects/cashmere/src/lib/sass/scroll-nav.scss
+++ b/projects/cashmere/src/lib/sass/scroll-nav.scss
@@ -34,3 +34,14 @@
         font-weight: 500;
     }
 }
+
+@mixin hc-scroll-nav-content-container() {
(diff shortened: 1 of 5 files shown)


projects/cashmere/src/lib/sass/scroll-nav.scss
```diff
@@ -0,0 +1,47 @@
+@import '../../sass/colors';
+@import '../../sass/mixins';
```
andrew-frueh, line 2: I think this is what's causing the build error.  I think it should be:
`@import './colors';
@import './mixins';`

andrew-frueh: went ahead and pushed an update to this branch to resolve the error

PR #1222, opened 2020-04-08
refactor(scss): remove view encapsulation

### LINKED ISSUE #941: (Opt-in) Add ViewEncapsulation.None to all components
Some components are view-encapsulated, so someone trying to use the component via CSS classes will be unable to do so. The pagination component is one example.

In order to allow level 2 of customizable components, all Cashmere components should be marked with ViewEncapsulation.None.

isaaclyman:
Add ViewEncapsulation.None to all components and import all component scss files to cashmere.scss.

I tested each component to ensure that behavior wasn't changed, but it's possible that some combination of loading one component and then viewing another could cause CSS bugs, despite our best efforts to use well-scoped class names. So we'll have to keep an eye out for those.

Fixes #941


tslint.json
```diff
@@ -69,7 +69,7 @@
         "component-selector": [true, "element", "hc", "kebab-case"],
         "use-input-property-decorator": true,
         "use-output-property-decorator": true,
-        "use-host-property-decorator": true,
+        "use-host-property-decorator": false,
```
isaaclyman, line 72: While I understand the reasoning for preferring `@HostBinding` in a regular front-facing project, in our case where we're using unencapsulated styles and `:host` isn't available to us, I believe the convenience of the `host` property in the Component annotation outweighs the downsides for adding static host classes.

corykon:
A thought - when this gets released, we'll want to be sure that we encourage people to double check styles throughout their apps. If they happen to be reusing some of our class names, possible there could be some unintended collisions.

isaaclyman:
Good call out - hopefully nobody's using class names prefixed with `hc-`, but it's not impossible.

PR #1234, opened 2020-04-23
build(scss): build standalone css file

### LINKED ISSUE #1100: Make all Cashmere component/shared styles accessible through a built CSS file
It should be possible to use Cashmere styles without a SCSS preprocessor or build toolchain. Ideally, the NPM package would include a single raw CSS file that could be linked in HTML, which would contain all shared and component styles.

Once all components are converted to mixins, let's work on setting this up in our build toolchain.

isaaclyman:
This adds onto the postbuild step to create dist/cashmere/cashmere.css, a vanilla CSS file that contains all Cashmere styles. It's ~7000 lines long, if anyone was wondering.

Fixes #1100


corykon:
Any thoughts on creating a guide for using this new flat css file?

isaaclyman:
Yes, see the remaining "To do" tasks in https://github.com/HealthCatalyst/Fabric.Cashmere/projects/5

PR #1235, opened 2020-04-24
fix(formfield): more flexible label content

### LINKED ISSUE #1225: Ability to add help text to form controls.
Please add the ability to allow help/direction text between the label and various form control types.  It would be nice if the text could be styled based on specific implementation (or at least be sized smaller, less pronounced than the label).  The current use case mainly applies to vertical label/control orientation.  Example provided below...

image

andrew-frueh:
Allow additional custom markup inside hc-form-fields.

I think this change makes form-field more dynamic.  Beyond adding the help subtext, a user could add an image, omit hc-label and make their own custom label, etc.

So Jayson - your markup will look something like this for any form-field (with a class on the span to style the help text accordingly):
```
<hc-form-field>
            <hc-label>Run frequency:</hc-label>
            <span>Additional instructions may be included on this line</span>
            <hc-select placeholder="Select frequency:" [formControl]="selectControl">
                <option value="daily">Daily</option>
                <option value="weekly">Weekly</option>
                <option value="monthly">Monthly</option>
            </hc-select>
            <hc-error>Please select the run frequency for this item</hc-error>
</hc-form-field>
```
Let me know if you have any concerns.

closes #1225

### DIFF
--- a/projects/cashmere/src/lib/form-field/hc-form-field.component.html
+++ b/projects/cashmere/src/lib/form-field/hc-form-field.component.html
@@ -11,7 +11,7 @@
     </span>
     <div class="hc-form-field-content-wrapper">
 
-        <ng-content></ng-content>
+        <ng-content select="hc-label-ext"></ng-content>
 
         <div
             *ngIf="hasInput"
@@ -37,9 +37,7 @@
             [class.hc-form-field-non-input-inline-tight]="inline && tight"
             [class.hc-form-field-invalid]="_shouldShowErrorMessages()"
         >
-            <ng-content select="hc-select"></ng-content>
-            <ng-content select="hc-checkbox"></ng-content>
-            <ng-content select="hc-radio-group"></ng-content>
+            <ng-content></ng-content>
         </div>
         <div class="hc-form-field-error-wrapper" *ngIf="_shouldShowErrorMessages()">
             <hc-error *ngIf="_control._errorMessage && _control._errorMessage.length">{{ _control._errorMessage }}</hc-error>
--- a/projects/cashmere/src/lib/form-field/hc-form-field.component.html
+++ b/projects/cashmere/src/lib/form-field/hc-form-field.component.html
@@ -10,6 +10,9 @@
         </label>
     </span>
     <div class="hc-form-field-content-wrapper">
+
+        <ng-content></ng-content>
+
         <div
             *ngIf="hasInput"
             [class.hc-form-field-flex]="!inline"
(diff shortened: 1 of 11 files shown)


jayols:
In the case of adding an icon or image, would that be done within the span tags or as separate elements between hc-label and hc-select (or some other form control)?

projects/cashmere/src/lib/form-field/hc-form-field.component.html
```diff
@@ -34,7 +37,9 @@
             [class.hc-form-field-non-input-inline-tight]="inline && tight"
             [class.hc-form-field-invalid]="_shouldShowErrorMessages()"
         >
-            <ng-content></ng-content>
+            <ng-content select="hc-select"></ng-content>
```
corykon, line 40: I'm not super familiar with the innards of hc-form-field or using selective transclusion stuff, so I'm curious why we needed to specifically have have specific ng-content tags for select, checkbox, and radio groups, but not input, textareas, other for controls

andrew-frueh:
Hey @jayols - to your question, it's the latter. Basically you could add as much additional markup as you want inside `<hc-form-field>`, but you wouldn't want to put it inside `<hc-label>`.  hc-label should just be for the main instruction (if you're using it), because it'll get the "required" red asterisk added.  That aside though, any additional markup you add should be supported, and it'll be displayed before the the form element in the DOM.

jayols:
Ok, thanks.  In a prior conversation I recall you asking if the "extra" content should be enclosed in its own tags.  I might suggest we consider that if there is any concern around devs getting confused as to where this flexibility is allowed.  Just a thought...

andrew-frueh:
I see what you're saying - just allowing any additional content might be confusing.  Any thoughts on what the best naming would be for that tag?  `<hc-form-field-content>`,  `<hc-label-secondary>`, `<hc-subtext>`?

jayols:
Yeah, it seems like it is secondary to the label - my current use case being help text.  Would "aux" or "ext" be useful (e.g. hc-label-ext or hc-label-aux)?

I could see the layout as follows:
<hc-label>some label</hc-label>
<hc-label-ext>some content here</hc-label-ext>
<input ... />
<hc-error>

andrew-frueh:
Cool - yeah that works.  @corykon mentioned that he might prefer going without the additional tag to just keep the markup a little leaner.  I'm 50/50 - so I'm gonna put it out to the steering committee and take a vote.  Either way, I'll be sure and include another example on the form-field page illustrating the functionality for whichever approach we go with.

andrew-frueh:
@corykon @jayols - looks like the consensus is to go with `hc-label-ext` as the container for the additional content.  Just pushed an update that includes that markup as well as a new example to demonstrate.  Looks like this:

Screen Shot 2020-04-27 at 3 29 01 PM

I included a default text style on the container so we have a base standard for how secondary text should be displayed.  I believe it pretty closely matches to the mockup you were working against, Jayson.

Let me know if either of you have any questions or concerns before I merge this.

jayols:
Just to confirm - if no hc-label-ext container exists, it will look like the standard label/form item styling (i.e. no extra space, padding, etc.)?

If yes, then I think we're good!

andrew-frueh:
Yep - you got it. Totally optional and form-field is unchanged if it’s omitted.  Thanks for the review!

Issue #951, opened 2019-08-15
Radio Button Enhancements


corykon:
Ran into a couple things while implementing cashmere radio buttons in popbuilder.

**Describe your idea**
1. Can we allow `(click)` handlers on hc-radio-button?
2. Can we have a focus outline when tabbing through radio groups?
3. Lets not use `!important` in _radio-group.scss

**What problem are you trying to solve?**
1. We used to have click handlers for running some additional logic when clicking a radio button, aside from manipulating a model. We should be able to use a setter method on the model to achieve the same thing, but I'm not sure why the hc-radio-button needs stop propagation and disallow the use of a `(click)` handler 

2. I can tab through cashmere radio groups and edit the selection using the up and down arrows, but it's hard to tell that I'm focused on the radio group.

3. Lets not force implementers to use `!important` if they want to override something.

![radio-button-tabbing](https://user-images.githubusercontent.com/12416432/63115700-fdef7e80-bf54-11e9-837a-bfac9ba74ce6.gif)

corykon:
NOTE: it looks like cashmere does have a `(change)` output that could pretty similarly to a `(click)` output. I missed that in the API section

PR #1242, opened 2020-04-30
fix(hc-radio): add focus styles

corykon:
A couple items in the issue were already addressed, so this just enhances keyboard accessibility by adding simple focus. Basically just makes it work more like native radios.

cashmere-radio
native-radio

re #951


andrew-frueh:
Question @corykon - for the tab (focus) state, would it be better to mimic the hover styling where the circle border turns blue?  Rather than putting a box around it?

corykon:
@andrew-frueh Only issue there is that you couldn't tell the radio has focus if it were already selected (it usually is) because the border is already blue. Totally open to suggestions though. Native browser stylings aren't awe-inspiring.

PR #1248, opened 2020-05-08
Ng 8 upgrade

mhollier:


corykon:
Sweet! So glad this is going in. Nice work @mhollier 

Do we have any idea how this might impact cashmere implementers who are on a version of angular before 8? Do we have anyone at Catalyst who's not there yet?

mhollier:
@corykon I don't know the impact of apps on earlier versions of Angular. I know that Atlas and Operations Console are on Angular 8. What other apps are using Cashmere?

corykon:
@mhollier I want to say most everyone uses Angular now. Last I heard Care Management was still on v6, not sure about Leading Wisely

joeskeen:
We may have some old Medicity and Able Health applications not on Angular, but there are a lot of web apps outside of DOS Platform that are using Cashmere.
I do know that Care Management is using at least Angular 7, but I think they are on 8. My vibe from talking with all teams last year is that those who weren't on 8 were planning to move there. This change is what most teams were waiting for.

Sent from Outlook Mobile<https://aka.ms/blhgte>

________________________________

PR #1272, opened 2020-06-09
Fix unwanted horizontal scrolling in IE due to larger drawer container width

pheebner:
I made modifications to the Cashmere drawers so that in the "closed" state the drawers are actually inside the drawer containers, just hidden so that the user can't see them. The open animation sets the drawer outside the container just before the animation starts. The close animation puts the drawer back inside the container after it has finished the close. This allows the drawer containers to have the same width as they would have normally if there was no drawer being used. The only time the width of the container is modified is when the drawer is being animated in or out.

The reason for this change is that we encountered a bug in internet explorer that causes unwanted horizontal scrolling if a Cashmere Scroll Nav is used inside a Cashmere Drawer. The Element.scrollIntoView() method in internet explorer seems to think that the extra width created by the drawer is fair game for scrolling horizontally.

I have made the same modifications to the menu-drawer.

### DIFF
--- a/projects/cashmere/src/lib/drawer/drawer.component.ts
+++ b/projects/cashmere/src/lib/drawer/drawer.component.ts
@@ -36,6 +36,8 @@ export function validateAlignInput(inputStr: string) {
     }
 }
 
+const openStateAnimation = '400ms cubic-bezier(0.25, 0.8, 0.25, 1)';
+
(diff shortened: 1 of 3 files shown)


pheebner:
You're welcome @andrew-frueh and thank you!

PR #1278, opened 2020-06-12
docs(launch screen): created launch screen styles guide

### LINKED ISSUE #922: How to Implement Launch Screen
**Where does our documentation fall short?**
How do I implement the launch screen?

_Tell us which Cashmere documentation page this request is about and what's missing or unclear. The rest of this form is optional._
Teams have just heisted the code from the cashmere index page. Wish should provide better instructions for how to do this, since its a little trickier when you're operating pre-angular bootstrap.

**What would you add or change?**
Prob add a new page under styles called something like "Application Launch Screen"

corykon:
closes #922


joeskeen:
@corykon @andrew-frueh The launch screen has a fractional-second blip where there is a white bar on the left side of the screen before the CSS files load over HTTP.  We fixed this in Atlas and IDEA by adding a `body { margin: 0; padding: 0; }` to the styles inside of the app root element.  We should add this to the Cashmere site and this new guide as well.

corykon:
Great catch. I’ll do that.

PR #1279, opened 2020-06-13
docs(navbar): add report an issue to help menu

### LINKED ISSUE #1263: "Report an Issue" in the standard help menu
Suggestion to add “Report an Issue” to the standard help menu. Based on internal discussion among PMs and users.

andrew-frueh:
Updates recommended spec for standard help menu.

Also updated the "last modified" date for all our style pages while I was in there.  I know those won't always be perfect because they require a manual update and it's easy to forget - but I think it's worth it.  @corykon - actually I think you might need to add that to your Launch Screen style page.

closes #1263

### DIFF
--- a/guides/styles/about-modal.md
+++ b/guides/styles/about-modal.md
@@ -1,6 +1,6 @@
 # App "About" Modal
 
-###### Last updated June 28, 2019
+###### Last updated Dec 10, 2019
 
 :::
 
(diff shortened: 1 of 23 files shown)


andrew-frueh:
Hah - I was 50/50 on the exclamation vs the bug.  I figured a bug might be too "gross" ;-)  But I'm good with switching to that if you thinking it's clearer.

corykon:
Haha I figured you must have considered it. Totally good either way, but at
least to me it's clearer, and has a little more personality.

PR #1286, opened 2020-06-16
fix(date range): smoother ui

### LINKED ISSUE #1244: Calendar Date Picker locks me from picking the next month of data
I ran a report for March then wanted to switch it to April but the controls blocked me. This needs to throw in an error in red at Apply time rather than locking the calendar browsing arrows.

image

corykon:
warn about date range being invalid on save, instead of locking the ui while selecting dates

fix #1244


andrew-frueh:
Out of curiosity, will this fix resolve the issue #1287 that @jonest just submitted?

corykon:
I’ve haven’t dug in enough to know, but my guess would be no...

corykon:
@jonest  @andrew-frueh looks like it should fix #1288 though!

PR #1302, opened 2020-06-26
Pagination Fixes/Changes

pheebner:
- Don't sanitize page number until component has been initialized
- Don't set page number to 0 if length input is 0, set it to 1
- Don't emit events before initialization has occurred.


projects/cashmere/src/lib/pagination/load-more-pagination.component.spec.ts
```diff
@@ -49,10 +41,9 @@ describe('PaginationComponent', () => {
 
     describe('when _loadNextPage is called on the first page', () => {
         beforeEach(() => {
-            fixture.detectChanges();
-            component.pageNumber = 1;
             component.length = 100;
             component.pageSize = 10;
+            component.pageNumber = 1;
```
corykon, line 46: If we were to add a call to `_sanitizePageNumber()` in the setter for `length`, would this test pass regardless of the order we set the properties?

pheebner: Hmm I suppose it wouldn't pass. I guess I was thinking that once the component is initialized it would be reasonable to require the page number to be set after the other fields. Now that I think about it though, I guess there are cases where the developer doesn't have control over the order in which the inputs are changed.

I just had a thought, what if we were to try to sanitize the page number in `ngOnChanges()`? Could that be a solution? I'm thinking maybe at that point all of the inputs have been changed and we can sanitize confidently.

corykon:
Hmm yeah could try that. Maybe see if the onChanges call was triggered as a result of one of those 3 inputs updating, then call the sanitizer if it was. I’m not as familiar with when in the component lifecycle ngOnChanges is called though so you’d have do some investigating, perhaps still have your check in there to make sure things have been initialized

projects/cashmere/src/lib/pagination/load-more-pagination.component.spec.ts
pheebner: @corykon messed around with this a little more. The ngOnChanges call only seems to happen when the bound values (from the parent component) are changed as opposed to every time the component values are changed. I decided to stick with the original setTimeout approach here, except that instead of reusing the sanitized page number I'm recalculating it after the timeout to account for any other changed inputs. Does this seem good to you?

corykon:
@pheebner I would love it if we could avoid the setTimeout. I may need to dig in more to really understand the underlying issues, but I'm wondering if we may be able to simplify things. I'd like to try having a call to the `_sanitizePageNumber()` in each of the three setters in this file, and then leverage `initialized` observable inside of the `Initializable` class to make sure we're not sanitizing too soon.

Would you be open to pairing on this sometime this week?

pheebner:
@corykon Yeah I'd be down to pair later this week! I'm out tomorrow but will be back Wednesday and could do anytime after 12 PM MT.

pheebner:
Whoops, I screwed up that rebase...going to make a new pull request

PR #1317, opened 2020-07-03
feat(button): additional colors

### LINKED ISSUE #1252: More button colors
In complex interfaces, we sometimes feel there is a scarcity of color options for Cashmere buttons. The [Cashmere colors guide](https://cashmere.healthcatalyst.net/styles/color) has a primary and secondary palette, but these colors are not easily applied to buttons (there are hover states, active states, split buttons, etc.).

Given the need to balance concerns of unified design across HC products (avoiding a rainbow of button colors), but still wanting at least a few more color options, some options might be:

1. Include an Input parameter on Cashmere buttons that allows a secondary palette color or a custom color to be used, then applies the proper shading/tinting for hover and active states
2. Add a couple more buttonStyle options

andrew-frueh:
hc-button support for all colors in the primary & secondary palettes

@isaaclyman @corykon let me know what you think about this approach.  It allows for additional colors, but only those in the primary/secondary palettes, so it won't be a complete free for all.

closes #1252


projects/cashmere/src/lib/button/button.component.scss
```diff
@@ -39,6 +40,64 @@
...
+    &.hc-button-purple-gray {
+        @include hc-button-custom( $purple-gray );
+    }
+    &.hc-button-yellow {
+        @include hc-button-custom( $yellow );
+    }
+    &.hc-button-yellow-orange {
+        @include hc-button-custom( $yellow-orange );
+    }
+    &.hc-button-tan {
```
isaaclyman, line 98: Given the programmatic capabilities of SCSS, there's probably a way to do all of these in a loop and save some space. This works though.

Issue #1299, opened 2020-06-24
PaginationComponent Initialization issues


pheebner:
The Ops Console team has been experiencing a few issues with the PaginationComponent, and it all seems to have to do with its initialization. Below are some examples of what we have experienced. I have looked and debugged through the PaginationComponent code and I've also included my best guesses as to why these issues are occurring:

1. **Page number changed from what we want** - We want to display 18 jobs with a page size of 10, which should show 2 pages. I want to initialize the component to page 2, but when I try to do this the PaginationComponent switches it back to page 1. I believe this is because the order in which the inputs are set is not controlled, so the page number is being set before the page size. This means the component will have the default page size of 20 at the time, which will cause the component to change the page to 1 (since all 18 jobs can be shown on one page with a page size of 20). Then when the page size is changed to 10 (what we want) the page number stays at 1.

2. **Page number changed to 0** - Sometimes when we are still loading data the length of that data we pass into the pagination component is 0 until it has loaded. We want to have the page number set to 1 for when the data is loaded, but it gets set to 0, so nothing is highlighted once the data is loaded. I don't think the intent would be to set it to zero, but correct me if I'm wrong. I think this occurs because when the PaginationComponent code sanitizes the page number, it sets it to the total amount of data if the page number is larger. Should this be the behavior or is this expected?

andrew-frueh:
closed in #1321

PR #1321, opened 2020-07-03
fix(pagination): check all input after timeout

pheebner:
Sanitize page number and emit events after timeout.
This will prevent the page number being sanitized
before all inputs are set.

Issue #1299

~~BREAKING CHANGE: No longer emitting events during initialization.~~

### DIFF
--- a/projects/cashmere/src/lib/pagination/load-more-pagination.component.spec.ts
+++ b/projects/cashmere/src/lib/pagination/load-more-pagination.component.spec.ts
@@ -2,7 +2,7 @@ import {ComponentFixture, TestBed} from '@angular/core/testing';
 import {ButtonModule} from '../button';
 import {LoadMorePaginationComponent} from './load-more-pagination.component';
 
-describe('PaginationComponent', () => {
+describe('LoadMorePaginationComponent', () => {
     let fixture: ComponentFixture<LoadMorePaginationComponent>;
     let component: LoadMorePaginationComponent;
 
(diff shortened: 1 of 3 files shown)


joeskeen:
I would personally change the commit message to not make it a breaking change.  I have some ideas around making Pagination better overall, and I'd hate to do back-to-back breaking changes.  I think the delay is fine as it fixes a problem, and I doubt it will break anyone.  @andrew-frueh @corykon thoughts?

corykon:
@joeskeen I'm good with that. I agree we probably need some bigger architecture changes on this thing to get it perfect, but PJ's fix is a good stopgap. So maybe we just have @pheebner redo the commit one more time, and just skip the breaking change tag?

pheebner:
@joeskeen and @corykon just amended the commit message to take out the breaking change tag as requested

projects/cashmere/src/lib/pagination/load-more-pagination.component.spec.ts
```diff
@@ -2,7 +2,7 @@ import {ComponentFixture, TestBed} from '@angular/core/testing';
 import {ButtonModule} from '../button';
 import {LoadMorePaginationComponent} from './load-more-pagination.component';
 
-describe('PaginationComponent', () => {
+describe('LoadMorePaginationComponent', () => {
```
joeskeen, line 5: Good catch!  And thanks for looking over the tests!

PR #1324, opened 2020-07-06
fix(accordion): accordion multiline toolbar

### LINKED ISSUE #1322: Styling on Accordion doesn't handle content height greater than 50px
**Summary**

The way the accordion is styled doesn't allow it to gracefully handle blocks of content with a height greater than 50px.  The styling needs to be adjusted so it's more dynamic.

**Reproduction**

Screen Shot 2020-07-05 at 7 03 36 PM

andrew-frueh:
Accordion can line-wrap multiple lines of text in the toolbar.  Also eliminated some unnecessary horizontal padding from the toolbar and content area.  I suspect most devs were overriding these values to clear them - but I left the vertical padding on the content area so this wouldn't be a visual break for those who weren't overriding.

closes #1322


andrew-frueh:
Thanks for reviewing while you're on vacation @corykon :-)
I've got an update for the pagination controller coming today too

corykon:
No prob! Did you happen to see my PR on the form controls that was submitted to your
1211 branch?

Would love to get that looked at and merged before we get too far behind Dev, esp w/ the Nuemont devs going to town on stuff all break long

PR #1331, opened 2020-07-15
feat(checkbox groups): adds checkbox group directive to component

### LINKED ISSUE #995: Checkbox Groups
**Describe your idea**

Similar to Radio Buttons, we should include a CheckboxGroup component.

This will allow for inline groups for checkboxes, it will also help support parent/child checkbox collections as was suggested in #953

**Additional context**

The implementation of Checkbox groups should be aligned with RadioGroups.  It should also include functions to check all, uncheck all, and return a list of what has been checked.

CameronAFaust:
closes #995


projects/cashmere/src/lib/checkbox/checkbox.component.scss
```diff
@@ -24,6 +24,10 @@
     }
 }
 
+.hc-children-group {
```
andrew-frueh, line 27: @CameronAFaust - since view encapsulation is off, we have to be really specific with our css naming to ensure there isn't any accidental naming collision (View Encapsulation fyi means that css classes apply globally - they aren't just limited to the scope of the component).  So let's change this naming to `hc-checkbox-children-group`

projects/cashmere/src/lib/sass/checkbox.scss
```diff
@@ -142,3 +142,8 @@
 @mixin hc-checkbox-stubbed() {
     pointer-events: none;
 }
+
+@mixin hc-children-group() {
```
andrew-frueh, line 146: same comment on the naming of the mixin

andrew-frueh:
@joeskeen I'll wait till you approve this one as well before I merge it

PR #1332, opened 2020-07-15
1203 hc tooltip

### LINKED ISSUE #1203: Improvements to using HcPop as Tooltips

Currently using HcPop as a tooltip has some disadvantages:

* Each tooltip needs its own `<hc-pop>...</hc-pop>` definition.  This is much more verbose than using the `[title]=""` attribute.  If you have many tooltips, this can quickly become the majority of your template file.
* Tooltips wired up using the `hover` trigger show immediately.  If you have tooltips on a control-dense view (such as a table), moving your mouse across the screen causes a ton of tooltips to flash in and out.
* Tooltip definitions are separated from the place they are used.  It's also unlikely that they will be reused.

**Describe your idea**

Implement one of the following improvements:

* Create a new `hcTooltip=""` directive that expects a string value, and pops up after a 500ms delay.  This could/should be based on HcPop under the hood.
* Create a new trigger for HcPop called something like `hoverDelay` for popping up on hover after a 500ms delay.  Allow HcPop to accept either an `<hc-pop>` reference or a plain string value.

**Additional context**

https://stackblitz.com/edit/angular-675wud?file=src/app/popover-tooltip/popover-tooltip-example.component.html

CristianIxcamey:
tooltip was made into a subclass oh popover along with allowing a hover delay to be input within popover-overview.
closes #1203


andrew-frueh:
@joeskeen - at our standup today, let's talk about how we want documentation to work for this.  Currently the docs site is building a page for `HcTooltipComponent`, which I don't think we want, because if I understand right we really only want folks to use the directive, and the component is just an internal element.  So let's look at if we hide that component from docs, and whether the tooltip example should go back under HcPop or have its own section.

PR #1334, opened 2020-07-17
feat(highlight pipe): added highlight pipe to cashmere

### LINKED ISSUE #578: Add Highlight Pipe to Cashmere
It's a common use case in apps to have a search or filter function, and it is nice to highlight the search term within the results.  Both of the apps I'm working on at HC have a Highlight pipe, and I'm sure other apps do as well.  Fixing bugs (especially security bugs like the one I found today) is hard to communicate and duplicate in disparate implementations.  Making this a standard-issue pipe in Cashmere would be really great!

The `HighlightPipe` would take plain text (which shouldn't contain HTML) and a search term to highlight.  The pipe then searches the provided text for the search term and wraps each match with a span marked with the special highlight class.  The result would be HTML-formatted text.

*   No components would have to change, and no breaking changes to the API, this would be completely additive.

**Additional thoughts**

1. We already have one pipe in Cashmere, the Ellipsis pipe.  But it isn't referenced in the documentation anywhere.  We probably should create a new section for pipes in the documentation, with interactive examples.
2. Here is the proof of concept on StackBlitz: https://stackblitz.com/edit/cashmere-demo-r7phhz?file=styles.scss
    image
3. Here is the spec for the Highlight pipe we use in Atlas, that I would propose we add to Cashmere:
    ```
    HighlightPipe
        when text and search contain no HTML
            when search term doesn't match casing with the text
                should still find the search term in the text and wrap the match with a highlight span
            when search term is found in the string
                should wrap the match with a highlight span
            when search term is multiple words
                if those multiple words are not adjacent in the text
                    should wrap each matching word from the search in its own highlight span
                should wrap each adjacent word in the match with its own highlight span
            when search term is found multiple times in the string
                should wrap all matches with highlight spans
            when search term is a partial word found in the string
                should wrap the match with a highlight span
            when search term is `undefined`
                should return the input string
            when search term is not found in the string
                should return the input string
        when text contains HTML
            should highlight matches on the escaped HTML
            when the search term contains HTML
                should highlight matches on the escaped HTML
            when the search term is blank
                should still escape the HTML in the text
        when text is `undefined`
            should return the empty string
        when both search term and text are `undefined`
            should return the empty string
    ```

Victor1818:
closes #578

### DIFF
--- a/projects/cashmere/src/lib/pipes/highlight/highlight-pipe.md
+++ b/projects/cashmere/src/lib/pipes/highlight/highlight-pipe.md
@@ -49,14 +49,22 @@ Pipe a `string` value to the Highlight pipe representing query. If the query is 
 
 -   When the Highlight Pipe search term is multiple words, it should wrap each adjacent word in the match with its own highlight span.
 
--   When the Highlight Pipe search term is a partial word found in the string, it should wrap the match with a highlight span.
+*   When the Highlight Pipe search term is multiple words and if those multiple words are not adjacent in the text, should wrap each matching word from the search in its own highlight span.
 
--   When the Highlight Pipe search term is found multiple times in the string, it should wrap all matches with highlight spans.
+*   When the Highlight Pipe search term is a partial word found in the string, it should wrap the match with a highlight span.
 
--   When the Highlight Pipe text value contains HTML, it should highlight matches on the escaped HTML.
+*   When the Highlight Pipe search term is found multiple times in the string, it should wrap all matches with highlight spans.
 
--   When the Highlight Pipe search term is blank, should still escape the HTML in the text.
+*   When the Highlight Pipe search term is not found in the string, it should return the input string.
 
--   When the Highlight Pipe text value is `undefined`, it should return the empty string.
+*   When the Highlight Pipe search term is `undefined`, it should return the input string.
 
--   When the Highlight Pipe search term and text value are `undefined`, it should return the empty string.
+*   When the Highlight Pipe text contains HTML, it should highlight matches on the escaped HTML.
+
+*   When the Highlight Pipe search term contains HTML, it should highlight matches on the escaped HTML.
+
+*   When the Highlight Pipe search term is blank, it should still escape the HTML in the text.
+
+*   When the Highlight Pipe text is `undefined`, it should return the empty string.
+
+*   When the Highlight Pipe search term and text are `undefined`, it should return the empty string.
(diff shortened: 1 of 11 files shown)


andrew-frueh:
@joeskeen - looks like changes to Victor's angular.json file got submitted in this PR, and we probably don't want to commit that.  Would would be the easiest way for Victor to roll back that change?

andrew-frueh:
awesome - thanks for removing that file @Victor1818

projects/cashmere/src/lib/pipes/highlight/highlight-pipe.md
```diff
@@ -0,0 +1,50 @@
...
+import {NgModule} from '@angular/core';
+import {PipesModule} from '@healthcatalyst/cashmere';
+
+@NgModule({
+    exports: [PipesModule]
+})
+export class CashmereModule {}
+```
+
+#### Using the Highlight Pipe in character mode
```
andrew-frueh, line 37: I would change the header here to `Case Sensitivity`.  Character mode is actually one of the inputs to the ellipsis pipe so it doesn't apply here.  The description in this section is fine though

projects/cashmere/src/lib/pipes/highlight/highlight-pipe.md
```diff
@@ -0,0 +1,50 @@
...
+Pipe a `string` value to the Highlight pipe representing query. If the query is found in the text, the value is then highlighted disregarding case sensitivity.
+
+```html
+{{ 'at' }}
+<!-- will highlight At…at…  -->
+{{ 'At' }}
+<!-- will highlight At…at…  -->
+```
+
+#### Edge cases
```
andrew-frueh, line 48: Just looking at the unit tests Joe built, we probably want to edit these edge cases.  For example, what happens if the text contains HTML, what happens when the search term is blank, etc.

If you want to take a crack at this @Victor1818 - you can use the language @joeskeen built into his unit tests directly.  For example, one edge could be taken from line 70 in the spec file, and you could add it as a bullet in these edge cases like: "When the text value for the Highlight Pipe contains HTML, it should highlight matches on the escaped HTML".  And you could do something similar for just about all his tests in the spec - just in bullet list form under this Edge cases section

joeskeen:
Approved once you resolve @andrew-frueh 's comment

andrew-frueh:
Hey @Victor1818 - the content for the additional edge cases looks good.  Thanks for getting those added.  When I pulled up the Usage tab for that pipe though, this is what I was seeing:

Screen Shot 2020-07-23 at 9 48 38 AM

The markdown renderer was getting confused by the line breaks between the bullets.  Be sure and pull up the test site to check for stuff like this.  For things like this that live in the generated docs pages, you won't be able to do your normal `npm run start`.  You'll need to `npm run build` and then `ng serve`, because the build step is where the documentation is compiled.

I'll go ahead and pushed a commit with the fix to the markdown so we can get this wrapped up, but keep that in mind going forward.

PR #1335, opened 2020-07-17
refactor(ivy compatibility): minor changes to prepare for ivy

corykon:
A collection of the majority of the manual changes I needed to make while upgrading to angular 9 w/ Ivy, to prep us for when it's time to make the actual transition.

### DIFF
--- a/projects/cashmere/src/lib/scroll-nav/scroll-nav.module.ts
+++ b/projects/cashmere/src/lib/scroll-nav/scroll-nav.module.ts
@@ -1,6 +1,6 @@
 import {NgModule} from '@angular/core';
 import {CommonModule} from '@angular/common';
-import {ScrollDispatchModule} from '@angular/cdk/scrolling';
+import {ScrollingModule} from '@angular/cdk/scrolling';
 
 import {HcScrollNavContentComponent} from './content/scroll-nav-content.component';
 import {HcScrollNavComponent} from './nav/scroll-nav.component';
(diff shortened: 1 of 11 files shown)


projects/cashmere/src/lib/typeform-survey/typeform-survey.component.ts
```diff
@@ -57,7 +57,7 @@ export class TypeformSurveyComponent {
         if (!document.getElementById(this._id)) {
             this.getScripts();
         } else {
-            (<TypeformWindow>window).typeformEmbed.makePopup(this._fullUri, {
+            (<TypeformWindow>(<unknown>window)).typeformEmbed.makePopup(this._fullUri, {
```
joeskeen, line 60: I know you are working with some existing code here, but ever since TypeScript introduced JSX support, the `<>` form of casting has been deprecated-ish.  I would rewrite it as `((window as unknown) as TypeformWindow).`

projects/cashmere/src/lib/scroll-nav/scroll-nav.module.ts
```diff
@@ -10,7 +10,7 @@ import {ScrollNavLinkDirective} from './nav/scroll-nav-link.directive';
 @NgModule({
     declarations: [HcScrollNavContentComponent, HcScrollNavComponent, ScrollNavTargetDirective, ScrollNavLinkDirective],
     exports: [HcScrollNavContentComponent, HcScrollNavComponent, ScrollNavTargetDirective, ScrollNavLinkDirective],
-    imports: [CommonModule, ScrollDispatchModule],
+    imports: [CommonModule, ScrollingModule],
```
joeskeen, line 13: Is there a certain version or version range of the CDK that this change will work with?

projects/cashmere/src/lib/typeform-survey/typeform-survey.component.ts
joeskeen: hmm... looks like the TypeScript team has changed their position on this slightly: https://www.typescriptlang.org/docs/handbook/basic-types.html#type-assertions
fwiw, it feels like we should try to be consistent as much as possible in Cashmere, and I would personally vote for the `as` syntax, but I won't hold up the review for it.

projects/cashmere/src/lib/scroll-nav/scroll-nav.module.ts
corykon: @joeskeen  ScrollDispatchModule was deprecated in version 7, and we're requiring cashmere implementers to be on angular 8

Issue #1329, opened 2020-07-10
Hide firefox active outlines in navbar


andrew-frueh:
**Summary**

Since our navbar already highlights the active item, we don't need Firefox's dotted line around the active navbar item.

**Additional context**

<img width="426" alt="Screen Shot 2020-07-10 at 10 54 13 AM" src="https://user-images.githubusercontent.com/22795893/87179370-de172680-c29b-11ea-89c1-f1baab3a88fc.png">

andrew-frueh:
Interestingly - the dotted outline doesn't appear when you first click on a navbar item, it's when you switch to another tab and come back to the page.

PR #1341, opened 2020-07-22
fix(hide firefox outline navbar): navbar outline ceases to exist

Victor1818:
re #1329


andrew-frueh:
Cool - yeah this PR looks correct.  Got one I submitted merged - #1341.  So we're good on this.  I tested the change locally and your scss change looks like it did the trick.  Go ahead and get those additional items added to the docs on your highlight pipe so we can get that merged too.

PR #1346, opened 2020-07-28
Fix ng-select caret vertical alignment

mhollier:
Verified fix in IE11, Chrome, Firefox, and Edge.

image

### DIFF
--- a/projects/cashmere/src/lib/sass/_ng-select-theme.scss
+++ b/projects/cashmere/src/lib/sass/_ng-select-theme.scss
@@ -185,6 +185,10 @@ $ng-select-value-font-size: 14px !default;
             pointer-events: auto;
         }
     }
+
+    .ng-select-container .ng-arrow-wrapper {
+        height: 33px;
+    }
 }
 
 ng-dropdown-panel.ng-dropdown-panel {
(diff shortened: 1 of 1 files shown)


projects/cashmere/src/lib/sass/_ng-select-theme.scss
```diff
@@ -185,10 +185,6 @@ $ng-select-value-font-size: 14px !default;
             pointer-events: auto;
         }
     }
-
-    .ng-select-container .ng-arrow-wrapper {
-        height: auto;
```
corykon, line 190: I'm seeing a tiny side effect where the field becomes 37px tall instead of 35px, which messes with the vertical alignment of the placeholder text and makes it so there's a difference in the heights of this component and a regular selectbox.

Instead of removing the styles here, could you change line 190 to set the height to 33px?

If I'm reading things right, I think the height: auto; styles were added to override the height being pulled in from a mixin shared with the select component styles.

mhollier: Will do. Good catch.

mhollier: Done!

Issue #1351, opened 2020-07-29
Deep linking support for all of the Cashmere documentation


joeskeen:
In order to implement #1123 , we first need a way to navigate to any piece of content on the site.  We can accomplish this by adding `id` attributes to header tags, for example:

```
<h2 id="some-header">Some header</h2>
```
This will allow navigating to the header using the hash `#some-header` at the end of the URL.
You can play around with this idea here: <https://stackblitz.com/edit/angular-heading-hash-links?file=src/app/app.component.html>

Where headers are unavailable, we can use anchors (`<a name="some-section"></a>`) to do the same thing.

Tasks for completing this:

- [x] Create a deep link to every example on each component page (see `example-viewer.component.html`)
- [x] Create a deep link to every tab in the example (TS, HTML, CSS, etc)
        _This could be a stinker_
- [x] Create a deep link for every header rendered from Markdown
        Currently we are using `markdown-it` for rendering Markdown to HTML, and it does not support rendering ids or anchors in headers.  Alternatively, the `marked` package supports GFM and rendering ids in headers, so we should explore replacing the `markdown-it` library with `marked` (https://github.com/markedjs/marked) - ![Marked GitHub stars](https://img.shields.io/github/stars/markedjs/marked?style=flat)
- [x] Create a deep link for headers within the generated API documentation - already done :)
        ~We use dgeni (<https://github.com/angular/dgeni>) for generating our API documentation from source.  It may be tricky, but we need to figure out how to get the proper headings and/or anchors inserted into the generated docs.  If this isn't possible, we may need to consider an alternative code documentation generator (like TypeDoc or something).~
        Luckily, dgeni already does this for us!
- [x] Fix the CSS for the Cashmere site to scroll under the navbar, instead of to the right of the navbar.  Otherwise the header gets cut off by the navbar when you go to the hash link.

PR #1353, opened 2020-07-30
Add Cashmere Bits

### LINKED ISSUE #960: Publish individual Cashmere components as Bits
**Describe your idea**

One thing also that Andrew and I have talked about is allowing individual components to be versioned and released (in addition to the all-in-one package we offer today).  This would allow bug fixes related to Cashmere in previously released versions of products to not require updating ALL of Cashmere to get the needed bug fix.
(One example of this is a bug in the TypeForm survey component that leaves the application in a bad state after the user completes the survey until the next page refresh.  We definitely want to fix that in all versions using TypeForm, but we don't want to have to upgrade all of Cashmere throughout the application to take advantage of that bug fix.)
There is a service that allows publishing individual components in this fashion called Bit (https://bit.dev).  While historically they've only hosted components for ReactJS, they are now opening a public beta of Angular and VueJS support.  I've been in contact with their support team, and I think we could with a few structural (non-breaking) changes be able to isolate Cashmere components and publish them individually.  The great thing about using Bit is that it simply utilizes NPM (meaning you don't have to install another package manager to consume individual 'bits').  For example, if we wanted to publish the TypeForm survey component by itself you could install it using npm i @bit/healthcatalyst.cashmere.typeform-survey.  I've figured out how we can maintain semantic releases on these individual components too, similar to how we are doing it for the overall package using semantic-release.
In addition to allowing individual components to be installed in release branches, having an individual component route could make it possible for other types of components to join the Cashmere family, particularly ones that require additional dependencies (since adding additional dependencies to Cashmere is not something to be done without serious consideration to the ramifications it brings to everyone who uses it, even if they don't use the particular component that requires it).

**Additional Context**

FYI I stood up a Cashmere collection (currently empty): https://bit.dev/healthcatalyst/cashmere

joeskeen:
Over time, there have been many great ideas for Cashmere that were rejected because they would require Cashmere to take on additional dependencies, which we try hard to not do.  But is there enough room in the world for these great ideas? Yes!

Introducing Cashmere Bits, a collection of Cashmere-family components that for whatever reason cannot be included in the main library.  These components are versioned and published as independent packages to <https://bit.dev/healthcatalyst/cashmere>, which can be installed using NPM.

This PR adds the first Bit, a pipe depending on the `change-case` NPM package.  You can use this to transform text to a number of different casing formats:
image
image

This PR also does some much-needed refactoring to the components viewer in the docs site to be more flexible.

Best part? I didn't have to change anything under `/projects/cashmere`!

Sorry, a lot of files changed, but this was implemented in a single day during some innovation time, so hopefully it shouldn't take too much mental capacity to review :) 

fix #960, re #1351


joeskeen:
Currently I'm investigating a Linux build issue (it works find on Windows, but is failing on Travis).  Please don't let that scare you away from reviewing the code 👍

andrew-frueh:
Exciting to see this idea coming to reality @joeskeen!  I think you're fastest path to getting an approval on this would be to schedule a quick live review meeting with the steering committee to go over the changes.  If you're good with that, feel free to schedule a time that works for you.

src/app/components/components.component.ts
```diff
@@ -1,52 +1,92 @@
...
+            this.activeCategory = this.activeItem.category;
+        }
     }
 
-    // Handle nav changes via the sidebar
-    navUpdate(id: any) {
-        this.router.navigate(['/components/' + id]);
+    // Handle nav changes via the sidebar or mobile dropdown
+    navUpdate(id: string) {
+        console.log('new selection', id);
```
joeskeen, line 88: remove this!

joeskeen: Done! 👍

Issue #879, opened 2019-05-28
add Angular Schematics for Cashmere developers


joeskeen:
**Describe the problem this feature request is solving**
Wouldn't it be nice if creating a new component in the Cashmere library was as easy as `ng g cashmere-component <new-component-name>`?

**Describe the solution you'd like**
With the Angular Schematics functionality built-into the CLI, this is totally possible!  It could create the new directory in the `projects/cashmere/src/lib` directory for the component, complete with all the files needed, add it to `public_api.ts`, add a `<new-component-name>-overview` example, and everything else needed for that.  Basically, you would be ready for `npm start` to develop your component in one command.

**Describe alternatives you've considered**

While other tools exist that could do this, sticking with the standard Angular CLI aligns with best practices.

** If any, which components will have to change, and how **

-   Components changed? **nope**
-   Are these breaking changes to public interfaces? **nope**

**Additional context**
Oh, and while we are at it, we could also make another schematic for `ng g cashmere-example <my-example-name>`.

PR #1357, opened 2020-08-04
build(new:example): add new example generation script

joeskeen:
Creating an example should only take a few minutes now, and require little reading of the
documentation! :)

Ever since I revamped the examples to make them work with StackBlitz, I've always felt a bit guilty as to how unfriendly the example creation process is.  The docs were lengthy, incomplete, and difficult to get everything right.

So I finally bit the bullet and wrote a script for generating a new example!

Now you can just run `npm run new:example` and it will scaffold All The Things needed for that example to run.  It even runs the build so that you can `npm start` and immediately start dev'ing your new example!

This work is in preparation for similar scripting for creating new Cashmere Components and Cashmere Bits (those processes will generate an "___ overview" example).  With these automations, the barrier to contribution to the Cashmere project should be greatly reduced!

### DIFF
--- a/guides/working-with-examples.md
+++ b/guides/working-with-examples.md
@@ -12,6 +12,16 @@ Adding meaningful examples can greatly enrich the documentation of components in
 
 :::
 
+##### Creating a new example
+
+From the root of the Cashmere repository, run `npm run new:example`. This will start an interactive prompt to gather all the information needed to create your example.
+
+_Please don't mess with any of the generated names in the files, otherwise your example may not work on the documentation site and/or StackBlitz._
+
+:::
+
+:::
+
 ##### Example files
 
 The following files are used in the creation of examples for the Cashmere documentation website.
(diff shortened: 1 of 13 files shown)


guides/working-with-examples.md
```diff
@@ -12,6 +12,16 @@ Adding meaningful examples can greatly enrich the documentation of components in
 
 :::
 
+##### Creating a new example
+
+From the root of the Cashmere repository, run `npm run new:example`. This will start an interactive prompt to gather all the information needed to create your example.
+
+_Please don't mess with any of the generated names in the files, otherwise your example may not work on the documentation site and/or StackBlitz._
```
joeskeen, line 19: The guide for examples is much more concise! :)

package.json
```diff
@@ -21,8 +22,8 @@
     "lint:prettier": "prettier --debug-check \"**/*.{ts,js,scss,css,md,json}\"",
     "build:lib": "ng build cashmere --prod",
     "postbuild:lib": "node scripts/postbuild.js",
-    "prebuild:bit": "cd projects/cashmere-bits && npx --package bit-bin bit config set analytics_reporting false && npx --package bit-bin bit config set error_reporting false",
-    "build:bit": "cd projects/cashmere-bits && npx --package bit-bin bit import && npx --package bit-bin bit build",
+    "prebuild:bit": "cd projects/cashmere-bits && npm run bit -- config set analytics_reporting false && npm run bit -- config set error_reporting false",
+    "build:bit": "cd projects/cashmere-bits && npm run bit -- import && npm run bit -- build",
```
joeskeen, line 26: I noticed that the former command was WAY slower than doing it this way.

scripts/new-example.ts
```diff
@@ -0,0 +1,271 @@
...
+    args = {...args, ...input};
+    args.name = paramCase(args.name);
+    args.requiredPackages = (args.requiredPackages || [])
+        .map((p: string) => p!.trim().toLowerCase())
+        .filter(p => !currentExampleDependencies.includes(p));
+
+    prettierConfig = (await prettier.resolveConfig(__dirname))!;
+}
+
+(async function newExample() {
```
joeskeen, line 130: This is the main algorithm for creating a new example:
* gather any missing information from the user
* create the new example files
* register the new example with the document items service
* install any additional NPM packages the new example needs to work
* run the build to make sure we are in a good state and that the example module is regenerated

The user should be ready to `npm start` after running this.

scripts/new-example.ts
```diff
@@ -0,0 +1,271 @@
...
+selector: 'hc-${args.name}-example',
+templateUrl: '${args.name}-example.component.html',
+// TODO: delete the SCSS file if you don't need it in the example
+styleUrls: ['${args.name}-example.component.scss']
+})
+export class ${pascalCase(args.name)}ExampleComponent {
+// TODO: implement your example here
+}
+`;
+    writeFileSync(exampleComponentFileName, prettier.format(componentFileContents, {...prettierConfig, parser: 'typescript'}));
```
joeskeen, line 162: Trying to avoid petty formatting changes later...

scripts/new-example.ts
```diff
@@ -0,0 +1,271 @@
...
+        writeFileSync(exampleModuleFileName, prettier.format(moduleFileContents, {...prettierConfig, parser: 'typescript'}));
+    }
+}
+
+async function registerWithDocumentItemsService() {
+    console.info(chalk.gray(`registering example with document items service...`));
+    const docItemsDirectory = join(__dirname, '../src/app/core');
+    let docItemsFile: string;
+    if (args.category === 'cashmere') {
+        docItemsFile = join(docItemsDirectory, './cashmere-components-document-items.json');
```
joeskeen, line 200: I found that manipulating JSON files was easier than manipulating TypeScript source files.  As such, all example metadata has been moved to JSON files.

src/app/components/components.component.ts
```diff
@@ -71,8 +71,8 @@ export class ComponentsComponent implements OnInit, OnDestroy {
         }
 
         this.allDocItems = this.docItemService.getDocItems(this.docType);
-        if (!this.id) {
-            this.navUpdate(this.allDocItems[0].id);
+        if (!this.id && this.categorizedDocItems[0] && this.categorizedDocItems[0].items && this.categorizedDocItems[0].items.length) {
```
joeskeen, line 74: Ran into an edge case where it navigated to `/components/undefined/examples`... this should prevent that.

src/app/core/document-items.schema.json
```diff
@@ -0,0 +1,54 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
```
joeskeen, line 2: My only hesitation moving from TS to JSON for the document items is I wanted to make sure there was still dev-time validation.  This JSON schema accomplishes that.

src/app/core/document-items.service.ts
```diff
@@ -1,194 +1,35 @@
...
-        category: 'popups',
-        examples: ['typeform-survey-overview'],
-        usageDoc: true
-    }
-];
-
-const bitDocs: DocItem[] = [
-    {id: 'change-case-pipe', name: 'Change Case', category: 'pipes', usageDoc: true, hideApi: true, examples: ['change-case-pipe-overview']}
-];
+const cashmereComponents: DocItem[] = Object.keys(cashmereComponentsDocumentItems)
```
joeskeen, line 18: Separating data from logic for the win! :)

joeskeen:
@andrew-frueh @corykon Don't be afraid to look at this one - it's really pretty straightforward.  If you would like me to walk you through it though please don't hesitate to ping me.

joeskeen:
Re: #879

PR #1362, opened 2020-08-05
fix(am-pm automatic switching): am and pm times switch automatically

### LINKED ISSUE #1338: AM/PM does not automatically switch when changing time
Not so much a bug as some behavior that some users are finding to be unexpected:

> Wanted to report a bug in scheduling within Ops Console. When you click the down arrow on the time it should automatically change from PM to AM, it does not do this though. In my example if I have it set to 12:30 PM and want it to run an hour earlier, I should be able to just click the down arrow on the hour and it would resolve to 11:30 AM. Current functionality is that it would resolve to 11:30 PM (11 hours later)... [Version DOS 20.2]

image

Victor1818:
closes #1338

### DIFF
--- a/projects/cashmere/src/lib/datepicker/calendar/calendar.component.ts
+++ b/projects/cashmere/src/lib/datepicker/calendar/calendar.component.ts
@@ -530,17 +530,11 @@ export class CalendarComponent implements AfterContentInit, AfterViewChecked, On
         } else {
             let curHour = +this.hours;
             curHour++;
-            const tempCycle = this._hourCycle > 12 ? 23 : 12;
+            const tempCycle = this._hourCycle > 12 ? 24 : 12;
             if (curHour > tempCycle) {
                 curHour = this._hourCycle > 12 ? 0 : 1;
             }
-            if (curHour === 12) {
-                if (this._period.value === 'am') {
-                    this._period.setValue('pm');
-                } else if (this._period.value === 'pm') {
-                    this._period.setValue('am');
-                }
-            }
+            this._changeMeridiem(curHour, 12);
             this.hours = curHour.toString();
         }
     }
@@ -555,14 +549,14 @@ export class CalendarComponent implements AfterContentInit, AfterViewChecked, On
             if (curHour < tempCycle) {
                 curHour = this._hourCycle > 12 ? 23 : 12;
             }
-            if (curHour === 11) {
-                if (this._period.value === 'am') {
-                    this._period.setValue('pm');
-                } else if (this._period.value === 'pm') {
-                    this._period.setValue('am');
-                }
-            }
+            this._changeMeridiem(curHour, 11);
             this.hours = curHour.toString();
+        }
+    }
+
+    _changeMeridiem(curHour, hourChange) {
+        if (curHour === hourChange) {
+            this._period.setValue(this._period.value > 'am' ? 'am' : 'pm');
         }
     }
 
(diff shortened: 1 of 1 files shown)


projects/cashmere/src/lib/datepicker/calendar/calendar.component.ts
```diff
@@ -532,7 +532,14 @@ export class CalendarComponent implements AfterContentInit, AfterViewChecked, On
             curHour++;
             const tempCycle = this._hourCycle > 12 ? 23 : 12;
             if (curHour > tempCycle) {
-                curHour = tempCycle;
+                curHour = this._hourCycle > 12 ? 0 : 1;
+            }
+            if (curHour === 12) {
+                if (this._period.value === 'am') {
```
jonest, line 538: Pretty minor, but looks like this whole conditional for `am`, `pm` switch is something that could probably be pulled into a single method

jonest: Or instead to make it more simple, just a ternary (assuming that it's always going to be `am` or `pm`? I'm not sure if there's a 24-hour setup for this

projects/cashmere/src/lib/datepicker/calendar/calendar.component.ts
```diff
@@ -532,7 +532,14 @@ export class CalendarComponent implements AfterContentInit, AfterViewChecked, On
             curHour++;
             const tempCycle = this._hourCycle > 12 ? 23 : 12;
             if (curHour > tempCycle) {
-                curHour = tempCycle;
+                curHour = this._hourCycle > 12 ? 0 : 1;
```
jonest, line 535: I don't really understand what's being accomplished here (I don't really know the code for the date picker stuff)

projects/cashmere/src/lib/datepicker/calendar/calendar.component.ts
Victor1818: @jonest A separate method is what I was also thinking too. I can try a few things with the ternary but if that doesn't work, would you have an alternative idea?

projects/cashmere/src/lib/datepicker/calendar/calendar.component.ts
Victor1818: @jonest This line allows the 12 hour or 24 hour clock to cycle instead of having a max of 12 and 24 hours respectively and a min of 1 and 0 hours respectively.

projects/cashmere/src/lib/datepicker/calendar/calendar.component.ts
```diff
@@ -546,12 +547,19 @@ export class CalendarComponent implements AfterContentInit, AfterViewChecked, On
...
             if (curHour < tempCycle) {
-                curHour = tempCycle;
+                curHour = this._hourCycle > 12 ? 23 : 12;
             }
+            this._changeMeridiem(curHour, 11);
             this.hours = curHour.toString();
         }
     }
 
+    _changeMeridiem(curHour, hourChange) {
```
jonest, line 557: TIL it's spelled `meridiem` for time 😃

jonest:
Whoops! I didn't think I had to do anything more to add my comments so sorry for the delay :/

projects/cashmere/src/lib/datepicker/calendar/calendar.component.ts
Victor1818: It just allows the time to reset at its lowest or highest hours instead of remaining at its lowest or highest hours

andrew-frueh:
Thanks for the code review and working with @Victor1818 to wrap this one up @jonest.  Really appreciate it.

PR #1366, opened 2020-08-10
fix(tooltip): set explicit line-height

### LINKED ISSUE #1365: Tooltip style issue
The tooltip component positions its caret/arrow connector in a way that is dependent on a line-height that is 14px, but that line-height is not explicitly set (it just inherits the default body line height).

Because of this, if an app implements the tooltip and has a different line-height for the body, the styling will be off.
image

Solution is pretty easy. Just need to set a specific line height to `.cdk-overlay-container` and we'll be safe.

corykon:
fix #1365


corykon:
@andrew-frueh all good there, no impact to menus or toasts.

Also just now added one more commit exporting the component file in the module. I wanted to reference the component in some typescript code but had a hard time finding it

Issue #1370, opened 2020-08-14
Need to clear timeouts on popover hover destroy


andrew-frueh:
**Summary**

With the tooltip directive, we're getting a console errors on window timeouts that still try to fire after the popover no longer exists.

**Additional context**

![Screen Shot 2020-08-13 at 6 11 16 PM](https://user-images.githubusercontent.com/22795893/90198982-9f8bf480-dd90-11ea-8f98-344ca62c7a0e.png)

PR #1373, opened 2020-08-20
1203 hc tooltip

CristianIxcamey:
Got the timer clearing on destroy. Looked into the issue with tooltip not disappearing on the scroll when within a scrollable container but was not able to come up with a good solution.


corykon:
Believe this closes #1370

Issue #1361, opened 2020-08-05
Keyboard Accessibility for Navbar


corykon:
James Jensen brought these to my attention:

When attempting to tab through the navbar:
- I have no visual indication which navbar links have focus
- navbar icons, (app-switcher/help, etc) don't accept focus

PR #1377, opened 2020-08-21
fix(navbar): allow keyboard accessibility

corykon:
- popover anchor with a trigger of "click" can be opened by enter key when that anchor has focus
- tweak styles for anchor links and icons when focused
- updated examples to illustrate proper tabindex settings in the navbar

re #1361


corykon:
@joeskeen @andrew-frueh Can this get a review? Would love to get this merged in for the next release. I've got a bug on the pop builder board that this will unblock for me

andrew-frueh (changes requested):
Hey @corykon - everything looks good in Chrome, but unfortunately in Firefox the tabindex seems to skip navbar items and go straight to the "more" dropdown.  Can you test this out in Firefox to see if you get the same results?

andrew-frueh (approved):
@corykon pointed me here:
https://stackoverflow.com/questions/3846868/why-wont-tabindex-work-with-firefox

Looks like it's a Firefox / MacOS issue

Issue #1352, opened 2020-07-29
IE: hc-select is buggy when options are objects


corykon:
It would awesome if hc-select supported this feature that you can use in angular on normal selects:
https://stackoverflow.com/a/44771382/2311702

Makes it easier to use objects as options.

corykon:
Looks like the selectedIndex property we're relying for this component isn't fully supported in IE.
![image](https://user-images.githubusercontent.com/12416432/91071098-ff866480-e5f4-11ea-96fe-9f5085f5c91a.png)

PR #1379, opened 2020-08-24
1352 hc select compare with

corykon:
re #1352 

- Added support for `compareWith` on `hc-select`
- Fixed buggy behavior when in IE
- Turned off hover styles when the component is disabled

Relied heavily on the angular codebase while implementing. Referenced this file - (https://github.com/angular/angular/blob/10.0.11/packages/forms/src/directives/select_control_value_accessor.ts#L28-L194)

### DIFF
--- a/projects/cashmere/src/lib/select/select.component.ts
+++ b/projects/cashmere/src/lib/select/select.component.ts
@@ -10,20 +10,24 @@ import {
     Self,
     Output,
     EventEmitter,
-    ContentChildren,
-    QueryList,
     ViewChild,
-    AfterContentInit
+    Renderer2
 } from '@angular/core';
 import {ControlValueAccessor, NgForm, FormGroupDirective, NgControl} from '@angular/forms';
 import {HcFormControlComponent} from '../form-field/hc-form-control.component';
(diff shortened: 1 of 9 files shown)


projects/cashmere/src/lib/select/select.component.ts
```diff
@@ -34,17 +38,14 @@ export class SelectChangeEvent {
...
+export class SelectComponent extends HcFormControlComponent implements ControlValueAccessor, DoCheck {
     private _uniqueInputId = `hc-select-${uniqueId++}`;
     private _form: NgForm | FormGroupDirective | null;
     private _tight: boolean = false;
     private _value: any = '';
-    private _valueData: any;
-
-    _componentId = this._uniqueInputId;
-
-    @ContentChildren(HcOptionDirective)
```
corykon, line 46: Following angular's lead and having select options register themselves with the select component, rather than querying for them from the parent.

joeskeen (approved):
Approved, but will defer until next major version as it is a breaking change.

andrew-frueh (approved):

PR #1383, opened 2020-09-04
fix(app-switcher): error message

### LINKED ISSUE #1382: Successful app switcher load after an initial load failure doesn't get rid of retry error message
**Summary**

If the app switcher fails to load the available applications, it will give you a message asking you to retry. If the retry is successful, the applications will load in but the retry message doesn't go away. I have a fix that I'm working on.

**Screenshot**

image

braydengramse:
Retry error message will now disappear after success rather than stick around

closes #1382


andrew-frueh (approved):
Nice work @braydengramse

corykon (approved):

PR #1400, opened 2020-10-05
Mobile menu scroll

### LINKED ISSUE #1386: Mobile Menu Doesn't Scroll
**Summary**

This was called out on the HAS20 site.  When the mobile menu extends beyond the height of the window, the user can't scroll the menu.

**Reproduction**

Open the Cashmere site, shrink your browser window to trigger the mobile menu, and shrink the height of your window to smaller then the height of the browser window.  Try to scroll either with the scroll wheel, or more importantly with click/drag as in a touch environment.

**Additional context**

Fix probably involves setting an `overflow` value on that menu and constraining the height to `100vh`

skilletspecial:
closes #1386 "Mobile Menu Doesn't Scroll"


andrew-frueh:
Awesome!! Thanks for picking this one off @skilletspecial - were you able to test on a mobile device or emulator to validate the fix?

skilletspecial:
I'm not sure why two of @corykon 's merge commits show up here.  I merged everything in before creating my branch.  🤷

skilletspecial:
@andrew-frueh I tested in the browser's device dev tools and things appear to work exactly as I would expect

corykon:
> I'm not sure why two of @corykon 's merge commits show up here. I merged everything in before creating my branch. 🤷

You might've branched off of master instead of dev? Shouldn't be a problem.

andrew-frueh (changes requested):
Hey @skilletspecial - I tested the changes and I can confirm that the changes are allowing the mobile menu to scroll.  So thumbs up on that.  But it looks like when you try to close the mobile menu, the `bottom: 0px` bit is causing an issue with the menu fully sliding under the navbar (see attached).  It makes it about halfway and stops - this doesn't happen when I disable the bottom: 0px bit.

To test, I added a bunch of additional `hcListItems` to the mobile menu in `app.component.html` so I could validate the scrolling.  So I think you're really close.  You might take a look at the `hide()` function in `navbar-mobile-menu.component.ts` to see if that `_yPos` value needs to be calculated differently.

andrew-frueh:
@skilletspecial I decided to look into what was causing the issue, and wound up getting sucked in.  It turns out we had a few other mobile issues that had crept in over time with the select dropdowns and top margins.  So I took care of those, and as I was messing around with the `bottom` position value, I realized that the css transition was part of our problem.  Not sure why we didn't go with Angular animations for the menu right from the beginning, but it really simplifies the component.  So switched to that - and one thing I realized when you put the bottom: 0px change in place - was that the menu might as well always be full screen height.  That simplifies things from both a scrolling & positioning perspective.

@skilletspecial @corykon can you give the changes a quick sanity check for me?

Looks like I have one unit test to fix.

andrew-frueh (approved):

skilletspecial:
Thanks @andrew-frueh!  It's been a crazy week.  Your changes look good to me.

corykon (approved):

PR #1411, opened 2020-10-12
feat(chips): adds blue chip type

andrew-frueh:
Hey Joe, I made some style adjustments to your PR.  Based on how it was being used, it made more sense for that notification above the Bit tile to be a chip - but as I made that switch it exposed a few shortcomings of the component.  I added a `tight` param that we have on several other components that removes margin/padding.  And I added a `width` param so the dev can decide whether the chip should take the full width of its space (it's currently set to auto).

I also cleaned up some formatting on your guide. Let me know if you have any concerns.

### DIFF
--- a/projects/cashmere/src/lib/chip/chip.component.html
+++ b/projects/cashmere/src/lib/chip/chip.component.html
@@ -1,4 +1,4 @@
-<div class="hc-chip hc-chip-{{color}}" [ngClass]="{'hc-chip-padding': !tight}" [ngStyle]="{'width': width}">
+<div class="hc-chip hc-chip-{{color}}" [class.hc-chip-padding]="!tight" [style.width]="width">
     <span class="hc-chip-content-{{color}}">
         <ng-content></ng-content>
     </span>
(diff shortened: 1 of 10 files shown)


projects/cashmere/src/lib/chip/chip.component.html
```diff
@@ -1,4 +1,4 @@
-<div class="hc-chip hc-chip-{{color}}">
+<div class="hc-chip hc-chip-{{color}}" [ngClass]="{'hc-chip-padding': !tight}" [ngStyle]="{'width': width}">
```
joeskeen, line 1: I would much prefer the following:

```html
<div class="hc-chip hc-chip-{{color}}" [class.hc-chip-padding]="!tight" [style.width]="width">
```

joeskeen (changes requested):
Just a couple small changes, for the most part looks awesome!

projects/cashmere/src/lib/chip/chip.component.html
andrew-frueh: Ah, good call - I'll update that.

PR #1412, opened 2020-10-12
refactor(docs): remove duplicate code

joeskeen:


src/app/styles/base-styles.component.ts
```diff
@@ -0,0 +1,14 @@
+import {AfterViewInit} from '@angular/core';
+import {SectionService} from '../shared/section.service';
+
+export class BaseStylesComponent implements AfterViewInit {
+    constructor(protected sectionService: SectionService) {}
+
+    ngAfterViewInit() {
+        this.loaded();
```
joeskeen, line 8: By default it will try to load on init, but the derived component will be able to call `loaded` at any time they wish (directly from the template in most cases).  This removes almost all the boilerplate code previously in this branch.

andrew-frueh (approved):

PR #1413, opened 2020-10-12
Tile header

### LINKED ISSUE #1144: Header directive on Hc-Tile
**Describe your idea**

Add an `hc-title-header` directive to `hc-tile`.

There are a couple really common ways that we display titles on a `hc-tile`.  We should standardize those into directives so developers don't have to style these on their own.

**Additional context**

Header example one:
Artboard – 2

Header example two:
Screen Shot 2020-01-30 at 11 13 29 AM

### LINKED ISSUE #1307: Add unit tests for the Tile component
**Summary**

_Let us know what went wrong. The rest of this form is optional._

**Reproduction**

Steps to reproduce:

1.

**Additional context**

andrew-frueh:
@corykon & @jonest - let me know what you think about this update.  Ty suggested this a while back, because there are common styles we use for tile headers that aren't really baked into Cashmere anywhere for devs to easily leverage.  I added two styles that I use regularly for tile headers, but if you have others, please let me know and I'll add them.

@joeskeen - let me know if I did the unit test right :-)

closes #1144
closes #1307

### DIFF
--- a/projects/cashmere-examples/src/lib/tile-overview/tile-overview-example.component.html
+++ b/projects/cashmere-examples/src/lib/tile-overview/tile-overview-example.component.html
@@ -1,5 +1,12 @@
 <hc-tile>
-    <div hcTileHeader [type]="headerType">Example Tile Header</div>
+    <div hcTileHeader [type]="headerType">
+        <ng-container *ngIf="headerType !== 'center'">Example Tile Header</ng-container>
+        <div *ngIf="headerType === 'center'" class="tile-example-center">
+            <hc-icon fontSet="hc-icons" fontIcon="hci-medical" hcIconLg></hc-icon>
+            <div>Example Tile Header</div>
+            <p>Example tile subheader description text</p>
+        </div>
+    </div>
 
     <div class="tile-example-select">
         <hc-form-field inline="true">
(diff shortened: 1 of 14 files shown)


corykon (approved):
Looks good to me. Any guidance we could offer on why someone would use "blue" vs "title"?

corykon:
Leading Wisely might be a good team to check in with on tile header styles

andrew-frueh:
Good question, I did give that some thought - and I agree that I think the directive has value beyond the header styles, particularly on the blue uppercase one.  It also allows us to include some margins that we wouldn't want on the typography styles.

@jacquie - re: Cory's point above, do you have any additional header styles that we should include for the tile directive?

andrew-frueh:
@corykon - do you think I should add a type called `center` or something like that?  It could be used to create those clinical, case, member tiles you attached.

andrew-frueh:
Ah that's super helpful @jacquie - that's very similar to the use case from @corykon.  I was already working on a `center` type, which would probably work in your case - but it would be better if I didn't bake the color into it (see attached).

Screen Shot 2020-10-13 at 11 07 44 AM

The markup would look like:
```
<div hcTileHeader [type]="headerType">
            <hc-icon fontSet="hc-icons" fontIcon="hci-medical" hcIconLg></hc-icon>
            <div>Example Tile Header</div>
            <p>Example tile subheader description text</p>
</div>
.```

andrew-frueh:
@corykon - I added a usage doc, can you give it a quick read through and make any adjustments you think would be helpful?

jonest:
Just now seeing this and I don't even remember suggesting that, haha! Looks like it's already merged so I won't bother looking at the code

PR #1418, opened 2020-10-16
fix(ng-select theme): fix missing padding

### LINKED ISSUE #1417: Ng-select styles are broken
Something with the latest upgrade is causing problems with the default styles for selected items for our ng-select theme:

image

corykon:
fix #1417


andrew-frueh (approved):
thanks @corykon

PR #1425, opened 2020-10-27
1420 scrollnav active

### LINKED ISSUE #1420: ScrollNav Active should follow highlighted item on scroll
**Summary**

As reported by @mhollier - if an item in the scrollnav is selected, and the page is scrolled - the selected item will still retain its blue border.  The border is a feature for accessibility and keyboard navigation, so we don't want to lose it - we just want the item with the border to be updated to the currently highlighted item as the page scrolls.

**Reproduction**

Steps to reproduce:

1. Visit the scrollnav example.
2. Select an item on the scrollnav sidebar
3. Scroll down in the component so the highlight changes
4. Note that the border stays on the originally selected item (the border should be updated when the highlighted item changes on scroll)

**Additional context**

image (18)

andrew-frueh:
Gives scrollnav links focus when they are set to active - ensures the highlight border follows as the page is scrolled, but users can still keyboard navigate using the highlight to tab through links.

Also included two small fixes for search indexing two style pages I noticed needed ids on their headers.

closes #1420

### DIFF
--- a/projects/cashmere/src/lib/scroll-nav/nav/scroll-nav.component.ts
+++ b/projects/cashmere/src/lib/scroll-nav/nav/scroll-nav.component.ts
@@ -36,5 +36,6 @@ export class HcScrollNavComponent implements AfterViewInit {
             e.classList.remove(this.ACTIVE_CLASS);
         });
         element.classList.add(this.ACTIVE_CLASS);
+        element.focus();
     }
 }
--- a/projects/cashmere/src/lib/scroll-nav/nav/scroll-nav.component.ts
+++ b/projects/cashmere/src/lib/scroll-nav/nav/scroll-nav.component.ts
@@ -36,6 +36,6 @@ export class HcScrollNavComponent implements AfterViewInit {
             e.classList.remove(this.ACTIVE_CLASS);
         });
         element.classList.add(this.ACTIVE_CLASS);
-        element.focus();
+        element.blur();
     }
 }
--- a/projects/cashmere/src/lib/scroll-nav/nav/scroll-nav.component.ts
+++ b/projects/cashmere/src/lib/scroll-nav/nav/scroll-nav.component.ts
@@ -32,10 +32,14 @@ export class HcScrollNavComponent implements AfterViewInit {
     }
 
     private setActiveClass(element: HTMLElement): void {
+        const focusedEl = document.activeElement;
         this._links.forEach(e => {
             e.classList.remove(this.ACTIVE_CLASS);
+            // If a nav item on the scroll list currently has focus, remove it so a highlight border doesn't persist on scroll
+            if ( focusedEl === e ) {
+                e.blur();
+            }
         });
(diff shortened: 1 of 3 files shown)


projects/cashmere/src/lib/scroll-nav/nav/scroll-nav.component.ts
```diff
@@ -36,5 +36,6 @@ export class HcScrollNavComponent implements AfterViewInit {
             e.classList.remove(this.ACTIVE_CLASS);
         });
         element.classList.add(this.ACTIVE_CLASS);
+        element.focus();
```
corykon, line 39: As I think about this more, changing the focus on scroll makes me a little nervous. Could make for an odd UX in some cases. Like if the user was filling out form or interacting with controls in the content area, and then scrolled, I think the browser would steal focus away from the form and place it on the scroll nav.

Could we instead just make the focus outline less noticeable, like a light dashed or dotted line or something?

corykon: Or maybe even just call blur on the nav element when you scroll and the next or prev section needs to be activated.

andrew-frueh: I think calling blur on scroll would work for me.  I do think it's odd to keep the clicked item highlighted on scroll, and blur would resolve that.  And I'd guess that calling blur on the nav element wouldn't remove focus from a form element in the example you gave.

So I'd be good with that change if you'd feel better about that approach.

andrew-frueh: Made that update @corykon - I like this change.

corykon: Sweet yeah UX wise I that's prob safer. I'd recommend something like this for calling blur, so we're only firing off events when we need to.

```
private setActiveClass(element: HTMLElement): void {
        const focusedEl = document.activeElement;
        this._links.forEach(e => {
            e.classList.remove(this.ACTIVE_CLASS);
            if ( focusedEl === e) { e.blur(); }
        });
        element.classList.add(this.ACTIVE_CLASS);
    }
```

corykon (approved):

projects/cashmere/src/lib/scroll-nav/nav/scroll-nav.component.ts
andrew-frueh: Even better!  Thanks @corykon 👍

andrew-frueh:
Yikes, @cikeddy discovered that my original fix with simply adding focus to `element` was totally broken in Chrome (worked in Firefox).  So extra glad we went with your suggestion @corykon 😬  I have tested the blur version in both Chrome & Firefox though and it works as expected.

mhollier:
Thanks @corykon !

PR #1433, opened 2020-10-30
fix(modal): zIndex incrementing

### LINKED ISSUE #1426: Modal z-index increases
Since we switched to Cashmere modals we have noticed that occasionally we'll have z-index issues with modals where dropdowns in the modal have ended up hidden behind it.

Some sleuthing by Nick into modal.service revealed that there is this on line 133:
`        this._zIndexCounter += 2;`

As a consequence modals start with a z-index of 2000 and then it increases by two every time you open a modal.
Our dropdown has a z-index of 2500 so as soon as you've opened the modal 250 or so times the dropdown hides behind the modal.

We are wondering if this incrementing is necessary as we can only have one modal open at a time - or if it serves some other purpose?

andrew-frueh:
only increment if multiple modals are allowed; don't count up forever

closes #1426

### DIFF
--- a/projects/cashmere/src/lib/modal/modal.service.ts
+++ b/projects/cashmere/src/lib/modal/modal.service.ts
@@ -26,7 +26,8 @@ export class ModalService {
     allowMultiple: boolean = false;
 
     // start at 2000 (reserved range for modals, see _variables.scss)
-    private _zIndexCounter = 2000;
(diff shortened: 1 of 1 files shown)


Techgeekster (approved):

jacquie (approved):
Thank you for getting to this issue so quickly!
The fix looks good 👍

PR #1455, opened 2020-12-02
fix(drawer): add ignoreEscapeKey property

### LINKED ISSUE #1454: Unable to prevent Drawer component from closing when user presses Escape
**Summary**

Operations Console is using the Drawer component as a container for several filter controls. We want to prevent the Drawer from closing when the user presses the escape key, but are unable to do this. It would be convenient if there was an option similar to modal dialogs (e.g. `ignoreEscapeKey`) to specify whether or not closing on escape is desired.

**Additional context**

image

mhollier:
fixes #1454


andrew-frueh (approved):
Well I can't get that unit test to fail locally, and it looks like it succeeded in the second of the two Travis runs. So I think it must have been a gremlin on that first round.  Everything else looks great @mhollier - thanks for the fix!

PR #1460, opened 2020-12-04
fix(datepicker): time input bug

### LINKED ISSUE #1456: Time picker error does not allow user to manually type in time.
**Summary**

When manually typing a time in the time picker. When the input loses focus the input is cleared. Works on the example site due to some SugarJS magic. But in StackBlitz and other sites not using Sugar it does not work.

andrew-frueh:
Fixes bug on time-only input parsing. Great catch by Brent, I'm surprised this didn't crop up sooner.

The issue is that the Javascript Date object won't parse a time-only string.  It will parse a date without a time, but not the other way around.  So we have to have a placeholder date to include on time-only strings.  But it can't just be a random date because that could interfere with date comparisons on the date-range component.

So I added a placeholder value that takes the value of whatever the picker is set to, or the current date if none is set.

closes #1456

### DIFF
--- a/projects/cashmere/src/lib/datepicker/datepicker-input/datepicker-input.directive.ts
+++ b/projects/cashmere/src/lib/datepicker/datepicker-input/datepicker-input.directive.ts
@@ -112,6 +112,9 @@ export class DatepickerInputDirective implements ControlValueAccessor, OnDestroy
         this._value = value;
         this._formatValue(value);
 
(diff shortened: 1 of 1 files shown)


mhollier:
Thanks!

mhollier (approved):

jonest (commented):
👍

PR #1462, opened 2020-12-10
fix(tabset): tab descendants

andrew-frueh:
allow tabs to be contained in parent divs


joeskeen:
Does this change the behavior of when you have a tabset inside of a tab?

andrew-frueh:
Yikes!  That's a REALLY good catch @joeskeen - I think you're right about that being a problem.

What we're trying to allow here is for tabs to be wrapped in a parent div within a tabset.  Maybe there needs to be more checks to make sure its only tabs that aren't within another tabset.

andrew-frueh:
Scrapping this change, keeping track of what's a nested tab and what's not is going to be nightmare - thanks again for catching the bug this would have introduced @joeskeen

PR #1477, opened 2021-01-13
User/heather.bloyer/add abbrev list

heathkat:
Hi Andrew!
Here's the abbreviations list for the style guide on Cashmere. 
I wasn't able to get multiple filters to work together. The search filter currently works but the categories filter does not. Will you have a look if you get time? If you don't have time we can just take that second filter out for now. It's not critical at this point. Thanks!


andrew-frueh:
Thanks Heather!  I'm going to close this PR - I'll merge your branch into mine for the site update.

PR #1478, opened 2021-01-13
Null in boolean checker

jacquie:
This adds a check for null or undefined to the `ParseBooleanAttribute `function in `utils.ts`
I did check where it is being used to make sure that passing null and returning false would always be the correct result.
I was thinking in particular there could be an issue where the default for an attribute is true eg in the Accordion where the default for `toolbarTrigger` is `true` but I think all will be well.


corykon (approved):
Thanks Jacquie! I like the addition :+1:

I'm a little confused about some other stuff going on in this function... like why should an empty string return true? What happens if a value is passed in that is of a type that isn't referenced specifically, like an object or a number? Maybe those are problems for another day though!

andrew-frueh:
@corykon regarding an empty string returning true - wouldn't that allow a component to include a parameter without a value and have it work?  I've seen that done in some places - like `<hc-menu-drawer align="left" ignoreEscapeKey>`.  Am I reading that wrong, or is that likely why the empty string returns true?

corykon:
@andrew-frueh I think you're right! That's news to me. Had no idea ignoreEscapeKey without a value would evaluate to an empty string in your example above. Makes total sense.

jacquie:
I made a new PR to replace this one which was based on the wrong branch :(

PR #1480, opened 2021-01-14
fix(utils.ts): parseBooleanAttribute fix

### LINKED ISSUE #1474: utils.parseBooleanAttribute causes error if null is passed to it.
**Summary**
utils.parseBooleanAttribute() does not take care of a null value being passed to it and will throw an error because it's trying to call `toLowerCase` on the null value.

andrew-frueh:
add check for null to parseBooleanAttribute

closes #1474


jacquie (approved):
Looks good!

PR #1507, opened 2021-02-02
fix(hcicons): hc-icons import

andrew-frueh:
corrected getting started and stylesheet fix

### DIFF
--- a/guides/getting-started.md
+++ b/guides/getting-started.md
@@ -92,14 +92,10 @@ In `angular.json`, place references for each font in your project's build target
 ```
 
 ### C) *(Optional)* Add Health Catalyst icons (`hcicons`).
-If your app needs access to Health Catalyst & healthcare-specific icons, perform the following:
+If your app needs access to the [Health Catalyst icon font](/foundations/icons), add the following to your app's global style sheet (`src/styles.scss`):
 
-1. Download `hcicons` [here](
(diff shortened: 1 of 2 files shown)


projects/cashmere/src/lib/icon-font/hcicons.scss
```diff
@@ -1,3 +1,5 @@
+$hc-icons-font-path: "" !default;
```
corykon, line 1: Found that this path by needed to be ".", because the font is a sibling to this stylesheet within node_modules

guides/getting-started.md
```diff
@@ -92,14 +92,10 @@ In `angular.json`, place references for each font in your project's build target
 ```
```
corykon, line 92: A note on a couple previous section while we're in here:

--------

Step 2B's code example should look like this:

```
@NgModule({
     imports: [CashmereModule],
     exports: [CommonModule, CashmereModule]
})
export class SharedModule {}
```


-----
Step 3C could be changed to be something like below. Important tweak to node_modules path- should be single "./" rather than "../" since node_modules is usually a sibling to angular.json

```
"styles": [
                "./node_modules/font-awesome/css/font-awesome.css",
                "./node_modules/notosans-fontface/css/notosans-fontface-allweight.css",
                 // your app's global stylesheet
                "src/styles.scss"
            ]
```

corykon (approved):
Couple things I ran into while troubleshooting w/ matt

PR #1512, opened 2021-02-06
Scroll Nav - enable [attr.hcScrollLink] for dynamic loading

Techgeekster:


andrew-frueh (changes requested):
Such great work @Techgeekster - special thank you for all the additional unit tests, very impressive stuff.  I've just got one really minor request before we merge.  Since `nativeElement` and `setDirectiveToNode` are public - they're going to show up in the API docs.  That's fine, we should just add a comment above them so they have descriptions in the docs.  Or if you think the are primary for internal library use only, you can add an underscore in front of the name and the docs generator will ignore it (`_nativeElement`).  Either way is fine with me.  Once we make that change this looks good to merge.
Screen Shot 2021-02-08 at 9 51 48 AM

Techgeekster:
> Such great work @Techgeekster - special thank you for all the additional unit tests, very impressive stuff. I've just got one really minor request before we merge. Since `nativeElement` and `setDirectiveToNode` are public - they're going to show up in the API docs. That's fine, we should just add a comment above them so they have descriptions in the docs. Or if you think the are primary for internal library use only, you can add an underscore in front of the name and the docs generator will ignore it (`_nativeElement`). Either way is fine with me. Once we make that change this looks good to merge.
> Screen Shot 2021-02-08 at 9 51 48 AM

Updated with your suggestions

andrew-frueh (approved):

PR #1516, opened 2021-02-09
docs: add feedback form and http data getter

heathkat:
I created a feedback form component and hooked it up to a google sheet and tested it (Andrew, you specified you wanted the data to be captured differently and brought into the EDW but I'm not sure how to do that so I did what I know how to do .... the google sheet can be integrated with a slack channel for easy tracking)

I switched the data source to a Google sheet so that the style guide list can be updated more easily and without requiring code changes each time. 

I fixed the validation errors when the edit list form is reset. Still not exactly the behavior we want but getting better. I'll keep improving it.


andrew-frueh (approved):
Nice work @heathkat - I made a few tweaks and cleanups but I think this looks really good.  One style change I made from the docs site is to remove the border and the question below the bottom tile.  I think we want this feedback element to be pretty secondary to the content on the page - and be visible, but not distracting.  So this styling is my take on it - but I'd love to hear what @corykon thinks.

I think this PR is the right way to start, and then we can decide where we'd like to add this element more globally on the site.

Screen Shot 2021-02-14 at 2 37 23 PM

corykon (approved):
That'll work great. If I were to change anything, it would be to move the yes/no buttons and the question closer together, tighten up that connection.

andrew-frueh:
Yeah that's a good call @corykon - I'm gonna make that change

andrew-frueh:
Definitely like the elements centered next to each other better @corykon 

Screen Shot 2021-02-16 at 11 21 06 AM

PR #1520, opened 2021-02-12
Scroll nav: min-height fix, set active first link on init only, enable nav scroll with content

Techgeekster:
fix min-height on last element if last element is dynamic. set active link on 1st link only if on init, enable scroll of nav links with content


andrew-frueh (approved):
Wow I'm very intrigued by `CdkScrollable` - would love to hear more about that.

PR #1525, opened 2021-02-16
docs(content): add edit buttons to usage items

heathkat:
Here's another go at my last PR. Hopefully this one will work better! 

I added edit buttons to each row in the abbreviations list. Clicking the button takes you to the add/change form and autopopulates some of the data for the form. 

I also changed the form validation a bit so that it now validates when you hit submit rather than having a disabled submit button and ever-present validation errors.


andrew-frueh (approved):
Great improvements @heathkat!  I'll wait till you validate that I didn't break anything in my commits, but otherwise looks good.

PR #1528, opened 2021-02-17
Scroll-Nav: fix for  min-height, dynamic links, auto-scroll

Techgeekster:


andrew-frueh (approved):
Just tested locally and fix looks good @Techgeekster

PR #1534, opened 2021-02-17
fix(pagination): pageSize select update fix

### LINKED ISSUE #1533: Pagination pageSize doesn't update select dropdown
**Summary**

When the `pageSize` parameter is changed via code, it doesn't update the select dropdown

**Reproduction**

Steps to reproduce:

1. Set a pageSize value
2. Change that value after init either with a button or with a timeout
3. The number values displayed in the associated table will change, but the select dropdown doesn't update

Stackblitz reproduction:
https://jwn2ac--run.stackblitz.io

andrew-frueh:
fixes bug where programatically changing pageSize didn't update select

closes #1533

### DIFF
--- a/projects/cashmere/src/lib/pagination/pagination.component.html
+++ b/projects/cashmere/src/lib/pagination/pagination.component.html
@@ -1,7 +1,7 @@
 <div class="hc-pagination-container">
     <div *ngIf="!hidePageSize" class="hc-font-sm hc-page-detail-container">
         <label class="hc-page-detail-label">Showing:</label>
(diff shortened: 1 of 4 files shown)


andrew-frueh:
@corykon let me know if you have any concerns about this fix.  Basically it looks like `[value]` on the pagination select wasn't updating reliably when the pageSize parameter was changed.  I know Angular is trying to move everything to FormControls anyway, so this seemed like the right switch.  But let me know if you have any concerns.

smoovebb (approved):

corykon (approved):

Issue #1549, opened 2021-03-04
Text on the Multiselect component is not readable for selected items


cikeddy:
**Summary**

The text color on the 'MULTISELECT CUSTOM TEMPLATES' of the Multiselect & Typeahead component does not change to white when selected, which makes it difficult to read.

**Reproduction**

Steps to reproduce:

1. Go to [MULTISELECT CUSTOM TEMPLATES](https://cashmere.healthcatalyst.net/web/components/multiselect/examples/#multiselect-custom-templates)
2. Select an item from the dropdown

Stackblitz reproduction (click the "Edit on Stackblitz" link on a component example and fork the template):
N/A

**Additional context**
This is what a selected item looks like:
![image](https://user-images.githubusercontent.com/46426253/109993108-f4b23800-7cd9-11eb-87db-236781aa072a.png)

The text on the 'MULTISELECT EMBED CHECKBOX' does change as expected:
![image](https://user-images.githubusercontent.com/46426253/109993143-fb40af80-7cd9-11eb-9fb3-2b1d6d82ce11.png)

andrew-frueh:
Nice catch @cikeddy! @corykon - this looks like a pretty quick fix, you okay to take this one?

PR #1554, opened 2021-03-12
docs(ng-select example): fix selected styles

corykon:
re #1549


andrew-frueh (approved):
Love those one line fixes!

PR #1555, opened 2021-03-12
User/yaroslavb/ie 11 support

hassankhan986:
Continuation of work from https://github.com/HealthCatalyst/Fabric.Cashmere/pull/1550


hassankhan986:
outstanding issue :
image

hassankhan986:
on running npm test , get the following error, it is not cashmere specific but to do with typescript package compatibility etc..
image

Issue #1538, opened 2021-02-23
Modal - Mobile Support


andrew-frueh:
**Describe your idea**

When at mobile screen sizes, the modal should go full screen for a better experience (rather than displaying as a popup for desktop resolutions). @cikeddy and Alejo have done testing on prototypes and determined with users that this the optimal layout for mobile.

**What problem are you trying to solve?**

Inset floating modals aren't great UX on a small screen

**Additional context**

This is Cassie's modal in a desktop environment:
<img width="1105" alt="Screen Shot 2021-02-23 at 11 24 04 AM" src="https://user-images.githubusercontent.com/22795893/108890072-57188380-75ca-11eb-990d-12450a35aff9.png">

...and this is how it should appear when on a small screen:
<img width="388" alt="Screen Shot 2021-02-23 at 11 23 52 AM" src="https://user-images.githubusercontent.com/22795893/108890107-64ce0900-75ca-11eb-9217-e2ae26cb5f0a.png">

andrew-frueh:
@jacquie - you might be interested in this as well as Leading Wisely is focusing more on mobile.

jacquie:
Yes we are starting to work on better responsiveness right now and are definitely interested in this

cikeddy:
Thought about this more and it might be good to have full screen as an option that devs can set rather than a constant on mobile. Based on what I've seen, there are certain cases (usually complex tasks) where we want fullscreen modals, like the search  example above, and some where we'll want a regular overlays ([https://www.appcues.com/blog/mobile-app-modal-windows](examples)).

andrew-frueh:
closed in #1561

PR #1561, opened 2021-03-26
feat(modal): responsive full screen modal

corinnen:
re #1538
Modals will take up the full screen width and height when opened in a small browser (excluding small modals)
image


andrew-frueh (approved):

andrew-frueh:
@corinnen will add an issue to add a parameter that makes the responsive functionality something that can be turned off, but we'll get this merged as it is for now

PR #1569, opened 2021-04-01
Enterprise/cd

### LINKED ISSUE #1562: HcDataTableSource has a search bug
HCDataTableSource returns some unexpected results when using the default filterPredicate function.

LW noticed this on an admin screen. Searching for the string "12" brought back a record that didn't make sense, until digging in to the source code and @AdamGMiller pieced this together:
image

 It looks like the source code munges together all data properties into one string with spaces separating them, which can do some wacky things.  I'd recommend we just separate things with spaces here-
https://github.com/HealthCatalyst/Fabric.Cashmere/blob/f66bd9de4647b5c1f31e260bd8655c17d0fbdfb9/projects/cashmere/src/lib/table/table-data-source.ts#L206

WistyDagel:
Fixed both the issue #1562 and added unit tests to prove it now works

closes #1562

### DIFF
--- a/projects/cashmere/src/lib/table/table-data-source.spec.ts
+++ b/projects/cashmere/src/lib/table/table-data-source.spec.ts
@@ -141,37 +141,37 @@ describe('HcTableDataSource', () => {
                 ].toString()
             );
         });
+    });
 
-        describe('filterPredicate()', () => {
-            it('filters through the table by its id', () => {
-                let filter = "1";
-                const result = dataSource.filterPredicate(ELEMENT_DATA[0], filter);
-                expect(result).toBe(true);
-            });
+    describe('filterPredicate()', () => {
+        it('filters through the table by its id', () => {
+            let filter = "1";
+            const result = dataSource.filterPredicate({...ELEMENT_DATA[0]}, filter);
+            expect(result).toBe(true);
+        });
 
-            it('filters through the table by its name', () => {
-                let filter = "Hydrogen";
-                const result = dataSource.filterPredicate(ELEMENT_DATA[0], filter);
-                expect(result).toBe(true);
-            });
+        it('filters through the table by its name', () => {
+            let filter = "Hydrogen";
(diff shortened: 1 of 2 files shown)


andrew-frueh:
@WistyDagel @MarielaCisneros - Finally got back to this at the end of the day, and wow, what a deceptively tricky problem to solve on these unit tests!  Looks like all checks are passing now so I think we're good, and sometimes these unit tests really do feel like voodoo.  I can't say with 100% certainty what was going on here, but my best guess is that these tests are sometimes running at the same time, and because your tests were all pointing to objects in the same array - they might have been coming back with empty data if everything wasn't synced up.

The hardest part about debugging this was just that it was so unpredictable, so I had to do things like stop/start `npm run test`.  Or close down VS Code and run the test fresh.  But that definitely made this problem way harder.

At one point I tried just hard coding the object into the first field of the filterPredicate function, and that seemed to work consistently.  But that's pretty ugly to leave it that way, so I ended up cloning the object using the [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) and that seems to get around any memory collisions that were happening.

Like I said, not 100% sure that's why it's working now - but that's my best guess.

Also wanted to point out that I moved all the new tests into a separate describe called `filterPredicate()`.  That way when Jasmine reports back - the line would read something like:
> HcTableDataSource filterPredicate() filters through the table by its name

Give the updated code a look through before standup tomorrow and let's discuss before we merge this one.

andrew-frueh:
Wow, nope.  After I posted that comment I decided to run the test one more time locally, and it failed again.  It looks like even cloning the object is still problematic b/c it's pointing to that Array too.  So it seems like the only bulletproof way to get the tests to pass EVERY time is to hard-code the test object into the filterPredicate function.  I haven't been able to get it to fail that way 🤞

andrew-frueh:
Oh boy and now it's failing linting because my line is too long with that object.  We gave you guys a great bug to start on 😆

andrew-frueh (approved):

PR #1572, opened 2021-04-05
fix(typography): unordered list

### LINKED ISSUE #1571: Fix hanging indent for unordered list items in docs styling
In Cashmere docs, unordered list items do not have a hanging indent like ordered list items do. This should be fixed in the style settings. 

image

heathkat:
In scss: - adds hanging indent to uls to match ols. Adjusts space between li's and between number and li.
In docs: adds reference usage content.

closes #1571


andrew-frueh (changes requested):
Thanks Heather!  So the changes you made in `styles.scss` are specifically for the Cashmere site.  Which is fine, and maybe we want to keep those - but even better would be if this is reflected in the Cashmere library.  Take a look at the file `/projects/cashmere/src/lib/sass/_typography.scss`.  That's where the actual Cashmere styles live, and you'll see in there that @cedar-ave has already done some pretty cool stuff with nested list indenting (line 177).  But they need to be in an `<article>` tag to get those styles.  So you'll notice in the markdown files for the personas - I very often drop an `<article></article>` tag around the Goals or Needs sections so I can take advantage of Marla's nested lists.

So my recommendation here would probably be to roll back your changes to styles.scss, and see if you can build this fix into _typography.scss, and then wrap your lists in the markdown in an <article> tag.

Let me know if you have any questions.

andrew-frueh:
@corykon - I'm totally fine with making changes to the list styles in `style.scss`.  I'd err on the side of pulling as much _out_ of there as possible and getting it into `typography.scss`.  Do you have any specific tweaks to the app styles.scss for Heather to make as a part of this PR?

corykon:
@andrew-frueh  Sorry I inserted myself here. I saw your comment on the PR after I'd added my comment on the issue (#1571). Was trying to cue up some simple stuff for the Neumont folks. 

I'm cool with the approach you've outlined- introducing a new opinion into the library and ripping out the wonky cashmere site styles, rather than tweaking the cashmere site styles directly.

PR #1575, opened 2021-04-06
feat(date-range): the dev can now center the date-range component

### LINKED ISSUE #1570: Unable to set position of DateRange
**Summary**

When trying to get the DateRange modal to not be pinned to the button I just pushed, I am unable to do this. It appears that the modal is pinned to the button and is not currently able to be moved within the window. Our Date filter button is on the left hand side of our window and would like the date range modal to move to the middle of the screen.

There currently is a popperPlacement parameter on the DateRange modal, but it does not control anything. Any value still places the modal just below the button.

Maybe we could have an option for `position` that allows us to unpin it from the button and move it in the window where we need.

**Reproduction**

Steps to reproduce:

1.

**Additional context**

WistyDagel:
date-range component can either be attached to the button or centered on the screen now. the unused
popperplacement param has been removed as well.

closes #1570


andrew-frueh (changes requested):
Really really nice work on this one Chris & Mariela!  Only one small requested tweak to the copy added to that DateRange example.  Let's change the sentence to:

>In this example, the center value is set to `true` in `DateRangeOptions`, which centers the component to the screen, instead of anchoring to the target.

andrew-frueh (approved):
Good job adding the `<code>` tags on that example as well 👍

PR #1579, opened 2021-04-07
docs(markdown): markdown parser fix

andrew-frueh:


corykon (approved):
I think this'll break a couple tests in markdown.directive.spec.ts. 

We might want to also update the type of the `markdownContent` property in markdown-content.component.ts

Good to go after that I think

PR #1586, opened 2021-04-09
docs(markdown): article tag wrapper for lists

### LINKED ISSUE #1571: Fix hanging indent for unordered list items in docs styling

andrew-frueh:
ol and ul lists in markdown files are now wrapped in an article tag

@heathkat I had a little time this morning so I went ahead and took care of this.  Such a great suggestion - look at how much unnecessary code this removes (568 lines!).  And it really improves the formatting on many of the developer guides that had lists but weren't previously wrapped in article tags.  So a huge thank you for calling this out - and again to @cedar-ave for building these styles into the library.

@kei02003 - wanted to make sure you were aware of this fix now, because we were using `<article>` tags around all our lists in the Analytics markdown files.  But now that article tag will be automatically added, so one less thing to remember.  I took care of removing all the existing article tags.

closes #1571

### DIFF
--- a/guides/analytics/powerbi/about.md
+++ b/guides/analytics/powerbi/about.md
@@ -44,10 +44,8 @@ The Sidebar vertical text box is used on the About page for app details and shou
 The first two items in the sidebar should be the name of the accelerator and the current version number.
 You should be able to simply edit the text in the template.
 
-The remaining items in the sidebar are unique to your application. Card visuals are used to display items that rely upon data fields (such as a last refresh date). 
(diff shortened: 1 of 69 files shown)


corykon (approved):

heathkat:
I think this looks good @andrew-frueh. Thank you for doing the work. Do you know why the list styling doesn't get applied on the trademarks page? Also, what do you think about adding a few pixels of padding between list items for when there are multiple multi-line list items? 
image

image

PR #1591, opened 2021-04-12
feat(create component): how to create a new component

### LINKED ISSUE #1442: Guide for creating a new Component
I've been working on a component this week, and even though I've done it several times, I forgot some of the steps you need to take to get a new component set up in the library. There's a guide for creating new examples (including an npm command you an run to do most of it for you), but not one for creating a component from scratch.

Bare minimum would like to see a simple checklist of things to do. Here's a rough outline-

**1- Create new folder in `projects/cashmere/src/lib`**
- Should contain at least one .ts file w/ any needed components, directives, services, styles, tests, etc. (Provide link here to guide about using mixins, following css standards)
- Should contain `module.ts` file that exports things that need to be public
- Should contain `index.ts` file that exports things that need to public

**2- Import & export your new module in `src/app/shared/cashmere.module.ts`**
**3- Export your component's `index.ts` file from `projects/cashmere/src/public_api.ts`**
**4- Create examples (see https://cashmere.healthcatalyst.net/guides/working-with-examples)**

----------------------------
Highly related to https://github.com/HealthCatalyst/Fabric.Cashmere/issues/879, but this something that could be added more quickly in the meantime.

MarielaCisneros:
this guide show two different ways to create a new component

fix #1442

### DIFF
--- a/guides/creating-new-components.md
+++ b/guides/creating-new-components.md
@@ -1,0 +1,76 @@
+# Creating a New Component
+
+###### Last updated April 12, 2021
+
+:::
+
+##### Overview
+
+The Cashmere library is heavily built on it's component-based architecture. By adding new and exciting components to this library you are only enriching what it has to offer to it's developers. Here are some things to remember when creating new components.
+
+:::
+
+:::
+
+##### Creating the Component Manually
+
+<!-- We will be using our banner component as an example of how to create a new component. -->
+
+1. Start by creating a new folder for your component in ```projects/cashmere/src/lib```
+
+This folder should contain the following files:
+
+* "component-name".directive.ts
+
+- "component-name".component.spec.ts
+
+* "component-name".component.ts
+
+- "component-name".html
+
+* "component-name".module.ts - this file exports all the public properties of your component.
+
+- "component-name".scss - this will be used to style your component. For more on how to follow our styling conventions, [click here](https://cashmere.healthcatalyst.net/web/guides/using-customizing-components?section=style-guidelines-for-new-components&selected=using-customizing-components)
+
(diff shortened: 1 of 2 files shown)


andrew-frueh (approved):
This looks great @MarielaCisneros & @WistyDagel - well written and appreciate how you split the steps into manual or automated.

guides/creating-new-components.md
```diff
@@ -0,0 +1,76 @@
...
+
+- "component-name".component.spec.ts
+
+* "component-name".component.ts
+
+- "component-name".html
+
+* "component-name".module.ts - this file exports all the public properties of your component.
+
+- "component-name".scss - this will be used to style your component. For more on how to follow our styling conventions, [click here](http://localhost:4200/foundations/color)
```
corykon, line 33: Could we link here instead? https://cashmere.healthcatalyst.net/web/guides/using-customizing-components?section=style-guidelines-for-new-components&selected=using-customizing-components

This explains the pattern we recommend using sass mixins instead of css rules directly.

corykon (approved):
Love it. Just have one small change, then good to merge. Thanks!

andrew-frueh:
@MarielaCisneros - just want to give you a heads up on a git change I made to this branch.  On the commit message, for changes like this that only change the site (not the Cashmere library), make sure you use the `docs` prefix.  That way when the build runs, it knows to only update the site - not create a new npm package with a new version number for the library.  You can `feat` on your commits, which would have made a new minor version of the library which we don't need to do in this case.

So I went ahead and squashed your commits into new one with the `docs` prefix.  And with that, I think this PR is good to merge

PR #1595, opened 2021-04-15
Create trademarks.md

kadchrist:
Moved CORUS from unregistered to registered marks.


andrew-frueh (approved):

andrew-frueh:
Nice job making the update @kadchrist - I've got some other updates to the site I need to push live now too.  So I'll start the build and this change should be live in the next hour or so.

PR #1603, opened 2021-04-19
fix(navbar): calculate link widths after page load

### LINKED ISSUE #1473: Navbar - "More" button disappears when cache cleared and page refreshed at narrow width
Operations Console is experiencing a bug in the Navbar component. The Navbar buttons get cut off and the "More" button does not appear. We were also able to reproduce this on the Cashmere demo site.

**Reproduction**

Steps to reproduce:

1. Reduce screen width so that there is not enough room for all Navbar buttons and the "More" button appears.
2. Clear the browser cache for the page and refresh the page (Empty Cache and Hard Reload in Chrome).
3. Notice that the buttons are now cut off and the "More" button is not there.

I talked with @andrew-frueh about this and he suggested it could have to do with line 148 in the [navbar.component.ts](https://github.com/HealthCatalyst/Fabric.Cashmere/blob/780a8e235d7177c319fbb6a611e73b1a1f79c5a0/projects/cashmere/src/lib/navbar/navbar.component.ts#L148) code here.

navbar-more-issue

andrew-frueh:
remove dependency on setTimeout, listen to `window:load` event

@corykon @MarielaCisneros @WistyDagel - inspiration struck today after we were talking about this, and I think I've got a fix that removes the `setTimeout` dependency.  Unfortunately, Angular's lifecycle hooks don't seem to be getting us what we want in this instance, but the standard [Window load event](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event) does.  It's fired after the page is fully rendered, which is what we need to get a proper width calculation on all the link elements.

So I _think_ this works...at least everything I could think to test seemed to check out.  Can you all give it a try as well?  And @pheebner - not sure if you can test it out before the fix is merged, but would love to hear your thoughts as well.

Note that I removed one unit test that was no longer useful since we aren't using the Angular lifecycle hook anymore.

closes #1473

### DIFF
--- a/projects/cashmere/src/lib/navbar/navbar.component.spec.ts
+++ b/projects/cashmere/src/lib/navbar/navbar.component.spec.ts
@@ -59,15 +59,6 @@ describe('NavbarComponent', () => {
         });
     });
 
-    describe('on calling all the lifecycle hooks', () => {
-        it('should call refreshNavLinks', waitForAsync(() => {
-            spyOn(testHostComponent.navbarComponent, 'refreshNavLinks');
-            testHostComponent.navbarComponent.ngAfterViewInit();
-            testHostFixture.whenStable().then(() => {
-                expect(testHostComponent.navbarComponent.refreshNavLinks).toHaveBeenCalled();
-            });
-        }));
-    });
     describe('on calling refreshNavLinks', () => {
         describe('and adjust the elements according to the navbar size', () => {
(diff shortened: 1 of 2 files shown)


corykon (approved):
This great, feels way better than a timeout.

WistyDagel (approved):
Ran through this version of the code and it seems like a way better option compared to the original timeout version. Great idea!

PR #1606, opened 2021-04-26
Modalredesign

WistyDagel:
When setting the parameter "isDraggable" to true, the modal can now be dragged across the window.


andrew-frueh:
@WistyDagel - looks like we ended up with some extra commits in this PR that we probably don't want to get into `dev`.  What I'm going to do is making a new branch off of dev and run`git cherry-pick fa2af72` to just pull in the last of your commits.  I'll add my style tweaks as an additional commit and push up a new PR.

Issue #647, opened 2018-11-28
Draggable and resizable modals


corykon:
**Describe the problem this feature request is solving**
We use modals in our app that are draggable and resizable. Probably won't be able to switch to cashmere modals until this functionality is added, or there's a way to plugin this type of functionality.

**Describe the solution you'd like**
New parameters, `isDraggable` to make the modal draggable by its title bar, and `isResizable` to make it resizable from the bottom right corner.

**Describe alternatives you've considered**
Find this a way to plugin this functionality on my own without forcing this functionality in Cashmere... but seems like lots of apps could benefit from this work.

**If any, which components will have to change, and how**
Modal component. No breaking changes, just new functionality.

**Additional context**
Screenshot from our app, draggable by title, resizable by bottom right corner.
![image](https://user-images.githubusercontent.com/12416432/49185562-7fe43c80-f31f-11e8-839c-d5c81cc37548.png)

corykon:
This may be a prime candidate for a bit component- I've been using a 3rd party library for resize/move functionality in popbuilder

PR #1608, opened 2021-04-27
Modal draggable

andrew-frueh:
Half of the #647 ticket adding the option for hc-modal to be draggable.


projects/cashmere/src/lib/modal/modal-window.component.ts
```diff
@@ -6,7 +6,11 @@ import {ActiveModal} from './active-modal';
 @Component({
     selector: 'hc-modal-window',
     template: `
-        <div [class]="'hc-modal hc-modal-' + _size"><ng-content></ng-content></div>
+        <hc-modal-drag [disableDrag]="_isDraggable">
```
andrew-frueh, line 9: @corykon - I want to talk this template over with you before a merge.  I have a feeling there is a more elegant way to do this that doesn't include the `hc-modal-drag` tag at all if `_isDraggable` is false.  The tough part is that it's the parent, so it's tricky to use `ngIf` here.  Also because if you put two `ng-content` blocks in an if/else setup, Angular doesn't know which one to actually put the content in. So what we've got here works - I don't love the extra tag where it isn't needed.  If you've got any Angular tricks up your sleeve, let me know.

corykon: No angular tricks coming to mind... I think my preferred approach would have been to avoid the `hc-modal-drag` component completely, and just directly use cdkDrag in this template, maybe something like this:
```
<div [class]="'hc-modal hc-modal-' + _size" cdkDrag [cdkDragDisabled]="!_isDraggable">
     <span [hidden]="!_isDraggable" class="hc-modal-drag-handle></span>
     <ng-content></ng-content>
</div>
```

andrew-frueh: Ya know - while I was doing the cleanup I didn't even take a step back to consider revising to do it without the component, I bet that wouldn't be too hard to change.  I'll give it a look this afternoon.

andrew-frueh:
@corykon - simplified this one some, and realized as I was in there that we almost broke the new responsive functionality from @corinnen.  So I've adjusted that and draggable modals work fine until the `md` breakpoint where they switch to fullscreen and drag handle disappears.  Let me know if you have any other concerns/suggestions.

corykon (approved):
Looks perfect to me :+1:

WistyDagel (approved):
Love all the revisions you have made! Thank you for taking the time to improve the modal drag even further.

Issue #956, opened 2019-08-22
Refresh Code Block Styles


corykon:
**Describe your idea**
Any chance we could refresh the styles for code blocks? Obviously not a high priority item, but I found myself wanting to tweak things. 

I didn't in my suggestions below, but you could probably snag some nice Cashmere/Catalyst colors to plug in.

![image](https://user-images.githubusercontent.com/12416432/63553723-914a2600-c4f8-11e9-9b40-4442f8a27e94.png)

PR #1609, opened 2021-04-28
fix(code styles): refresh code block styles

corykon:
re #956 Decided against doing a dark mode. It would've been difficult to show it without more intensive changes to the way we display markdown.

Changes are pretty minimal, but code blocks are cleaner and now use cashmere colors. Before on the left, and after on the right

codestyles1

codestyles2


andrew-frueh (approved):
WOW!  This is great @corykon - I didn't even know you were planning on this.  Looks great and really appreciate it

PR #1613, opened 2021-04-30
Button toggle fix

### LINKED ISSUE #1604: Button Toggle throws error if `[selected]` is tied to model
This is a use case I'm working on in Leading Wisely. I wanted to use the button toggle, and need to have its state tied to a property in my class, `isGridView`, but keep getting 'Expression changed after checked' errors. I wrestled for awhile, tried to follow the example as closely as I could. See stackblitz below.
https://stackblitz.com/edit/btntoggle-issue?file=src/app/button-toggle/button-toggle-example.component.html

Seems like the only way to use the component is to hard code the [selected] @Input with true or false, but for this use case it makes a lot more sense to tie it to a property in my class.

I wish I could do something like this and have it just work:

```
<hc-button-toggle-group [formControl]="isGridView">
  <hc-button-toggle [value]="true">Grid</hc-button-toggle>
  <hc-button-toggle [value]="false">List</hc-button-toggle>
</hc-button-toggle-group>
```

```
export class IsGridViewComponent {
    ...
    readonly isGridView = new FormControl(true);
    ...
}
```

corykon:
fix #1604 

Now possible to bind button toggle group component directly to a model, while still supporting the original implementation.


andrew-frueh (approved):

MarielaCisneros (approved):

WistyDagel (approved):
Everything looks pretty clear and straightforward! I'd say the biggest thing that we had missed was that "providers" section... so no wonder nothing was working right. Thanks for diving in and getting this to work Cory!

PR #1615, opened 2021-05-03
fix(unit tests): modal unit tests

### LINKED ISSUE #576: Add unit tests for the Modal components
Add unit tests for the Modal components:

- modal
- modal-header
- modal-body
- modal-footer
- modal-overlay
- modal-window

MarielaCisneros:
create some unit tests for all modal components.

closes #576


corykon (approved):
Tests look good! Only change would be to move those utils that were copied from datePicker into a more general place so both components can reference them.

Maybe to `projects/cashmere/src/lib/test-utils`

corykon:
Thanks for moving those utils :+1: I think that's a better location. I would say there probably some more cleanup in order for those... for example, there are two different functions hanging around that are both for parsing booleans. But I think that can be handled in a separate issue. Agree/disagree @andrew-frueh?

Could you also take a look at the linting errors (if you haven't already)?

andrew-frueh (changes requested):
yes, we want to get rid of the `boolean-property.ts` file and leverage the `parseBooleanAttribute` in `utils.ts`

andrew-frueh (approved):

projects/cashmere/src/lib/modal/modal.spec.ts
```diff
@@ -30,24 +37,164 @@ describe('hc-modal', () => {
...
+    it('should the overlay click be disable', () => {
+        windowComponent._ignoreOverlayClick = true;
+        windowFixture.detectChanges();
+        expect(windowComponent._ignoreOverlayClick).toBeTruthy();
     });
 
     it('should change size depending on options', () => {
         let div: DebugElement = windowFixture.debugElement.query(By.css('.hc-modal'));
         expect(div.properties['className']).toContain('hc-modal-auto');
+        // Small
```
Techgeekster, line 76: Each of these size tests should be in their own test.  You can surround the tests with a general description by using 'describe', like this:

```
describe('should change size', () => {
     it('when window is small'), () => {
          ...
     });

     it('when window is medium'), () => {
          ...
     });
});
```

Techgeekster (commented):
I'm concerned with the file structure of the tests.
modal.spec.ts contains the tests for ALL modal components, and I feel like each component should have it's own spec file so you are only testing what is on that component.
Testing all the components inside modal.spec.ts creates a complexity that could cause issues down the road.

The file structure should be similar to this:
modal-body.component.ts
modal-body.component.spec.ts
modal-footer.component.ts
modal-footer.component.spec.ts
modal-header.component.ts
modal-header.component.spec.ts
... and so on

@andrew-frueh @corykon What do you guys think?

projects/cashmere/src/lib/modal/modal.spec.ts
andrew-frueh: Great suggestions @Techgeekster - thank you!

projects/cashmere/src/lib/modal/modal-header.component.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+        }).compileComponents();
+    });
+
+    beforeEach(() => {
+        headerFixture = TestBed.createComponent(ModalHeaderComponent);
+        headerComponent = headerFixture.componentInstance;
+        headerFixture.detectChanges();
+    });
+
+    it('should create an footer component', () => {
```
andrew-frueh, line 28: Just a really minor fix here @MarielaCisneros - this should read: 'should create a header component'

PR #1617, opened 2021-05-05
fix(picklist btn): fixes submit btn issue

### LINKED ISSUE #1605: Picklist: Clicking "Select All" will trigger submit() on an enclosing form
See repro here: https://stackblitz.com/edit/picklist-form-submit-bug?file=src%2Fapp%2Fpicklist-simple%2Fpicklist-simple-example.component.html

Should be easy to fix. I _**think**_ we might be able to just set `type="button"` on the "Select All" btn.

WistyDagel:
the select all button no longer submits the response right away

closes #1605


corykon (approved):
Perfecto

PR #1620, opened 2021-05-06
fix(data range): optional parameter

### LINKED ISSUE #1618: DateRange open function breaking change
**Summary**

The recently revised `open` function on DateRange introduced an untended breaking change with the addition of the second `center` property:

```open(hostElemRef: ElementRef<any>, center: boolean)```

It should default to false and be an optional parameter (not required).

MarielaCisneros:
change the center property, instead of being default to false now is an optional parameter

closes #1618


andrew-frueh (approved):
Cool, that was easy @MarielaCisneros :-)

Looks like the `center` value just gets passed down a couple functions until it hits `if (center)` which works with either `false` or `null`.  So that worked out great.  No compiler errors so I think we're good to go.  Thanks!

PR #1624, opened 2021-05-11
1472 eslint

corykon:
- Installed eslint
- Removed tslint
- Some intitial changes per new linter


andrew-frueh:
Cory will wrap this one up

PR #1625, opened 2021-05-12
docs(content): personas product index

heathkat:
adds Personas by Product index at https://cashmere.healthcatalyst.net/content/products


corykon (approved):
Good with all the content/code changes. Just need to take care of a couple linting issues and this is good to go-
image

andrew-frueh:
Great start @heathkat!  If you don't mind, I'm going to make a couple tweaks before we merge this branch.  I want to add a little additional logic so that when you click on a persona from a product summary - the back button above the persona returns you to the summary.  I'll work on that tomorrow and add a commit to this PR.

andrew-frueh (approved):
@heathkat - think I've got this one ready to go.  I adjusted the persona viewer so that it'll jump back to the product persona summary when you click the back button in the top left.  Only thing to keep in mind is all your links in the product summary markdown files should end with `?referrer=healthcare-ai` where the value is the route of the product persona summary.

One annoying thing is that markdown files won't allow you to do `routerLink` on anchors.  It only renders them as normal `href` params on the anchor - which reloads the app when you click the link.  There might be a way to fix that in our markdown renderer, but that's a job for another day.

heathkat:
Thank you @andrew-frueh! Yes, I noticed that about routerLink not working in markdown. I started trying to fix it and then decided it was a bigger deal than I could figure out quickly.

PR #1626, opened 2021-05-17
fix(mobile ux): components no longer zoom on mobile

### LINKED ISSUE #1602: Mobile Zooming on Inputs
**Summary**

**Reproduction**

Steps to reproduce:

2. Click any input field

**Additional context**

WistyDagel:
closes #1602

### DIFF
--- a/projects/cashmere/src/lib/sass/input.scss
+++ b/projects/cashmere/src/lib/sass/input.scss
@@ -1,4 +1,5 @@
 @import './colors';
+@import './mixins';
 @import './form-field';
 
 @mixin hc-input() {
@@ -37,6 +38,11 @@
 
     &::-ms-input-placeholder {
         color: $slate-gray-300;
+    }
+
+    //Change input font size for mobile users
+    @include media-breakpoint-down(md) {
+        font-size: 16px;
     }
 }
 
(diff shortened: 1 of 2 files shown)


corykon (approved):

andrew-frueh:
@corykon - any concern that these breakpoints increase the font size not only for mobile devices for _any_ smallish browser window?  Like if I'm on a laptop but have a smaller window, suddenly the fonts on my inputs will be way bigger.  Should we be looking for adding logic here to ensure this change happens only on mobile devices?

corykon:
That’s a good point. What exactly is the breakpoint we’re using, and do we
expect desktop browser windows to actually get that small?

Getting logic just right for detecting whether the device is actually
“mobile” is tricky, and usually requires some JavaScript.

We could just keep the functionality out and instead include the
information somewhere in a guide for app teams to make the best decision
for their use case.

corykon:
@andrew-frueh Sounds like we want to abandon this PR? If anything, the right solution might just be a guide explaining how to implement something similar if needed for a particular app's use case.

andrew-frueh:
Agreed @corykon - I'm debating whether we should add a param called `mobile` (kind of like `tight`) that renders the input with the larger font size.  Having two separate params isn't great though - would be better to have something like `fieldStyle` with values of either `normal`, `tight`, or `mobile`.  And we could keep the `tight` param as something deprecated for a few releases.  Let's discuss tomorrow.

PR #1627, opened 2021-05-17
feat(modal): the component can now be resized

### LINKED ISSUE #647: Draggable and resizable modals

WistyDagel:
When setting the isResiable property to true, the modal can be resized by the user.

closes #647

### DIFF
--- a/projects/cashmere-examples/src/lib/modal-overview/modal-overview-example-modal.component.html
+++ b/projects/cashmere-examples/src/lib/modal-overview/modal-overview-example-modal.component.html
@@ -7,6 +7,6 @@
     <hc-modal-body>Data: {{ activeModal.data }}</hc-modal-body>
     <hc-modal-footer>
         <button hc-button buttonStyle="link" (click)="cancel()">Cancel</button>
-        <button hc-button buttonStyle="primary" (click)="close()">Ok</button>
+        <button hc-button buttonStyle="primary" (click)="close()">OK</button>
     </hc-modal-footer>
 </hc-modal>
--- a/projects/cashmere-examples/src/lib/modal-overview/modal-overview-example.component.html
+++ b/projects/cashmere-examples/src/lib/modal-overview/modal-overview-example.component.html
@@ -13,5 +13,7 @@
     Open Draggable Modal from Component
 </button>
 <button hc-button buttonStyle="secondary" (click)="openTemplate(content)" class="demo-button">
-    Open Auto-Sized Modal from Template
+    Open Resizable <!-- Auto-Sized -->Modal from Template
 </button>
+<br/><br/>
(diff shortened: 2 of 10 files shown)


corykon:
Code wise looks like a good approach! I am seeing a linting issue failing the build:
image

Tip to try and avoid those kind of build failures - It's a good idea to run `npm run lint` before you check in so can catch those ahead of time.

corykon (approved):

andrew-frueh (changes requested):
More nice work here @WistyDagel - finally got to pull this down and play around with it.  A few things I'd like to discuss and see if we can clean up for a merge on this one:

1. Would be ideal if we could add some kind of constraints on the resizing, so that it can't be shrunk past a minimum width.  Right now you can shrink it down to a teeny tiny modal that kind of feels like a mistake.
2. As the modal is resized it - will no longer be centered on the screen.  Maybe our best bet here is to _only_ allow the modal to be resized vertically.  That will help with item one above and will also eliminate any centering issues.  That's actually what GitHub does with this input text box I'm typing in now.
3. I'm pretty sure I remember you saying that resize and draggable don't play nicely with each other.  Wondering if we can brainstorm a way to get them both working - if jQuery UI can do it, I bet we can too :-)  And if not, let's make sure it fails gracefully and only allows one to be set at a time (so nothing is broken).  And let's make sure the documentation explains if you can only use one or the other.

corykon:
@andrew-frueh My thoughts on your 3 suggestions-
1. Totally agree. We have min-widths on resizable modals in PB
2. Losing centering hasn't bothered me on PB, and if it does, you can just drag it back to the center. 
3. Really would be a bummer to not have the two work together. I'd support pressing pause on this CSS approach to see if we can find a javascript based method to make things work the way really want. Even if we have to fork/borrow/steal from the jquery codebase

andrew-frueh:
@corykon - agreed on your second point, provided that dragging & resizing can work together :-) if they can't, then you just end up with a really big off-center modal.  So yeah, I'm with you on item number 3, let's crank on this one a little more and see if we can get both drag and resize on the same modal.  Then I won't be so worried about the off-center problem.

andrew-frueh (changes requested):
@WistyDagel looks like you've got one small linting error and a few failing unit tests (related to changing css classes).  Looking forward to reviewing live today!

andrew-frueh:
Updated PR coming with corrected dev branch merge

PR #1630, opened 2021-05-21
fix(datepicker): fixed the minutes max/min

### LINKED ISSUE #1628: DATEPICKER - Cannot set MINUTES time to exactly '00' using the up/down arrows
**Summary**

On the DatePicker control, cannot set the MINUTES time to exactly '00' using the up/down arrows.  For example '3:00 PM'

**Reproduction**

image

WistyDagel:
closes #1628

### DIFF
--- a/projects/cashmere/src/lib/datepicker/calendar/calendar.component.ts
+++ b/projects/cashmere/src/lib/datepicker/calendar/calendar.component.ts
@@ -567,7 +567,7 @@ export class CalendarComponent implements AfterContentInit, AfterViewChecked, On
             let curMin = +this.minutes;
             curMin++;
             if (curMin > 59) {
-                curMin = 59;
+                curMin = 1;
             }
             this.minutes = curMin.toString();
         }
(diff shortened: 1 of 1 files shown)


corykon (approved):

projects/cashmere/src/lib/datepicker/calendar/calendar.component.ts
```diff
@@ -567,7 +567,7 @@ export class CalendarComponent implements AfterContentInit, AfterViewChecked, On
             let curMin = +this.minutes;
             curMin++;
             if (curMin > 59) {
-                curMin = 59;
+                curMin = 1;
```
andrew-frueh, line 570: hey @WistyDagel - this should actually get set to 0 instead of 1 so that 3:59 loops to 3:00 instead of 3:01

andrew-frueh (changes requested):
One fix here and then good to go

andrew-frueh (approved):

PR #1632, opened 2021-06-02
fix(mobile zooming inputs): change the font size for mobile view

### LINKED ISSUE #1602: Mobile Zooming on Inputs

MarielaCisneros:
"fix #1602"

### DIFF
--- a/projects/cashmere-examples/src/lib/banner-overview/banner-overview-example.component.html
+++ b/projects/cashmere-examples/src/lib/banner-overview/banner-overview-example.component.html
@@ -21,7 +21,7 @@
 <div class="example-content">
     <hc-form-field>
         <hc-label>Notification banner type:</hc-label>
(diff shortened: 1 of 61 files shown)


corykon:
The points I mentioned during our standup:
- rather than having two differently named properties that do the same thing (`fieldstyle` and `selectStyle`), lets just name them both `fieldStyle`. That way devs don't have to memorize different properties depending on the component they're using at the time.
- Make sure we are actually sticking our generated class name onto the html element. For example, in select.component.html, do something like `class="hc-select-input {{selectStyle}}"`
- Because `fieldStyle` should default to 'normal', we don't need to explicitly set it. That means most example files don't actually need to have any changes in this PR.

andrew-frueh:
@MarielaCisneros @WistyDagel - thanks for giving this one a go.  I'll close this PR for now and Cory and I will pick up what you started in the branch and finish it up.

Issue #1635, opened 2021-06-03
Table sort icons can get really small


njwhc:
**Summary**

If the column headers in a table are long and wrap to multiple lines, the sort icons shrink.

![image](https://user-images.githubusercontent.com/30876456/120705272-da246200-c474-11eb-99f9-6e964cba1735.png)

**Reproduction**

[Stackblitz reproduction](https://stackblitz.com/edit/angular-7aamd8?file=src%2Fstyles.scss)

Steps to reproduce:

1. Make a table that uses `hc-sort-header`
2. Use long values for column headers

**Workaround / Potential Change**

In the application I work on, we have added a style override to our Cashmere tables to force the icons to be a consistent size.

```css
/* This is added to styles.scss in the above linked Stackblitz, uncomment it to experiment. */
.hc-sort-header-icon {
  min-width: 20px !important;
  margin-left: 5px !important;
  align-self: flex-end;
  margin-bottom: 4px;
}
```

As we're talking CSS, there are numerous other ways to get sort icon size consistency. For instance, applying either `display: table` or `position: fixed` to `.hc-sort-header-icon` will both yield consistently sized icons (perhaps with some margin adjustments).

However, the above override also aligns the icons to the bottom of the column header row:

![image](https://user-images.githubusercontent.com/30876456/120705708-66cf2000-c475-11eb-8712-0534ef0f7eef.png)

As far as I know, we have not observed any negative side effects from the above override. It may be worth considering this change in Cashmere itself.

corykon:
I'd support the change. 👍 Size of the sorting icons should never change - and I'm sure that wasn't the original intention.

PR #1636, opened 2021-06-03
fix(sort): make sort icon consistent size

njwhc:
Fixes https://github.com/HealthCatalyst/Fabric.Cashmere/issues/1635


andrew-frueh (approved):
Thanks so much for the fix @njwhc

njwhc:
> Thanks so much for the fix @njwhc

You're most welcome. Though, to be fair, this isn't really my fix. @jacquie actually came up with the override this is based upon. I'm just here to steal the credit. 😄

PR #1637, opened 2021-06-03
fix(sort-icon): no longer shrinks

### LINKED ISSUE #1635: Table sort icons can get really small

WistyDagel:
The header-icon no longer shrinks depending on the size of the table. It is now at a fixed width.

closes #1635

Old version:
image

New Version: 
image

### DIFF
--- a/projects/cashmere/src/lib/sass/sort.scss
+++ b/projects/cashmere/src/lib/sass/sort.scss
@@ -63,8 +63,11 @@
     background-repeat: no-repeat;
     display: block;
     height: 13px;
-    margin-left: 15px;
     width: 13px;
+    min-width: 20px !important;
+    margin-left: 5px !important;
+    align-self: flex-end;
+    margin-bottom: 4px;
 }
 
 @mixin hc-sort-header-asc() {


andrew-frueh:
Duplicated PR of #1636

PR #1638, opened 2021-06-09
feat(modal): resizable modals

### LINKED ISSUE #647: Draggable and resizable modals

andrew-frueh:
adds resizable functionality to modals that works with draggable

closes #647

### DIFF
--- a/projects/cashmere/src/lib/modal/modal-window.component.spec.ts
+++ b/projects/cashmere/src/lib/modal/modal-window.component.spec.ts
@@ -42,42 +42,6 @@ describe('hc-modal-window', () => {
         expect(windowComponent._ignoreOverlayClick).toBeTruthy();
     });
 
-    describe('should change size', () => {
-
-        it('when window is small', () => {
-            let div: DebugElement = windowFixture.debugElement.query(By.css('.hc-modal'));
-            expect(div.properties['className']).toContain('hc-modal-auto');
-            windowComponent._size = 'sm';
-            windowFixture.detectChanges();
-            expect(div.properties['className']).toContain('hc-modal-sm');
-        });
-
-        it('when window is medium', () => {
-            let div: DebugElement = windowFixture.debugElement.query(By.css('.hc-modal'));
-            expect(div.properties['className']).toContain('hc-modal-auto');
-            windowComponent._size = 'md';
-            windowFixture.detectChanges();
-            expect(div.properties['className']).toContain('hc-modal-md');
-        });
-
-        it('when window is large', () => {
-            let div: DebugElement = windowFixture.debugElement.query(By.css('.hc-modal'));
-            expect(div.properties['className']).toContain('hc-modal-auto');
-            windowComponent._size = 'lg';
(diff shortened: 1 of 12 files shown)


projects/cashmere/src/lib/sass/modal.scss
```diff
@@ -207,3 +212,10 @@ $modal-border: 1px solid $gray-200;
     justify-content: center;
 
 }
+
+@mixin hc-modal-resizable {
+    resize: both;
+    overflow: hidden;
+    min-height: 300px;
```
corykon, line 219: This seems pretty tall for a min height. I've got a modal as short 170px in PB

projects/cashmere/src/lib/modal/modal-window.component.spec.ts
```diff
@@ -42,42 +42,6 @@ describe('hc-modal-window', () => {
         expect(windowComponent._ignoreOverlayClick).toBeTruthy();
     });
 
-    describe('should change size', () => {
```
corykon, line 45: Bummer to rip out tests that could probably be fixed. But I'm not enough of a stickler to let this prevent a merge

corykon (approved):
Looks good to me. Couple minor things I called out

projects/cashmere/src/lib/modal/modal-window.component.spec.ts
andrew-frueh: Ugh, I know I thought that same thing - these tests are still relevant and probably shouldn't be removed.  They would just have to be engineered another way.  Maybe you and I can try to figure out a fix to get them back in this afternoon in our 1:1.

projects/cashmere/src/lib/sass/modal.scss
andrew-frueh: @corykon I took a look into this one and header/footer both have `min-heights` of 55px and body has a `min-height` of 120px.  So adding those up I brought the resizable min height down to 230px.

projects/cashmere/src/lib/modal/modal-window.component.spec.ts
andrew-frueh: Had to start a new set of unit tests for the modal service which didn't exist previously, but allowed me to test the size and resizable options.  Probably some additional coverage could be added on that one, but at least it's a start - and we haven't lost the coverage we had before.

corykon (approved):
Looks good. Like the revised min-height approach :+1:

PR #1651, opened 2021-07-21
fix(modal): don't force implementers to type modals

corykon:
_pretty_ sure this should fix the annoyance that Mark and Brad have run into


andrew-frueh (approved):
Ah there it is :-)  Thanks for digging this up.

corykon:
> Ah there it is :-) Thanks for digging this up.

That was all @mhollier!

mhollier:
Thanks!

Issue #1643, opened 2021-07-09
Navbar directly reference font awesome


corykon:
More menu uses `fa-chevron-down`. Lets base64 encode that.
![image](https://user-images.githubusercontent.com/12416432/125125536-f4c2aa00-e0b6-11eb-92d2-82aca3d2ca11.png)

PR #1663, opened 2021-08-20
fix(icons): base64 a bunch of icons

corykon:
re #1643

found direct references to font awesome icons in the navbar, old picklist, date pickers, and app switcher. They're now base64 encoded, to avoid the requirement that implementing apps install font awesome.


jacquie (approved):
Looks good - thanks for taking care of these last icons!

PR #1712, opened 2021-11-08
feat(input): mobile param

### LINKED ISSUE #1602: Mobile Zooming on Inputs

andrew-frueh:
flag to increase input font size to prevent zooming on mobile devices

@corykon @aaronbrethorst - added the param we were discussing to the input component.  Also reworked one of the examples on the input page.  We had a redundant one for input suffix, so I made it into an example that demos the different styles available on input.  Let me know if you have any suggestions.

closes #1602


projects/cashmere/src/lib/sass/form-field.scss
```diff
@@ -145,6 +145,10 @@ $wrapper-padding-bottom: ($error-margin-top + $line-height) * $error-font-scale;
     padding: $infix-padding 10px;
 }
 
+@mixin hc-form-field-infix-mobile() {
+    font-size: calculateRem(16px);
```
corykon, line 149: I was a bit confused when I first saw `calculateRem()`. I've been using this mixin everywhere for font size rems: 
`@include fontSize(16px)`. 

Turns out fontSize() mixin just uses calculateRem() under the covers. Not totally sure why we two versions in the library to do the same thing.

So I guess nothing really needs to be changed here... but should figure out why both exist and why we'd use one over the other.

corykon (approved):

projects/cashmere/src/lib/sass/form-field.scss
andrew-frueh: Hah, good point - I've mostly used `fontSize` as well.  In this case I just did a copy/paste from what was being used elsewhere in this stylesheet.  But pretty sure those two functions are totally redundant like you said.

aaronbrethorst:
Sorry I didn't respond sooner. lgtm!

PR #1717, opened 2021-11-16
Target Angular 12

corykon:
We're due for an upgrade to angular 12. Tackling this before I start on other v12 issues so we have as much time as possible before the release to catch any more issues.

I've tested all of the components and page, seems to be working well now.

**For posterity sake - a summary of some of the changes preformed:**
- Angular and cli to v 12.2.13 
     - https://update.angular.io/?l=3&v=11.0-12.0
     - npx @angular/cli@12 update @angular/core@12 @angular/cli@12 
- CDK to v 12.2.13
     - Need to add more providers to our HcTable for it to work
- Updated ng-select to v7
- To fix compile errors w/ raw-loader, installed raw-loader 0.5.1 (an older version that seems to force IVY to play nicely)
- Use Math.div() for sass to avoid warnings about deprecated division
- Menu and popover sass files now have forked overlay styles from cdk rather than importing from '~@angular/cdk/overlay' because it would intermittently fail to resolve it.
-  Ran npm audit to take care of a good chunk of vulnerability issues


mhollier:
Nice work @corykon! Do you happen to know if there are other alternatives to using raw-loader? It looks like it is deprecated, at least according to this https://v4.webpack.js.org/loaders/raw-loader/.

corykon:
> Nice work @corykon! Do you happen to know if there are other alternatives to using raw-loader? It looks like it is deprecated, at least according to this https://v4.webpack.js.org/loaders/raw-loader/.

Yeah that link you posted mentions [asset modules](https://webpack.js.org/guides/asset-modules/), but I decided that might need to be a battle for another day. Trick will be to figure out how to tweak webpack to configure that stuff when all of that is hidden behind the angular cli.

andrew-frueh (approved):

PR #1739, opened 2021-11-22
fix(navbar): active supports strings

### LINKED ISSUE #1673: Active @Input() on navbar doesn't work
Leading Wisely team attempted to use this `[active]` property, but it doesn't seem to work. Not sure if it's supposed to work alongside uri property. 

Will work on putting together a StackBlitz repro later.
image

andrew-frueh:
allow the active property to accept string as well as boolean values


corykon (approved):

andrew-frueh:
Just logging a note here that I tried everything I could think of the recreate the SlackBlitz error that @corykon created in Cashmere and I can't do it.  On the Cashmere site, `active` seems to work as expected, overriding the router state.

PR #1742, opened 2021-11-23
Changes to support building in Ivy

corykon:
There's more changes to the build to come, but this will at least fix the builds on the azure pipeline.

Current known issue - a second `npm run build` will always fail. In order to run a full successful build, you'll need to wipe out node_modules and rerun `npm i`.


andrew-frueh (approved):
Assuming you and Joe saw the failed test.  I won't even try and wrap my head around all you guys have sorted through here

corykon:
@andrew-frueh Shoot... no we hadn't rerun tests. I look at those.

PR #1745, opened 2021-11-29
fix(navbar): fix `active` input

### LINKED ISSUE #1673: Active @Input() on navbar doesn't work

corykon:
fix #1673

Took some battling, but eventually got it to repro and was able find a fix.

Solution was to use 'force-active' class instead of just 'active', because we were using an 'active' class for the built in route-based underlining and sometimes it was causing conflicts.


andrew-frueh (approved):
Ah okay that explains why I was struggling to reproduce it.  Fix looks good, and I'll go ahead and merge the other little update I made to allow for active as a string or boolean

PR #1747, opened 2021-11-30
feat(modal): modal focus upgrades

### LINKED ISSUE #1645: Apply focus back to trigger for modals
**Describe your idea**
For our modals, it would be awesome to have a setting that allows us to automatically apply focus back to the tigger (e.g. button) that opens it once the modal is closed. 

This helps direct all users back to where they left off in the experience by providing a visual indication of the trigger and is better for accessibility (keyboard navigation).

andrew-frueh:
Added focus trap to open modal and add options to restore or auto focus. Great suggestion on this one @cikeddy!

@corykon - I followed the pattern you established with hcPop almost exactly on this one, including the `restoreFocus` and `autoFocus` options. As best as I can tell, it _should_ work even with multiple modals. But if you can give this one a pretty thorough review, I'd appreciate it - want to make sure I didn't accidentally introduce any bugs.

closes #1645


corykon (approved):
Looks and works great.

PR #1749, opened 2021-12-01
fix(popover): update after reposition

### LINKED ISSUE #1647: Tooltip Positioning gets stuck
Tooltip positioning won't change from what it's initially set to unless you also change something else - like the popover trigger.

Repro video:
tooltipIssue

### LINKED ISSUE #1721: Allow popovers to open from given x/y coordinates
This requirement was discovered while working with the PopInsights team. They had a case where a context menu that is opened by a function call needed to open at specific x/y coordinates.

This could be accomplished with the following:
1. Allow for the popover's `open()` function to accept x/y coordinates to specify the location to display.
2. (Bonus) Allow for a popover w/ no anchor (just opened at a given location)

### LINKED ISSUE #1748: Make UI Colors the default page when clicking foundations?
This is possibly a silly one... 

The number one thing I do when coming to cashmere is to look at the list of color variables. I do it several times a day. I go to cashmere home, click on "Foundations", then click on "UI Colors". Every time I do it I think "man we should default to UI colors instead of brand colors".

Could we make it so UI Colors is the default, save a click? I'm not sure what determines the order of those links... but if it's alphabetical sort, we could change to "Application UI Colors".

Maybe its just me, and maybe I should just bookmark the ui colors page. But we might be able to tweak this even faster that it took me to write this issue...

andrew-frueh:
force the popover to update after a reposition event

@corykon the issue here (which you already called out in a comment), is that the popover needs to get destroyed for overlay positioning changes to take effect.  Fortunately you already had the code in place to do that, so just needed to connect the reposition event.

I also reordered the popover examples while I was in there.

closes #1647
closes #1748
closes #1721

### DIFF
--- a/projects/cashmere/src/lib/pop/popover-anchoring.service.ts
+++ b/projects/cashmere/src/lib/pop/popover-anchoring.service.ts
@@ -3,6 +3,7 @@ import {ElementRef, Injectable, NgZone, OnDestroy, Optional, ViewContainerRef} f
 import {
     ConnectionPositionPair,
     FlexibleConnectedPositionStrategy,
+    GlobalPositionStrategy,
     HorizontalConnectionPos,
     Overlay,
     OverlayConfig,
@@ -75,6 +76,9 @@ export class HcPopoverAnchoringService implements OnDestroy {
 
     /** Emits when the service is destroyed. */
     private _onDestroy = new Subject<void>();
+
+    /** Stores the fixed position coordinates */
+    _fixedPos: number[] = [];
 
     constructor(private _overlay: Overlay, private _ngZone: NgZone, @Optional() private _dir: Directionality) {}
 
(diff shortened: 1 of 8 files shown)


andrew-frueh:
@corykon I got the fixed position support added.  It was MUCH easier doing it the way we discussed - where the popover is just anchored to body or some container with no trigger.

projects/cashmere/src/lib/pop/popover-anchoring.service.ts
```diff
@@ -252,9 +267,8 @@ export class HcPopoverAnchoringService implements OnDestroy {
                 case NotificationAction.TOGGLE:
                     this.togglePopover();
                     break;
-                case NotificationAction.REPOSITION:
-                    break;
                 // TODO: When the overlay's position can be dynamically changed, do not destroy
+                case NotificationAction.REPOSITION:
```
corykon, line 271: Curious what this change was needed for

corykon (approved):
Awesome work, glad we're getting this in there.

projects/cashmere/src/lib/pop/popover-anchoring.service.ts
andrew-frueh: Sorry I mashed together multiple tickets in this PR.  That fix you highlighted was for #1647.  The only way for those positioning changes to take effect is to make sure the tooltip gets destroyed and reinitialized when a position change happens.

corykon: Ah got it! Thanks

Issue #1750, opened 2021-12-04
Remove Bit


corykon:
**Summary**

To build cashmere, we have a big long build script in our package.json. It looks like this:
`npm run clean && npm run release-mode && npm run build:lib && npm run build:bit && npm run build:examples && npm run search-prep && npm run build:user`

Since upgrading to angular 12 (which requires compiling w/ Ivy), this npm build script will only ever succeed the first time you run it after wiping out node_modules and reinstalling dependencies (`npm i`). We've found that the errors occur only after running the `npm run build:bit` step. If you skip the build:bit step during the second time through, everything works swimmingly over and over.

This can be worked around, so for now, we are leaving the main build script as it is and adding some notes where needed in the [documentation](https://cashmere.healthcatalyst.net/web/guides/contribution-guide). We've also added a `build:dev` script in case you need to rerun the full build and don't want to go through the trouble of reinstalling dependencies.

**To do**
1. Fix the build. To do this, we can try one of the following:
     - [Upgrade from Bit Legacy to Bit Harmony](https://harmony-docs.bit.dev/reference/migration/). Harmony _should_ be more compatible w/ newer versions of angular.
     - Leave Bit as is on the legacy version, ang just find a way for the `npm run build` command to clean up after itself (looking at you IVY compiler...) so that it always succeeds. 
2. Remove the temporary `build:dev` script from package.json.
3. Get rid of the note about current build issue in the [Guidelines for Contribution](https://cashmere.healthcatalyst.net/web/guides/contribution-guide)

corykon:
For now, we've decided to rip out bit rather than going through the process of upgrading our system to Bit Harmony. It's an awesome and promising tool, but the deprecated version is causing problems, and we've not had instances come up where Bit was used or needed. 

If/when the need does arise, then we'll look into setting up Bit Harmony.

PR #1751, opened 2021-12-04
build(ivy): setup workarounds for known build issues

corykon:
`npm run build` still fails the second time you run it, so this PR sets up a workaround. As needed, devs can use `npm run build:dev` while working on contributions and avoid the issues caused by the Bit build.

An issue to really fix the build issue was added here- #1750

### DIFF
--- a/guides/contribution-guide.md
+++ b/guides/contribution-guide.md
@@ -6,7 +6,7 @@
 
 ##### All May Contribute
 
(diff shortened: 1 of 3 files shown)


andrew-frueh (approved):
Made one tiny typo fix, but otherwise looks great.  Let's wrap this up and get this out the door!  Let me know when the release is finished and we're ready to announce.

PR #1754, opened 2021-12-07
Examples cashmere package

andrew-frueh:
For some reason without this change, `npm i` kept pulling 11.3 of Cashmere in examples instead of 12.1


andrew-frueh:
Tested this one again, and for some reason `npm i` in the root still wants to pull version 11.3 of Cashmere into the examples project.  Not sure if there is a lag or something on npm since 12 just got released today.  Give it a look and check what it's doing for you.  If we do merge this change, we'll have to make sure and update the dependency every time we have a major number release.

corykon:
@andrew-frueh are we thinking we no longer need this?

andrew-frueh:
I'm not sure why `*` is still pulling 11.3 for me instead of 12.1, but I'm thinking maybe that's just a glich in my env.  We'll see.  I'll close for now.

PR #1759, opened 2021-12-08
build(sass): reverting change from "/" to math.div

### LINKED ISSUE #1757: StackBlitz Broken in v12
StackBlitz is saying `math.div` is an undefined function:

Screen Shot 2021-12-07 at 7 50 00 PM

...also see this related error in the build process:

Screen Shot 2021-12-07 at 7 49 28 PM

corykon:
I had changed us over to `math.div()` as called for by sass. https://sass-lang.com/documentation/breaking-changes/slash-div

`Math.div()` isn't playing nicely with stackblitz when imported by a library, so I switched back. We'll get warnings again with this when building... but shouldn't break anything.

fix #1757

### DIFF
--- a/projects/cashmere/src/lib/sass/_functions.scss
+++ b/projects/cashmere/src/lib/sass/_functions.scss
@@ -1,4 +1,3 @@
-@use 'sass:math';
 @import './variables';
 
 /// Slightly lighten a color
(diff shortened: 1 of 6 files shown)


andrew-frueh (approved):
Ah okay it looked like it had something to do with the packager turning math.div back into / or something on the compiler error.  Guess we'll just adjust this when we go to 13.

PR #1762, opened 2021-12-11
fix(stackblitz): enable Ivy and fix bit module refs for stackblitz

corykon:
Hallelujah!

Started with a working Ivy-based stackblitz example, and kept copying stuff back and forth until I tracked down the changes needed. Found that Ivy wasn't enabled in the tsconfig file, and stackblitz didn't like the way were trying to import our bit module.

Also made a change to not hide debug tools by default. I find them useful and I'm sure other devs will, too.


andrew-frueh (approved):
Wow wow wow, I had no idea you were working on this until I logged in this morning!  Thank you SO much for getting this figured out.  I think we're FINALLY ready to announce the new release today :-D I'll get this merged and get the build started.

PR #1772, opened 2021-12-29
fix(popover): allow right-clicks to close popovers

### LINKED ISSUE #1390: Context Menu: Right click elsewhere should close menu
I expected that the first context menu wold close and a second one would open when I right clicked a second time. A bug, or just different but ok behavior?
context-menu-ux

I'll admit though, I don't hate that I'm able to get to the built in browser context menu.

andrew-frueh:
@corykon @aaronbrethorst - do you think I need to remove the event listener I'm adding here?  When the popover closes, the CDK overlay element that it's attached to is destroyed. Wasn't sure if that takes care of the garbage collection for me or if we still need to removeEventListener.

closes #1390

### DIFF
--- a/projects/cashmere/src/lib/pop/popover-anchoring.service.ts
+++ b/projects/cashmere/src/lib/pop/popover-anchoring.service.ts
@@ -295,7 +295,7 @@ export class HcPopoverAnchoringService implements OnDestroy {
             .subscribe(() => this.closePopover());
 
         // Allow right-clicks to close an open popover
-        if ( this._overlayRef.backdropElement ) {
+        if ( this._overlayRef.backdropElement && this._popover.interactiveClose ) {
             this._overlayRef.backdropElement.addEventListener( 'contextmenu', event => {
                 event.preventDefault();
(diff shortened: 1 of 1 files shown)


aaronbrethorst:
@andrew-frueh That's a level of subtlety of the JS event handler lifecycle that I am not confident answering. I was actually dealing with something very similar yesterday and ended up including a `removeEventListener` call just to be safe—although I am not clear if it was actually necessary...

corykon:
@andrew-frueh Memory leak stuff is very confusing - I would think you'd need the removeEventListener in this case!

corykon (approved):

andrew-frueh:
Thanks @aaronbrethorst and @corykon - I agree w/ you both, better to err on the safe side.  I'll get a `removeEventListener` added just to make sure there isn't a zombie listener floating around in memory.

andrew-frueh:
Well I'm glad I gave this a second look, because I realized there is a more Angular way of event listening from Typescript using `Renderer2` instead of javascript's `addEventListener`.  It definitely makes the garbage collection easier.

PR #1777, opened 2021-12-31
feat(app-switcher): appClick output emitter

### LINKED ISSUE #1535: App Switcher - fire output events when an application is selected
**Describe your idea**

It would be helpful to know what application a user has selected from the App Switcher. For example, in Operations Console, we would like to send an event to Application Insights for tracking.

andrew-frueh:
add appClick to app-switcher and move component to separate page in docs

@corykon - as I was working on the request from @mhollier I realized that app switcher doesn't have it's own page in the docs, it was just lumped in with the usage page on Navbar.  With adding this feature, it seemed like it really should have it's own page, and it goes nicely with the environment switcher component page you built.  However when I did that I discovered that a lot of the code for app switcher didn't have comments for the api docs, so I had to fix that up.

Anyway, the `appClick` output is a nice addition as per Mark's suggestion, and I think this is a good improvement to the library documentation.

closes #1535


mhollier (approved):

mhollier:
Nice work. Thanks Andrew!

PR #1780, opened 2022-01-11
feat(date-range): closed emitter

### LINKED ISSUE #1776: Date Range: emit event on cancel
On Pop Insights dashboards for performance reasons we want to suspend loading the dashboard data when the date range picker is opened.
Currently there is no way to know if the date range picker gets closed with the cancel button with no new date range picked and so in our case no way to let the component know to continue loading the dashboard data with the current date range.

It would be good if Cashmere could emit an event when the date range picker is canceled.

andrew-frueh:
adds closed emitter to fire when picker is dismissed or saved

closes #1776


andrew-frueh:
Looks like have some failing tests.  Will correct...

corykon (approved):
:+1:

jacquie (approved):
Thanks for adding this - looks good 👍

PR #1790, opened 2022-01-18
feat(toasts): custom image

### LINKED ISSUE #1253: Custom Icon / Image on Toast Messages
**Describe your idea**

Toast messages can only be of type `'success' | 'warning' | 'info' | 'alert'`, which covers many use cases, but sometimes you may want to display a different icon or image to the user, for example:

image

Currently any custom icon requires you to create a new custom toast template, which can lead to a lot of duplicate code to get what you want and make it look like the normal toast messages (not have it look a little off).

If the toast message allowed an optional `image` property that would allow a custom image or icon to be passed in, it would cover a lot more use cases.

**Additional context**

One challenge we are going to run into with this is how to actually pass that data in.  Right now toast messages are created using code like this:

```ts
this.toasterService.addToast({
  type: 'success',
  header: 'Success',
  body: 'The operation was a success'
})
```

What would that look like with a custom image?

```ts
this.toasterService.addToast({
  type: 'info',
  header: `New Message from ${message.fromUser}`,
  body: message.body,
  image: message.avatarUrl
})
```

How about a custom icon?

```ts
this.toasterService.addToast({
  type: 'success',
  header: `Success`,
  body: 'Item was added to your Favorites',
  image: { fontSet: 'fa', fontIcon: 'fa-star' }
})
```

Does this look good from an API standpoint, or do we want to handle it differently?

andrew-frueh:
ability to set custom images / icons on standard toasts

@corykon @joeskeen - making the custom image 30 x 30px like the icons felt too small to me.  So for custom images I made it 60x60px instead.  That work?  Any real benefit to giving the dev another param to set that image size?

closes #1253

### DIFF
--- a/projects/cashmere-examples/src/lib/toaster-overview/toaster-overview-example.component.html
+++ b/projects/cashmere-examples/src/lib/toaster-overview/toaster-overview-example.component.html
@@ -69,6 +69,16 @@
         <input hcInput type="number" [formControl]="progressValue" [disabled]="toastProgress.value!='2'" />
     </hc-form-field>
 </div>
+<div class="progress-row">
+    <hc-form-field>
+        <hc-label>Custom Image:</hc-label>
+        <hc-radio-group [formControl]="customImage" inline="true">
+            <hc-radio-button [value]="null">None</hc-radio-button>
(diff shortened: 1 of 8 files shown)


corykon (approved):
Looks good to me. I don't have strong opinions on image size, trust your judgement there 👍

joeskeen (approved):

PR #1791, opened 2022-01-21
docs(feedback): app insights

andrew-frueh:
add app insights tracking hook for feedback form


heathkat (approved):
So this will track feedback form submissions in App Insights? Cool!

corykon (approved):

andrew-frueh:
@heathkat - yep, this means that every time a feedback form is submitted a hook gets fired to put that data in App Insights.  However we'll still keep your google doc integration b/c it's connected to Slack.  The main reason for the app insights addition is that I'm getting a Leading Wisely dashboard built to visualize the Cashmere analytics, and I wanted to make sure feedback form submissions are a data point we can include.

Issue #1786, opened 2022-01-12
Add banner for upcoming browser support change


corykon:
Decision has been made to drop support for Internet Explorer. This will happen in cashmere when we move to Angular v13 (#1714) in May, and then become official for Health Catalyst as a whole on June 15th, 2022. There is a blog post in the works where this will be formally announced, as well as some documentation/guidance for engineers on how to make the change.

![image](https://user-images.githubusercontent.com/12416432/149238248-e8268d8b-3028-4c7e-85ec-3b58b8656680.png)

Might be good to place a banner of some kind referring to this announcement ahead of time so that devs see it coming.
Could place that warning on the homepage and then possibly also something here: https://cashmere.healthcatalyst.net/web/guides/supported-browsers.

aaronbrethorst:
🎉

corykon:
After talking w/ @andrew-frueh, leaning against the banner. Feels like overkill. We'll just put something on the browser support page.

Issue #1714, opened 2021-11-09
Target Angular v13 (May 4, 2022)


corykon:
Cashmere needs to target the version that is on LTS. Version 13 end active and enters LTS on May 4, 2022.
![image](https://user-images.githubusercontent.com/12416432/141010846-5bd2742b-dcdf-4b00-ac72-18fd76d1ff37.png)

- [ ] Rip out IE Support. Make mention of this in browser support guide. Look for other places that might be impacted throughout the docs by this change.
- [ ] Try converting from using "/" for division to using math.div [Sass Breaking Change](https://sass-lang.com/documentation/breaking-changes/slash-div)
- [ ] Update getting started guide that mentions a specific version of the angular CDK

corykon:
Considering pushing this to line up with the June 15th end-of-life date for IE on Windows 10.

PR #1797, opened 2022-01-25
docs(browser support): add note on plans for IE 11

corykon:
re #1786

Will revisit with the release of Cashmere v13. Made a note on that issue #1714. 

image


andrew-frueh (approved):
Looks great 👍

Issue #1798, opened 2022-01-27
CoreJS versioning issue on cashmere 12 upgrade


corykon:
![image](https://user-images.githubusercontent.com/12416432/151271554-0a522fed-8944-466e-b77c-8155c3fc70fc.png)

andrew-frueh:
@corykon - okay if I assign you to this one?

jacquie:
What about the sass '/' deprecation part? I can create a separate issue for it if that works better.

corykon:
> What about the sass '/' deprecation part? I can create a separate issue for it if that works better.

@jacquie, Yes! Haven't forgotten that. It's actually listed as an item over in #1714 
Forgot to note that here.

jacquie:
@corykon Ah! Sounds good.

PR #1806, opened 2022-02-07
build(deps): update guidance on core-js to allow for newer versions

corykon:
re #1798


jacquie:
I think that will work, we have v3.20.3 in PopInsights

jacquie (approved):
Looks good to me!

PR #1814, opened 2022-02-16
fix(polyfills.ts): paths to core-js items

heathkat:
fix paths to items in the core-js npm package

These paths in polyfills.ts seem to have changed from core-js/es6/reflect and core-js/es7/array to core-js/es/reflect and core-js/es/array.


corykon:
@heathkat Were you getting some build failures that led to making these changes?

Am I reading right that the package-lock.json is deleted? That's something we definitely want to keep around - it helps the build system run more smoothly.

heathkat:
@corykon, yes. I was getting build failures because the paths in polyfills.ts did not match the paths in the npm package it was targeting. Maybe I have a different version of that package? I don't mind if you don't want to merge those. I can just keep them updated in my local version. And yes, there were conflicts in my version of package-lock.json. I thought I was just deleting my local version? It gets generated whenever I run npm install. But yeah, if any of this causes problems for others, feel free to ignore.

corykon:
I'm going to close this one for now - we do want to make sure package-lock file sticks around, and it would be best not to update the polyfills file because it might break others who are still on an older version of core-js. Logged a new bug here: https://github.com/HealthCatalyst/Fabric.Cashmere/issues/1822

PR #1827, opened 2022-03-01
docs(node): specify required node version

aaronbrethorst:
Mention in the Contribution Guide that the project does not build reliably under the current Node LTS version (node 16 + npm 8), and that the previous LTS version, 14, should be use instead.

Also include link to `nvm`, which makes it easy to maintain multiple Node installations on the same machine, and add an `.nvmrc` file to make it easier to use the correct Node version when `nvm` is employed.

### DIFF
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,0 +1,1 @@
+lts/fermium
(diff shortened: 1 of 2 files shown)


andrew-frueh (approved):
Bravo again on coming up with this fix @aaronbrethorst

PR #1831, opened 2022-03-09
feat(instance-switcher): create instance switcher component

bskeen:
I added an instance switcher component, along with a supporting measurable component, to Cashmere.


corykon:
This looks awesome @bskeen 

Some thoughts-
- Lots of good logic in the example code. Thoughts on moving previous selected, onSelected() onClose()" logic into the component as overridable default behavior? Seems like a lot of implementation would be repeating this same code. At the very least I would maybe add some comments in the example code explaining what is going on.
- How could we better handle when there are too many instances to fit into the width allotted? Horiz scroll? "More" menu
- Style suggestions.
    - Would be nice to have them feel a little more "tabby" when selected
    - The add button could call for a little less attention.
    - If you're open to these changes I'd be happy to jump into some css
 
image

corykon:
@cikeddy I shared some styling suggestions w/ ben, would love to hear your thoughts on if any of those feel worth looking into.

Also ran into a couple things with the more menu. In this screenshot it just seems like "more" is a blue button - it may not read as selected if you didn't know better. Also would be cool to have the name of the selected instance visible somehow. Maybe the selected instance jumps out of the menu or something? 
image

bskeen:
I updated the logic to move the selected item out of the more menu:

image

I talked a little with @cikeddy about tying the chips a little more closely to the content (in your words above, making them more "tabby"). I'm not sure we had a very clear approach for doing that.

I also am looking now at moving more of the example logic into the component itself, I will update you on how that goes.

aaronbrethorst:
Apologies in advance for the drive-by feedback, and take it or leave it as you see fit, but maybe we could lift some design cues from Chrome's tab bar? I feel like it resolves some/many of the challenges that Cory described earlier.

image

corykon:
@bskeen the logic moving the selected instance out of the more menu looks perfect 👌 

@cikeddy @aaronbrethorst The chrome tabs are great. I'd be down to brainstorm on the design treatment a little bit if think it could be worth it, even for a v2.

bskeen:
@corykon I have updated the component so that some default logic for handling tab selection/removal is there. Do we want to move forward with merging this, and then maybe make the chips look more like tabs in a later iteration of this component? Or would you like to meet and discuss how that could look?

corykon (approved):

PR #1835, opened 2022-03-17
fix(date-range): allow blank dates when to/from dates aren't required

### LINKED ISSUE #1833: Date Range Component: Should be able to completely clear out date
When start and end date are optional, it should be possible to clear both dates out and hit apply.

From Guillermo Gomez on Slack

> I got excited about the news of the new cashmere release so I was trying to test it in the docs...
> playing with it in Stackblitz it seems that even when I have both the start and end dates required option marked as false, when I clear the default dates, I get an error when trying to apply my changes (click on the Apply button).

image

@aaronbrethorst any interest in hopping into that code once more?

aaronbrethorst:
A date range picker where 'to date required' and 'from date required' are set to `false` should allow a completely empty range with two null values for the date to be created. This change amends the range validation function to support the aforementioned scenario and adds a test to verify it.

Closes #1833

### DIFF
--- a/projects/cashmere-examples/src/lib/date-range/date-range-example.component.html
+++ b/projects/cashmere-examples/src/lib/date-range/date-range-example.component.html
@@ -22,7 +22,7 @@
                 {{ range.toDate | date: 'MM/dd/yyyy' }} or earlier
             </span>
             <span *ngIf="!range.fromDate && !range.toDate">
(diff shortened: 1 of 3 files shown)


andrew-frueh (approved):
This looks great @aaronbrethorst - and bonus points for adding the additional unit test!!

PR #1838, opened 2022-03-23
feat(slider): adds hc-slider component

andrew-frueh:
Borrowed a ton from the precedent you set with Multiselect @corykon, but took a slightly different approach here in providing a component wrapper for `ngx-slider` rather than just a theme.  That allowed me to make it compatible with `hc-form-field` as well.

Only concern I have is that as I read more on the GitHub for `ngx-slider` is that as of February the original author is stopping working on it.  But hopefully he'll turn the library over to someone else, or worst case we can fork and make updates as needed.

But it's a great component, lots of little things the author thought of that I hadn't considered - like what to do with the labels when the two sliders are close.

Screen Shot 2022-03-22 at 7 37 23 PM


aaronbrethorst (approved):
Code looks good, and the example works perfectly. I tested in Chrome and on iOS, and the slider works as well with touch events as it does with the mouse cursor. 👍

corykon (approved):
Really cool! Getting it to work in with the cashmere form field is awesome .

PR #1849, opened 2022-04-04
fix(searchbar): no tooltip if clear button is not shown

### LINKED ISSUE #1844: Search bar awkward tooltip
You can get the "clear" title text on the search bar even if that button isn't available. Probably need to set display to none or turn off cursor event when it is hidden.
image

corykon:
fix #1844


aaronbrethorst (approved):
nice work, good fix 👍

andrew-frueh (approved):
Looks good!

Issue #1847, opened 2022-03-31
Refresh login page


corykon:
How would we feel about restyling this login page a bit?
![image](https://user-images.githubusercontent.com/12416432/161116930-0f651d50-f7d0-4b95-939b-108b7efb3516.png)

Here's a quick idea. This refresh wouldn't be exciting persay, but the white card for the login UI and the input fields are more cashmere-y and the blue background would match the launch and logout screens better. 
![image](https://user-images.githubusercontent.com/12416432/161119518-f0d5f9fb-3848-4f05-9f49-058e40565209.png)

Or one of these backgrounds from healthcatalyst.com would be cool and look more modern.
![image](https://user-images.githubusercontent.com/12416432/161118133-d41d1cde-044d-431e-9a6f-06f00b45af80.png)
![image](https://user-images.githubusercontent.com/12416432/161118140-dc43930c-af33-47c9-93b3-f869569e1c48.png)
![image](https://user-images.githubusercontent.com/12416432/161118145-2bdc6a44-fbfc-4102-8f73-2777cfc6c684.png)

corykon:
@andrew-frueh @aaronbrethorst Feedback on this version?

![image](https://user-images.githubusercontent.com/12416432/161154760-6077b7f6-87cf-4269-a502-91a6a7e3e704.png)

aaronbrethorst:
I think it looks great!

andrew-frueh:
A few thoughts @corykon:

- Let's use the two-color version of the logo.  That's the most official version of our brand and for something like a login screen I think that's pretty important.
- Personally I think you could make the logo just a little bit bigger.  Doesn't have to be the same width as the inputs, but maybe half way closer to that.
- The form layout and buttons and everything looks great
- Out of curiosity, did you build responsive breakpoints into this?  Would be interested to see how this collapses at breakpoint `sm`
- The blue background is pretty intense, but looking at hc.com, it does seem like we're going with more vibrant imagery - so that probably works.  It also feels way more modern than the old login screen.  So I'd say go with it.

PR #1850, opened 2022-04-06
feat(login styles): refresh login screen

corykon:
re #1847

image


andrew-frueh (approved):
Bravo @corykon - did you ever end up creating/acquiring a vector version of that background?

corykon:
@andrew-frueh still a jpg, but it's only 70k which isn't bad. 20k smaller than the previous background

PR #1855, opened 2022-04-12
Navbar CSS Fix

### LINKED ISSUE #1843: Navbar username cuts off descenders
Changing the line height in `.hc-navbar-username` to 1.3 seems to fix this. Adds some extra space for two-line usernames, but I think this is probably ok.

e37c51c4-9fa6-413d-a168-e98fc82f4d29

JakeSaintG:
Increased the line-height in navbar. 

closes #1843

### DIFF
--- a/projects/cashmere/src/lib/sass/navbar.scss
+++ b/projects/cashmere/src/lib/sass/navbar.scss
@@ -226,7 +226,7 @@ $navbar-fixed-shadow: 0px 2px 6px $shadow;
         padding-top: 2px;
     }
     display: flex;
-    line-height: 1.3;
+    line-height: 1.2;
     align-items: center;
     white-space: nowrap;
 }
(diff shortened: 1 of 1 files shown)


corykon (approved):

TonyCurtisLives (approved):

projects/cashmere/src/lib/sass/navbar.scss
```diff
@@ -226,6 +226,7 @@ $navbar-fixed-shadow: 0px 2px 6px $shadow;
         padding-top: 2px;
     }
     display: flex;
+    line-height: 1.3;
```
andrew-frueh, line 229: Thanks Jake! Could we do `1.2` instead of `1.3` on this?  I was just testing it and 1.3 looked a little wide to me.  @corykon - agree/disagree?

andrew-frueh (changes requested):

projects/cashmere/src/lib/sass/navbar.scss
JakeSaintG: Sounds good! I'll make that change.

andrew-frueh: Just double checked and 1.2 still allows for descenders like 'y' without it running into the second line

PR #1857, opened 2022-04-20
Fix styling issue with the Instance Switcher

bskeen:


andrew-frueh (approved):
Looks good @bskeen - do you want me to get a patch release pushed with this change ASAP so you can pull it into your app?

bskeen:
That would be great, thanks.

PR #1871, opened 2022-05-19
fix(tabs): deselect all tabs

### LINKED ISSUE #1870: Deselect all tabs
**Summary**

Adding this request for the Vitalware team.  There currently is not a way to deselect tabs after the first is selected.  A `none` string can be passed to `defaultTab` parameter to have no tab initially selected, but after that there isn't a way to deselect.

Proposal is to add the ability to pass a -1 value to `selectTab` which will deselect any currently selected tab.  This will need to work with both the router and tabContent model of the tabs.

andrew-frueh:
Adds the ability to pass -1 to selectTab to deselect all.  

A little tricky when using tabs with routerLinks - because when deselecting I don't think we want to change the current route.  So in that instance I'm really just hiding the routerOutlet until a tab is selected again.

I also realized while I was in there that when using tabs with routerLinks, the selectTab function wasn't actually routing when called :-(  So that's been corrected here.

The first tab example has been updated to demonstrate deselecting.

closes #1870

### DIFF
--- a/projects/cashmere-examples/src/lib/tabs-horizontal/tabs-horizontal-example.component.html
+++ b/projects/cashmere-examples/src/lib/tabs-horizontal/tabs-horizontal-example.component.html
@@ -1,7 +1,7 @@
 <div class="tab-demo">
     <hc-tab-set direction="horizontal" (selectedTabChange)="selectionChanged($event)" #tabSet>
         <hc-tab tabTitle='Pending Tasks'>
-            <div class="tab-content">
+            <div class="example-tab-content">
                 This is where pending tasks would go.
             </div>
         </hc-tab>
@@ -9,7 +9,7 @@
             <hc-tab-title>
                 <i class="fa fa-check" style="color: #00a859; font-size: 20px;"></i>  Complete Tasks
             </hc-tab-title>
-            <div class="tab-content">
+            <div class="example-tab-content">
                 These are completed tasks
             </div>
         </hc-tab>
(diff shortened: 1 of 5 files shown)


aaronbrethorst:
@andrew-frueh I'm seeing an error when I try building this locally. LMK if I might be missing something obvious with this.

```
➜  cashmere git:(deselect-tabs) ✗ npm run build

# [500 lines of build noise elided...]

✖ Compiling with Angular sources in Ivy partial compilation mode.
projects/cashmere/src/lib/tabs/tab-set/tab-set.component.html:7:24 - error TS2341: Property '_routerDeselected' is private and only accessible within class 'TabSetComponent'.

7         <div [hidden]="_routerDeselected">
                         ~~~~~~~~~~~~~~~~~

  projects/cashmere/src/lib/tabs/tab-set/tab-set.component.ts:38:18
    38     templateUrl: './tab-set.component.html',
                        ~~~~~~~~~~~~~~~~~~~~~~~~~~
    Error occurs in the template of component TabSetComponent.
projects/cashmere/src/lib/tabs/tab-set/tab-set.component.html:8:35 - error TS2341: Property '_routerEnabled' is private and only accessible within class 'TabSetComponent'.

8             <router-outlet *ngIf="_routerEnabled" ></router-outlet>
                                    ~~~~~~~~~~~~~~

  projects/cashmere/src/lib/tabs/tab-set/tab-set.component.ts:38:18
    38     templateUrl: './tab-set.component.html',
                        ~~~~~~~~~~~~~~~~~~~~~~~~~~
    Error occurs in the template of component TabSetComponent.
projects/cashmere/src/lib/tabs/tab-set/tab-set.component.html:10:31 - error TS2341: Property '_routerEnabled' is private and only accessible within class 'TabSetComponent'.

10         <ng-container *ngIf="!_routerEnabled && tabContent" [ngTemplateOutlet]="tabContent"></ng-container>
                                 ~~~~~~~~~~~~~~

  projects/cashmere/src/lib/tabs/tab-set/tab-set.component.ts:38:18
    38     templateUrl: './tab-set.component.html',
                        ~~~~~~~~~~~~~~~~~~~~~~~~~~
    Error occurs in the template of component TabSetComponent.
```

andrew-frueh:
Yep, that was definitely a mistake.  Got that fixed @aaronbrethorst so it should build now.

aaronbrethorst (approved):
Looks good! 👍

corykon (approved):

PR #1873, opened 2022-05-20
fix(examples): fix rxjs-triggered build failures

aaronbrethorst:
We've been seeing obscure rxjs-related errors when trying to build and run cashmere-examples. It
appears that these are caused by the presence of two different versions of rxjs within our project.
This prospective fix seeks to resolve this issue by informing cashmere-examples that it should
always use the version 6 dependency. Note: this change should be removed as part of the Angular 13
upgrade, as that will bring us up to rxjs 7 across the board.

### DIFF
--- a/projects/cashmere-examples/package.json
+++ b/projects/cashmere-examples/package.json
@@ -11,6 +11,7 @@
         "@angular/core": "~12.2.13",
         "@ng-select/ng-select": "^7.4.0",
         "@angular-slider/ngx-slider": "^2.0.4",
+        "rxjs": "~6.6.0",
         "sugar": "^2.0.6",
         "change-case": "^4.1.1",
         "lodash": "^4.17.21",


corykon:
@aaronbrethorst I tried this approach, and was able to resolve some of the issues, but new ones cropped up. 
Basically anytime I run `npm i` from the root and let the postinstall script run on the package.json in projects/cashmere-examples, things would break. But if I ran `npm i` manually in the projects/cashmere-examples folder, things would resolve.

On my current branch I found that if I remove the peerDependencies entirely, than everything works great, no matter where I run `npm i` from. I think this ok because cashmere-examples is just used by the user-guide project, its not available anywhere externally so it doesn't need to worry about peer dependencies.

image

Hopefully that all makes sense - grab me and we can look at it together, but I think I'd recommend abandoning this PR.

aaronbrethorst:
@corykon doh, thanks for poking at it! hopefully this'll all go away soon anyway with the Angular 13 upgrade :)

PR #1887, opened 2022-06-03
Slide toggle component

### LINKED ISSUE #1861: Switch Component
**Describe your idea**

Create a boolean switch component for instances where a checkbox or button toggle aren't appropriate.  @stephfeller will design the component, Porter Lyman will do the initial development, and @andrew-frueh will pull it into Cashmere.

ab06da0d-0b60-4ac9-8d32-cd7254b42256

andrew-frueh:
Thanks again for all the foundational work you did on this one @pslyman!

I think I've got this implementation of the slide toggle ready to go, but let me know if you all have any suggestions.  Supports `formControl` and `hc-form-field`, has decent unit test coverage, and has the stylesheet broken out into mixins.

closes #1861


projects/cashmere/src/lib/slide-toggle/slide-toggle.component.ts
```diff
@@ -0,0 +1,165 @@
...
+    public onChange: (value: unknown) => void = () => undefined;
+    public onTouch: () => unknown = () => undefined;
+    public registerOnChange(fn: (value: unknown) => void): void {
+        this.onChange = fn;
+    }
+    public registerOnTouched(fn: () => unknown): void {
+        this.onTouch = fn;
+    }
+
+    ngDoCheck(): void {
```
andrew-frueh, line 144: I mentioned to @corykon that this implementation of `ngDoCheck` gives me heartburn, because a half-dozen of our form components do it.  It was done this way several years ago because there wasn't a good way to subscribe to form changes, but there _has_ to be a better way to do this.  On big forms you could have tons of components all calling doCheck several times a second 😬.  I've implemented it here as with other components in Cashmere, but I'll put an issue in for it.

corykon (approved):

pslyman:
Looks great! Just wanted to verify that "TRUE"/"FALSE" fit in the limited space without cutting off?

pslyman (approved):

andrew-frueh:
@pslyman - that's a great question, because I realized as soon as I started jumping into your code that getting the width right on these slide toggles is a non-trivial problem, especially when the two sides aren't the same number of letters.  If you take a look at the stylesheet, you'll see that I had to introduce some manual size adjustments for true/false to make sure there was enough space.  It's also the reason I didn't allow the dev to put any text they wanted in there, b/c it's hard to make it fully dynamic.

PR #1899, opened 2022-06-10
Getting started update

corykon:
Some changes to the documentation, examples, and build scripts around how we use sass. There have been changes around the way SASS imports within angular.
- tilde imports are deprecated and don't work in most cases, so had to switch to relative paths or absolute paths in some places
- in examples, the node_modules paths don't compile, but `@healthcatalyst/../..` paths do

I'm not in a hurry to get this PR in, but wanted to show what I'm finding that works. I wish I could double check this in stackblitz, but Cashmere 13 still isn't coming up. 

We should probably regroup and look together at all the issues we've seen with Angular & Cashmere 13, talk about how to avoid those in the future for major releases. Maybe some more diligent testing in a few projects before announcing the update.

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,7 +125,6 @@
     "lint-staged": "^7.1.2",
     "markdown-it-named-headers": "0.0.4",
     "ng-packagr": "^13.3.1",
-    "node-sass-tilde-importer": "^1.0.2",
     "path": "^0.12.7",
     "prettier": "2.1.2",
     "progress": "^2.0.3",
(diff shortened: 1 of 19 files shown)


package.json
```diff
@@ -125,7 +125,6 @@
     "lint-staged": "^7.1.2",
     "markdown-it-named-headers": "0.0.4",
     "ng-packagr": "^13.3.1",
-    "node-sass-tilde-importer": "^1.0.2",
```
andrew-frueh, line 128: Hah - I wondered about this package!  Looks like it's not necessary anymore

andrew-frueh (approved):

PR #1904, opened 2022-06-21
feat(drawer): animated param

### LINKED ISSUE #1901: Ability to turn off animation on hc-drawer
We are using the hc-drawer and would like the ability to pass a variable that allows the drawer to turn off the slide animation. 

When we leave certain tabs and then come back, the page redraws and the drawer slides out again. This makes it appear that the data is always updating. 

Thanks

andrew-frueh:
adds ability to turn drawer animation on or off

@robertjorg - let me know how this looks.  The only gotcha is that the animation disable will actually be on `hc-drawer-container` rather than `hc-drawer`.  The reason being there is animation on both the drawer and the content.  So the toggle needs to be on the container so it can enable/disable both.

closes #1901


andrew-frueh:
Looks like I've got some unit tests to fix 😔

corykon (approved):
The code and new example look good. Ready to merge when those tests are fixed up :+1:

robertjorg (approved):
These changes look good.

PR #1907, opened 2022-06-24
1905 package exports

corykon:
Update to the library exports to make sure we can import cashmere styles without absolute paths. Updated the docs to match.


andrew-frueh (approved):
Looking good!  Pulled the changes and `build` and `start` both run just fine for me.  Import syntax looks a lot better this way.

PR #1908, opened 2022-06-24
feat(table): hc-index-cell

### LINKED ISSUE #1816: Row Numbers on Data Table
**Describe your idea**

On the data table, we have a use case where we'd like to have the row numbers available. This is something we could handle in our app, but it seems like something that is unrelated to the data itself and could be used by many others directly from Cashmere.

From @corykon in a PBI where this was discussed (~credit to him for the idea in the first place~ Credit to Mike Buck actually 😄):

>One thing I just realized that seems like such an easy win item now that I consider it was to add row numbers to the Explore tab. Just like in Excel, when we sort things in different orders I can check the counts of certain things like patients less than 75 years old without having to export things to Excel. It wouldn't be part of the SQL probably something in the Client side code I bet.

**Additional context**

image

andrew-frueh:
adds the index-cell type for row number columns

@jonest @corykon let me know what you think of this approach.  My original approach was to add a `rowNumbers` param to hc-table, and then have it automatically add a column with numbers.  But the more I thought about it, the goal with hc-table is to be as flexible as possible, and it seemed like that kind of restricted the value of this to a pretty narrow use case.

So instead I added a new cell type called `hc-index-cell` which can be added to a header or body cell.  When applied it narrows the width and padding of the cell and applies the darker background.  It still respects the padding on sort headers though, and works fine with tight tables.

I updated the first two examples to demonstrate the new cell type.

closes #1816

### DIFF
--- a/projects/cashmere/src/lib/table/cell.ts
+++ b/projects/cashmere/src/lib/table/cell.ts
@@ -7,7 +7,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {Directive, ElementRef, Input} from '@angular/core';
+import {Directive, ElementRef, HostBinding, Input} from '@angular/core';
 import {CdkCell, CdkCellDef, CdkColumnDef, CdkFooterCell, CdkFooterCellDef, CdkHeaderCell, CdkHeaderCellDef} from '@angular/cdk/table';
 
 /**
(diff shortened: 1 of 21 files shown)


projects/cashmere/src/lib/table/cell.ts
```diff
@@ -135,3 +135,18 @@ export class HcCell extends CdkCell {
         elementRef.nativeElement.classList.add(`hc-table-justify-` + columnDef.justify);
     }
 }
+
+/** Row index cell template container that adds the right classes and role. */
+@Directive({
+    selector: 'hc-index-cell, td[hc-index-cell], th[hc-index-cell]',
+    host: {
```
jonest, line 142: I prefer consistency, so if this is what we're doing elsewhere, just disregard this, but the Angular style guide suggests to use the `@HostBinding` decorators instead of the metadata: https://angular.io/guide/styleguide#hostlistenerhostbinding-decorators-versus-host-metadata=

jonest: Based on the reasonings provided, it's probably not that important since it doesn't really seem to apply here

andrew-frueh: Hey @jonest - that's a great catch, and I'm pretty sure that code comes from back when this was all abstracted from Angular Material's table.  It looks like they haven't updated their syntax yet in their repo, but I agree - `@hostbinding` is what we use everywhere else in Cashmere.  I'll make the adjustment as a part of this PR.

jonest: Only noticed because I didn't know what that stuff meant so I looked up the docs on it, haha 😆

PR #1910, opened 2022-06-28
fix(measureable): example load fix

### LINKED ISSUE #1909: Measureable Example Load Bug
**Summary**

When you go to the Measureable component, it works as expected if you first load the Cashmere site from another page and then jump to Measureable.  But if you load the page directly from the url:

https://cashmere.healthcatalyst.net/web/components/measurable/examples

...the results are inconsistent.  Most times the "More" button triggers even though it shouldn't (should just see the two buttons initially).  Suspect it may have something to do with load times.

andrew-frueh:
@bskeen - just a small fix here that should get rid of the inconsistent behavior on initial load.  It's annoying to have to add an empty `setTimeout` but that seems to be the silver bullet to make sure the widths get calculated only after everything is finished rendering.

I also made an adjustment to the style of the buttons in the button row so they look different from the action buttons.  Just makes the functionality of the example a little clearer.

closes #1909

### DIFF
--- a/projects/cashmere-examples/src/lib/measurable-overview/measurable-overview-example.component.html
+++ b/projects/cashmere-examples/src/lib/measurable-overview/measurable-overview-example.component.html
@@ -3,7 +3,7 @@
         <hc-measurable
             *ngFor="let button of buttons"
             [itemKey]="button.key">
-            <button hc-button>
+            <button hc-button buttonStyle="secondary">
                 {{button.text}}
                 <hc-icon
                     hcIconSm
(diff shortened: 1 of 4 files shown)


andrew-frueh:
FYI @bskeen I think the issue on load is primarily just a Firefox thing.  Seems like it's not an issue on Chrome

bskeen:
@andrew-frueh I have seen some inconsistencies on load in the instance switcher, I'm wondering if it's a similar problem?

bskeen (approved):
Looks good to me.

PR #1917, opened 2022-07-13
docs(stackblitz): temporary stackblitz version fix

andrew-frueh:
Frustrating to have to do this - but I hate having broken Stackblitz links while we wait for a response from them.  What do you both think?


aaronbrethorst:
I think its better than leaving them broken!

corykon (approved):
Probably good to do this now. I realllly hope this doesn't last long though.

andrew-frueh:
Agree with both of you.  I'll see if there is any way to escalate an issue with Stackblitz.  Tricky if we aren't paying them though.  I'll get this merged for now.

PR #1918, opened 2022-07-16
Tabs more menu

### LINKED ISSUE #1880: Responsive Tabs and Tabset
**Describe your idea**

Update the tabset/tabs to be more responsive. It would be similar to the instance switcher and measurable components. When there is not enough space to display the tabs it would have a more menu dropdown showing the remaining tabs.

If there are too many tabs it can overflow the container and it looks a bit strange.

**Additional context**

andrew-frueh:
Huge thanks again @HaydenDalley35 for the suggestion and code examples on this one.  It's a massive improvement to the Tabs component - one of the very first Cashmere components.

This is a pretty big PR, and admittedly I snuck in a few fixes to things that aren't related to tabs.  So to make your lives easier, I'll schedule a 30 min call for early next week to walk you through what I did here.  And hopefully that'll make reviewing easier.  Thanks!

closes #1880

### DIFF
--- a/projects/cashmere-examples/src/lib/measurable-overview/measurable-overview-example.component.html
+++ b/projects/cashmere-examples/src/lib/measurable-overview/measurable-overview-example.component.html
@@ -36,9 +36,9 @@
         <button
             *ngFor="let button of moreButtons; index as i; trackBy: buttonTrackBy"
             hcMenuItem>
-            <span hcMenuText class="demo-more-text">{{button.text}}</span>
+            <span hcMenuText>{{button.text}}</span>
             <hc-icon
-                class="demo-remove-button"
+                hcMenuSubText
                 (click)="buttonClose(button.key, $event)"
                 fontSet="fa"
                 fontIcon="fa-times">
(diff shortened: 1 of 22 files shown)


bskeen:
@andrew-frueh A couple thoughts on the new arrow implementation:

- If there are a lot of tabs (as there are in some Vitalware products that would be using this), the scrolling might be too slow. I'm not sure if it would be possible to scroll more if it is just a click as opposed to holding it down? Almost paging through the tabs, but I wouldn't say it would have to be a screen width, just maybe a little more than what it is doing now, or speeding up the scrolling.
- Another possible suggestion would be to add scrolling support if it is in 'arrow' mode. In other words, if you are hovering over the tabs and scroll, it will go left or right for you. Not sure if that would fall in the scope for this PR, but maybe it could be a later enhancement?

Aside from my above comments, I think it works pretty well overall. I couldn't see any outright problems with it.

andrew-frueh:
Thanks for the quick review @bskeen - I'd be happy to increase scroll distance on each interval if others agree the scroll feels too slow.  Like you said, I don't think I'd want to up to a whole page because it would feel too jarring, but I could certainly double or triple the current amount.

bskeen:
@andrew-frueh ok great thanks! One more thought, would it be possible to gray out the arrows when we reach the end? It might be a good way to indicate that to the user.

PR #1921, opened 2022-07-19
feat(instance-switcher): make instance switcher editable

bskeen:
update the instance switcher so each instance is editable on a double click or context menu

### DIFF
--- a/projects/cashmere-examples/src/lib/instance-switcher-overview/instance-switcher-overview-example.component.html
+++ b/projects/cashmere-examples/src/lib/instance-switcher-overview/instance-switcher-overview-example.component.html
@@ -11,8 +11,8 @@
         (closed)="toggleOpen()">
     </hc-instance-switcher>
 
-    <button hc-button (click)="toggleOpen()" class="demo-button">Toggle Open</button>
-    <button hc-button (click)="toggleEditable()" class="demo-button">Toggle Editable</button>
-    <button hc-button (click)="toggleClosable()" class="demo-button">Toggle Closable</button>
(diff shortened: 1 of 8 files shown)


andrew-frueh (changes requested):
Nice enhancement @bskeen!  Two suggestions:

1. For the instance switcher example, can we either switch all those buttons to say: "Editable: On", etc.  Or switch the buttons to checkboxes.  Especially with this upgrade it wasn't clear to me to the current state of the example - whether it was currently editable or not.
2. I'd recommend that when the chip goes into edit mode, you put a white box around the input - so it basically looks like a Cashmere input inside the chip.  I think that will make it crystal clear that you can type in the chip.

bskeen:
Thanks for the suggestions, @andrew-frueh ! I just pushed the requested changes, if you would be able to take a look that would be great.

andrew-frueh (approved):
Looking good @bskeen - the only additional suggestion I'd offer (and I'd say this is optional), would be to close and disable a tooltip on the chip while it's in edit mode.  As you can see in the attached screenshot, it's a bit obnoxious to have the tooltip hovering over your input area :-)

Screen Shot 2022-08-03 at 2 49 42 PM

bskeen:
@andrew-frueh I was able to remove tooltips from instances being edited. I worked with Cory to find a way to avoid having the ternary statement in the template, but for some strange reason, it couldn't find the method in the component. Either way, this should function as desired. Please review when you have a minute and let me know what you think.

andrew-frueh (approved):
Ah nicely done @bskeen @corykon.  Hiding the tooltips on edit looks great to me.  I think this one is good to go!  I'll merge this PR.

PR #1942, opened 2022-08-29
fix(highlight-pipe): html tag fix

### LINKED ISSUE #1941: Highlight pipe breaks on html tags
**Summary**

When a block of html text is fed into the highlight pipe, any matches on the html tags or their containing properties will attempt to be highlighted and break the markup.

Need to be able to parse the html and only highlight content, not tags.

andrew-frueh:
Prevents the highlight pipe from trying to highlight html tags.

@aaronbrethorst @corykon wow this one really stretched my brain.  But Aaron your original take was on point, the only really safe way to fix this was to traverse the children of an html object.  All the potential regex solutions had holes in them.  The tricky part was that for the pipe we don't have a DOM object, just a string.  Fortunately today I learned about `DOMParser()` which is apparently widely supported and does exactly what we need here.  Seems really fast and pretty bullet-proof. I think my recursive function `searchChildren` should get to everything in a hierarchy but please check me on that.

To test the functionality, go to the Pipes section, Highlight Pipe, and try searching for "<", you should get nothing.  Compare that to the current demo site.

closes #1941


projects/cashmere/src/lib/pipes/highlight/highlight.pipe.ts
```diff
@@ -17,14 +17,30 @@ export class HighlightPipe implements PipeTransform {
...
             return preserveHTML ? text : this.escapeTags(text);
         }
     }
 
+    searchChildren( parent: ChildNode, regex: RegExp ): void {
+        if( parent.childNodes.length > 0 ){
+            parent.childNodes.forEach( child => this.searchChildren( child, regex ) );
+        } else {
+            if( parent.nodeValue ) {
+                parent.nodeValue = parent.nodeValue.replace(regex, match => `<span class="hc-text-highlight">${match}</span>`);
```
aaronbrethorst, line 39: clever.

aaronbrethorst (approved):
looks good to me! 👍

PR #1943, opened 2022-08-30
fix(radio): support child inputs

### LINKED ISSUE #1933: undefined ChangeEvent is emitted from HcRadioGroup when a radio button label has a numeric input
**Summary**

If you put an `<input type="number">` inside the label of an `hc-radio-button`, the parent `hc-radio-group` will emit `RadioButtonChangeEvent`s with value `undefined` when the increment/decrement buttons in the input are pressed.

**Reproduction**

In the following Stackblitz, click the up/down arrows in either numeric input. The first input is an hcInput wrapped in an `hc-form-field`, the second is a simple HTML input.

Stackblitz reproduction:

https://stackblitz.com/edit/angular-j7cpqf?file=src/app/radio-button-standard/radio-button-standard-example.component.html

andrew-frueh:
allows radio buttons to include input elements

closes #1933

### DIFF
--- a/projects/cashmere-examples/src/lib/radio-button-standard/radio-button-standard-example.component.html
+++ b/projects/cashmere-examples/src/lib/radio-button-standard/radio-button-standard-example.component.html
@@ -1,7 +1,17 @@
 <form>
-    <hc-radio-group>
+    <hc-radio-group (change)="updateSelectedOption($event)">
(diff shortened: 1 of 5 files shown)


isaaclyman (approved):

corykon (approved):
makes sense to me. nice work 👍

Issue #1946, opened 2022-09-01
TabSet component enhancement to have an option with no overflow style


guillegu:
Since Cashmere TabSet component was updated to include the tab overflow style, it'd be great to add an option to have no overflow style.

We need in our app to have a three static tabs and there should not be any overflow.

Right now the default behavior is that it looks like this

![image](https://user-images.githubusercontent.com/112650137/187975785-6a05a083-2b63-4263-af9b-e7791a0b068e.png)

whereas with that option it should look like this

![image](https://user-images.githubusercontent.com/112650137/187976062-a15a16bf-f1e7-4529-8857-5792d900f058.png)

PR #1947, opened 2022-09-08
fix(tabs): allow for no overflow controls

corykon:
re #1946

image


andrew-frueh (approved):
Well done, Cory, thank you!!

Issue #1945, opened 2022-09-01
Display selected overflow tab at end of visible list


jonest:
**Tab Overflow** - In OneNote, when I select a tab from the overflow dropdown, the selected tab becomes the last tab at the top

Before selection
![image](https://user-images.githubusercontent.com/3075414/187971909-125efc84-6a19-4a91-ad4d-993f862d272a.png)

After selecting, the tab is at the end and selected
![image](https://user-images.githubusercontent.com/3075414/187971957-b5269136-a2b4-48d6-904f-1a376583d4af.png)

Switching to another tab, the tab stays at the end
![image](https://user-images.githubusercontent.com/3075414/187972008-ae37e764-5926-46ff-b7bf-25db4f3281bf.png)

@bskeen mentioned some sort of functionality that may already exist? Maybe he can fill in some details here 😃

bskeen:
In the Instance Switcher, we currently have logic that will pop the currently selected instance outside of the more menu. We could probably use that same logic here.

PR #1948, opened 2022-09-08
fix(tabs): make sure selected tab is always shown

corykon:
re #1945

008a5483-5c4f-4a4c-b06c-f94deba751b2


andrew-frueh (approved):
Thanks again for knocking these out, Cory!

PR #1953, opened 2022-09-20
fix(tabs): base64 encode icons

### LINKED ISSUE #1952: Tab set uses Font Awesome in the more and arrows overflowStyle options
The tabset component uses Font Awesome for the chevrons in the 'arrows' and 'more' overflowStyle options, whereas other components like the Accordion use base64 encodings for their chevron icons.

We have eliminated Font Awesome in favor of the hc-icons set so we would greatly appreciate it if we could have base64 encodings in the tabset component too please

corykon:
fix #1952


jacquie (approved):
Looks good - thanks for getting this done so quickly @corykon

PR #1961, opened 2022-09-30
feat(instance switcher): add length/no entry

### LINKED ISSUE #1959: Small changes to Instance Switcher
**Describe your idea**

Ability to have a fixed with on the instance switcher tabs.
Allow when no name is entered to change an instance name that it will default to the name that was prior

Long tab names and fixing when no name is entered in a tab

robertjorg:
Update instance switcher to allow for maxWidth or fit content.
Also will default to the prior name if the new name entered is blank.
Closes #1959

### DIFF
--- a/projects/cashmere-examples/src/lib/instance-switcher-overview/instance-switcher-overview-example.component.html
+++ b/projects/cashmere-examples/src/lib/instance-switcher-overview/instance-switcher-overview-example.component.html
@@ -18,7 +18,7 @@
     <button hc-button (click)="toggleTooltips()" class="demo-button">{{getTooltipDescription()}}</button>
     <div class="property-form-row">
         <hc-form-field>
-            <hc-label>Max width (ex: 300px):</hc-label>
+            <hc-label>Max tab width (ex: 300px):</hc-label>
             <input hcInput [formControl]="maxChipWidthControl" />
         </hc-form-field>
     </div>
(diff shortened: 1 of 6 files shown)


corykon (commented):
Awesome to have you making contributions! Adding a width constraint to these will be super useful. Noticed a couple things that should be looked into before merging:
- For clarity, the `maxWidth` input could be named something like `chipMaxWidth`. At first I though this was about constraining the width of the whole bar.
- I noticed that I can end up in this state pretty easily where the overflow just stops working:
image

andrew-frueh:
Ah @corykon you make a good point about `max-width`.  I gave @robertjorg some bad direction there, because I was thinking about consistency with hc-tabs.  But the difference there is that you set the `max-width` param on each individual tab, not the whole tabset.  So it's a lot clearer what you are applying the max width to.  I think if we are setting a single max width value that applies to all the items in the switcher, your naming suggestion is on point.

Robert - I'm guessing that you always want the max width of the chips in the instance switcher to the same, correct?  I can't really think of a reason why you'd want a width constraint on one but not another (it's kind of a different situation for tabs).

robertjorg:
Yeah, that is correct. That is how it is working. I'm just trying to figure out Cory's second comment on the `More` button. That one is giving me more heartburn

robertjorg:
@corykon I renamed the variables and also talked with Ben and figured out where the more menu issue you were seeing was. The issue was happening prior, my change just made it more prominent. I believe the latest push should fix that issue for the Instance Switcher.

corykon (approved):
thanks @robertjorg

andrew-frueh (approved):

PR #1964, opened 2022-10-04
fix(input): mobile style init

### LINKED ISSUE #1949: Mobile Setting Broken on hcInput
**Summary**
The mobile setting on hcInput does not get applied consistently. It appears to only work if set after the initial load.

**Reproduction**

Steps to reproduce:

1. Set an hcInput to mobile="true".
2. Observe that the text size is not 16px.

andrew-frueh:
mobile styling was not applied when added initially

closes #1949


cikeddy (approved):
Checked it by pulling down the branch and setting mobile="true" on the search bar example. Looks like it's fixed!

PR #1969, opened 2022-10-10
feat(environment switcher): make env switcher generic

### LINKED ISSUE #1968: Override for Environment Switcher labels
We have a new feature to allow our users to choose a session time zone which adjusts the display of all times in the application. We really like the styling of the [Environment Switcher](https://cashmere.healthcatalyst.net/web/components/env-switcher/examples) for this feature, but unfortunately it's hardcoded for the Environments use case (collapsed tooltip and label for the expanded list).

Can we make this a more generic component or enable overriding those labels?

Examples
image
image

alexdow-hc:
remove hardcoded 'environment' labels and add override for dropdown header

close #1968


Techgeekster (approved):

corykon (changes requested):
@alexdow-hc this looks really good! I think my only ask would be to check out the spacing - it looks like git thinks every line was changed so it'll unnecessarily change a bunch of commit history.

corykon:
I'd also recommend using `feat` instead of `refactor` for your commit, since we're technically adding new functionality.

alexdow-hc:
@corykon Those issues have been resolved

corykon (approved):
Thanks @alexdow-hc. I'm noticing now that tests are failing, but we should be good to go after that.

PR #1974, opened 2022-10-12
fix(instanceswitcher): individual tool tip

### LINKED ISSUE #1973: Instance Switcher Tooltip
**Summary**

Instance Switcher tool tip uses the same tool tip for all tabs. We should be able to see different tool tips. For example, if the max tab width is set and the name goes beyond the end of the tab the tool tip should show the full name.

**Reproduction**

Steps to reproduce:

1. Here we see that each instance does not have a tool tip. The file only uses the IInstanceSwitcherTooltipText for all tabs.

export interface IInstance {
    /** The unique key corresponding to this instance. */
    instanceKey: string;

    /** The text that should be displayed on the instance switcher for this instance. */
    displayText: string;
}

export interface IInstanceSwitcherTooltipText {
    /** The tooltip text for the instances. */
    instanceText?: string;

    /** The tooltip text for the add button. */
    addText?: string;

    /** The tooltip text for the close button. */
    closeText?: string;
}

robertjorg:
Add individual tool tips for the instance switcher

closes #1973

### DIFF
--- a/projects/cashmere-examples/src/lib/instance-switcher-overview/instance-switcher-overview-example.component.ts
+++ b/projects/cashmere-examples/src/lib/instance-switcher-overview/instance-switcher-overview-example.component.ts
@@ -15,17 +15,18 @@ export class InstanceSwitcherOverviewExampleComponent {
     instances: IInstance[] = [
         {
             instanceKey: 'instance1',
-            displayText: 'Instance 1'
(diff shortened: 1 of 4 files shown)


andrew-frueh (approved):
Nicely done @robertjorg - great improvement

PR #1980, opened 2022-10-18
feat(tab-set): add selected tab input to tab set

bskeen:
I added an input that selects which tab is currently shown without emitting a tab change event.

feat 1972


jacquie (approved):
Looks good! Thanks for doing this, @bskeen 👍

PR #1985, opened 2022-11-01
fix(datepicker): allow blank values with new datepicker option

### LINKED ISSUE #1984: hcDatepickerParse error when clearing a datepicker value
**Summary**

Deleting the date in the datepicker input box causes the control to have the hcDatepickerParse error and is marked invalid.  There doesn't appear to be an obvious way to clear the date value normally.  My workaround was to listen for changes and set the value of the form control to undefined.  This seemed to "trick" the validator into considering it as valid:

```
/** The form control validator for whether the input parses. */
private _parseValidator: ValidatorFn = (): ValidationErrors | null => {
    return this._lastValueValid ? null : {hcDatepickerParse: {text: this._elementRef.nativeElement.value}};
};
```

**Reproduction**

Steps to reproduce:

1. Pick a date with the datepicker
2. Click in the input box attached to the datepicker and delete it with Backspace
3. Click outside the input box
4. The control has an error state.

For some reason I had a hard time getting a Stackblitz example working for a date picker in date mode, but [this example](https://stackblitz.com/edit/angular-32eyjc?file=src%2Fapp%2Fdatepicker%2Fdatepicker-example.component.html) with a time picker should hopefully show the issue.  

**Additional context**
Below is an animated GIF of the issue on the [Cashmere example page](https://cashmere.healthcatalyst.net/web/components/datepicker/examples) which is what I saw when developing locally.
C-DatePickerClear

aaronbrethorst:
Expose support for `allowBlankValues` in the DatePicker component.

fix #1984

### DIFF
--- a/projects/cashmere/src/lib/datepicker/datepicker.component.ts
+++ b/projects/cashmere/src/lib/datepicker/datepicker.component.ts
@@ -73,9 +73,9 @@ export class DatepickerComponent implements OnDestroy {
     get allowsBlankValues() {
         return this._datepickerInput && this._datepickerInput.allowsBlankValues;
     }
-    set allowsBlankValues(val) {
+    set allowsBlankValues(val: boolean) {
         if (this._datepickerInput) {
-            this._datepickerInput.allowsBlankValues = val;
+            this._datepickerInput.allowsBlankValues = parseBooleanAttribute(val);
         }
     }
 
--- a/projects/cashmere/src/lib/datepicker/datepicker.component.ts
+++ b/projects/cashmere/src/lib/datepicker/datepicker.component.ts
@@ -67,6 +67,19 @@ export class DatepickerComponent implements OnDestroy {
     private _scrollStrategy: () => ScrollStrategy;
 
     /**
+     * Whether the datepicker can be set to a blank value (by deleting its contents). Defaults to `false`.
+     */
+    @Input('allows-blank-values')
+    get allowsBlankValues() {
+        return this._datepickerInput && this._datepickerInput.allowsBlankValues;
+    }
+    set allowsBlankValues(val) {
+        if (this._datepickerInput) {
+            this._datepickerInput.allowsBlankValues = val;
+        }
+    }
+
+    /**
      * Whether the datepicker includes the calendar, time selector, or both. Defaults to `date`.
      */
     @Input()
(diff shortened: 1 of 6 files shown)


projects/cashmere/src/lib/datepicker/datepicker.component.ts
```diff
@@ -66,6 +66,19 @@ export class HcDatepickerContentBase {
...
     private _scrollStrategy: () => ScrollStrategy;
 
+    /**
+     * Whether the datepicker can be set to a blank value (by deleting its contents). Defaults to `false`.
+     */
+    @Input('allows-blank-values')
+    get allowsBlankValues() {
+        return this._datepickerInput && this._datepickerInput.allowsBlankValues;
+    }
+    set allowsBlankValues(val) {
```
cikeddy, line 76: Not sure if this matters, but should val be set to a type of boolean here? should you use parseBooleanAttribute(val) as well?

cikeddy (commented):
All looked good to me! Just had one small comment to confirm that something was left out intentionally.

projects/cashmere/src/lib/datepicker/datepicker.component.ts
aaronbrethorst: Good Q. @andrew-frueh ? Can you help answer?

andrew-frueh: Great catch on your review @cikeddy - yes, `parseBooleanAttribute` is the preferred way to handle a boolean input so that we can accept either string, number, or boolean values for that boolean param.  So if either of you could make that switch, then we should be good to merge this one.  Thanks to both of you!

projects/cashmere/src/lib/datepicker/datepicker.component.ts
```diff
@@ -73,9 +73,9 @@ export class DatepickerComponent implements OnDestroy {
     get allowsBlankValues() {
         return this._datepickerInput && this._datepickerInput.allowsBlankValues;
     }
-    set allowsBlankValues(val) {
+    set allowsBlankValues(val: boolean) {
         if (this._datepickerInput) {
-            this._datepickerInput.allowsBlankValues = val;
+            this._datepickerInput.allowsBlankValues = parseBooleanAttribute(val);
```
aaronbrethorst, line 78: thank you!

andrew-frueh (approved):

PR #2007, opened 2022-12-01
fix(tab-set): fix selectedTab input

bskeen:
I fixed a couple of issues with the selectedTab input causing it to break on component init


corykon (approved):
Thanks for this, Ben

PR #2014, opened 2022-12-13
fix(drawer): prevent drawer from getting 'stuck'

### LINKED ISSUE #2005: Drawer gets stuck open if a user calls `toggleOpen()` twice
**Summary**

Calling `toggleOpen()` while drawer is open causes drawer to stay stuck open.

**Reproduction**

When the drawer is open, call `toggleOpen()` again, then try to call `toggleClose()` and it won't close.

**Additional context**

It seems to be related to the animation promise not resolving when calling `toggleOpen()` the second time. I believe if we just add a check internally for whether the drawer is open, we can just return a promise right away with the expected result (or maybe just ignore it since the drawer wasn't actually opened?)

aaronbrethorst:
Immediately return a resolved Promise from the Drawer component's `toggle()` function if the desired state of the Drawer matches the current state of the Drawer. This will prevent the Drawer from getting 'stuck' open when an open Drawer is asked to open itself.

This allows callers to request a particular state without worrying about the Drawer's current state.

fix #2005

### DIFF
--- a/projects/cashmere/src/lib/drawer/drawer.component.ts
+++ b/projects/cashmere/src/lib/drawer/drawer.component.ts
@@ -259,6 +259,14 @@ export class Drawer implements AfterContentInit {
 
     /** Toggles the drawer */
     toggle(isOpen = !this.opened): Promise<DrawerPromiseResult> {
+        // If the current state of the drawer matches the desired state of the drawer,
+        // then return an already-resolved Promise whose result matches the current drawer
+        // state. Ideally, we'd punt and return null here, instead, but that would break the
+        // API contract defined by the Drawer, and that would be bad.
+        if (isOpen === this._drawerOpened) {
+            return Promise.resolve(new DrawerPromiseResult(isOpen ? 'open' : 'close'));
+        }
+
         if (!this._animationPromise) {
             this._drawerOpened = isOpen;
 


projects/cashmere/src/lib/drawer/drawer.component.ts
```diff
@@ -259,6 +259,14 @@ export class Drawer implements AfterContentInit {
 
     /** Toggles the drawer */
     toggle(isOpen = !this.opened): Promise<DrawerPromiseResult> {
+        // If the current state of the drawer matches the desired state of the drawer,
+        // then return an already-resolved Promise whose result matches the current drawer
+        // state. Ideally, we'd punt and return null here, instead, but that would break the
+        // API contract defined by the Drawer, and that would be bad.
+        if (isOpen == this._drawerOpened) {
```
jonest, line 266: I think we prefer the triple equals here 🙂

jonest (changes requested):

andrew-frueh (approved):
Agree with @jonest suggestion on the triple equal, but otherwise this looks like a good fix 👍

projects/cashmere/src/lib/drawer/drawer.component.ts
aaronbrethorst: good catch, done.

jonest (approved):

jonest:
Hope it's cool for me to complete it now 😄

PR #2043, opened 2023-02-10
fix(tabs): arrow overflow scrollIntoView

### LINKED ISSUE #2042: Cashmere Tabs is not scrolling into view on selectTab() function
**Summary**

Hello Cashmere team, first of all, thanks for your amazing work!!
I'm reaching you with a quick question regarding the Cashmere tabs.
I'm selecting the last tab with selectedTab function and it works but I need to click the arrows to find the tab, there is a way to auto-focus the selected tab, in this case at the end?

**Reproduction**

https://user-images.githubusercontent.com/125039293/218105467-fa287012-2eeb-4211-96df-32dee6c16d33.mp4

andrew-frueh:
Scrolls a tab into view when clicked or selected with selectTab

I kept the use case pretty narrow here because `scrollIntoView` can be pretty annoying if it's firing to often or in the wrong context.  So this is only triggered on horizontal tabs with arrow overflow that's active.  The developer can also elect to turn off scrollIntoView when they call the `selectTab` function.

To test, go to the Tabs page, and set the Horizontal Tabs example to "Arrows" overflow.  Then click the "Select Last Tab" button.

closes #2042

### DIFF
--- a/projects/cashmere-examples/src/lib/tabs-horizontal/tabs-horizontal-example.component.html
+++ b/projects/cashmere-examples/src/lib/tabs-horizontal/tabs-horizontal-example.component.html
@@ -51,8 +51,8 @@
     </hc-tab-set>
 </div>
 <br>
-<button hc-button title="Set Selection" buttonStyle="primary" (click)="tabSet.selectTab(1)">
-    Select Complete Tasks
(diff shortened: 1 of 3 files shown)


bskeen (approved):
Looks pretty good to me overall.

PR #2045, opened 2023-02-13
Tabs more fixes

### LINKED ISSUE #1965: Small issues with tabs
Tabs with arrows could do with a little finessing

- ~If you programmatically select a tab that isn't in view it doesn't scroll that tab into view. You can try this out in the docs example~
- ~If you select a tab that is partially in view in doesn't scroll to pull it completely in view~
- It only recalculates the widths if the totalWidth is zero or number of tabs changes - we have an option to edit the name of the tab and to add an icon to it at certain times - I have tricked it by setting the totalWidth to 0 but that is probably not a good thing!
- Sometimes the right arrow doesn't disable when you scroll to the end.  Researching that I discovered that leftScroll is not an integer so it rarely equals the exact end number for comparison.
- If you need to work with widths you have to wait for the tab to be rendered before it has a real width to work with. It might be worth making refreshTabWidths an observable so it can wait and the user can listen for it to be done.

### LINKED ISSUE #1998: Overflow Setting on Cashmere Examples
**Summary**

The tabs being used for Cashmere examples don't have an overflow state set. We should think about which one would be best and update the component properties.  

**Reproduction**

Steps to reproduce:

1. Add many files to a Cashmere example.
2. See that the tabs don't get set to the more menu or the arrows overflow.

**Additional context**

image

andrew-frueh:
Improvements to overflow width detection algorithm

closes #1998
closes #1965


corykon (approved):

projects/cashmere/src/lib/tabs/tab-set/tab-set.component.ts
```diff
@@ -336,28 +348,27 @@ export class TabSetComponent implements AfterContentInit {
         this._tabArrowsEnabled[1] = this._tabBar.nativeElement.offsetWidth === this._tabBar.nativeElement.scrollWidth - this._tabBar.nativeElement.scrollLeft ? false : true;
```
jacquie, line 348: This is the line that was not working for me to determine if the right arrow should be enabled or not.
The problem was that, although the scrollLeft value was supposed to be an integer, it wasn't so the `offsetWidth `was never exactly equal to the `this._tabBar.nativeElement.scrollWidth - this._tabBar.nativeElement.scrollLeft`.

jacquie (approved):
Looks good to me

PR #2046, opened 2023-02-13
fix(tabs): tabClickHandler should emit Event

### LINKED ISSUE #2044: Tab Component tabClick event
**TabClick event without data emitted**

Looking into the documentation for the TabComponent, it says that the `tabClick` event is emitting the `Event` interface from the DOM. This is not correct and in the tab component file, we can confirm that nothing is being emitted.
`projects\cashmere\src\lib\tabs\tab\tab.component.ts`

So, I don't know if the correct solution is to expose the `Event` from the DOM or just define the `tabClick` to be an `EventEmitter<void>`.

andrew-frueh:
Hey @juanslinger - fortunately this one was an easy fix.  I updated the Horizontal Tabs example slightly to demo the updated functionality.  Let me know if the change looks good to you.

...and as promised, $5 motivosity coming your way for finding a tabs bug/fix to add to the next release!  Thank you!!

closes #2044

### DIFF
--- a/projects/cashmere-examples/src/lib/tabs-horizontal/tabs-horizontal-example.component.html
+++ b/projects/cashmere-examples/src/lib/tabs-horizontal/tabs-horizontal-example.component.html
@@ -20,7 +20,7 @@
                 Use the <code>maxWidth</code> parameter on a tab to constrain the width of the title.
             </div>
         </hc-tab>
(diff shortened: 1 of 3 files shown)


juanslinger (approved):
Looking good the changes, thanks for working on them.

PR #2048, opened 2023-02-14
fix(form-field): remove DoCheck for invalid checks

### LINKED ISSUE #1888: Remove ngDoCheck form dependency
**Summary**

Many of our components that support `formControl` use `ngDoCheck` as a way to check if the error state on the control has changed.  It was built that way because there wasn't a good way to subscribe to changes on `ngControl`.  That was several years ago, I wonder if there is a better way now that doesn't involve a function that runs many times a second.

Components this affects:

- checkbox
- drawer
- input
- radio
- select
- slide-toggle

andrew-frueh:
switch from DoCheck to a getter for determining if a form field is valid

closes #1888

### DIFF
--- a/projects/cashmere/src/lib/checkbox/checkbox.component.ts
+++ b/projects/cashmere/src/lib/checkbox/checkbox.component.ts
@@ -11,7 +11,6 @@ import {
     ViewChild,
     Self,
     Optional,
-    DoCheck,
     ContentChildren,
     ViewEncapsulation,
 } from '@angular/core';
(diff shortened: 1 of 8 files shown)


corykon:
I'm open to it. Getter's with heavy logic aren't my favorite, but I don't know that they'd be worse than ngDoCheck. Have you done any testing to see if the functions are called more/less?

andrew-frueh:
That's a great idea, I was trying to figure out how to gauge performance, but volume of calls is probably a good way to do that.  Will test.

Issue #1236, opened 2020-04-27
Echarts Theme


corykon:
Since so many teams are using echarts, might be nice to have a shared theme.

This might be a good place to start: https://www.echartsjs.com/theme-builder/
(In Chinese, but Chrome does a decent job of translating.)

sheaivey:
Just adding a note here to remind my self to look into adding some docs once I consolidate some info on Apache ECharts.

New theme Builder: https://echarts.apache.org/en/theme-builder.html

sheaivey:
Just to follow up the "new theme builder" linked in the previous comment is not new and only works for v4.x Apache eCharts. Unfortunately there was not one for the latest version.

I have created HealthCatalyst light/dark themes for 4.x and 5.x. I'm going to attempt to attach them here for reference along with some screen grabs for a example report dashboard. 

<img src="https://user-images.githubusercontent.com/3945263/217679214-5e7c6401-53e0-4d9a-a665-f3b1d7f6a302.png" width="250" /> <img src="https://user-images.githubusercontent.com/3945263/217679216-a4c7f9cc-afc3-4b0a-b7db-de0cf0cd7b0e.png" width="250" />

**v5.x HealthCatalyst Apache eChart Themes**
[HealthCatalyst-light-echart-v5.x.project.txt](https://github.com/HealthCatalyst/Fabric.Cashmere/files/10690926/HealthCatalyst-light-echart-v5.x.project.txt)
[HealthCatalyst-dark-echart-v5.x.project.txt](https://github.com/HealthCatalyst/Fabric.Cashmere/files/10690924/HealthCatalyst-dark-echart-v5.x.project.txt)

**v4.x HealthCatalyst Apache eChart Themes**
[HealthCatalyst-light-echart-v4.x.project.txt](https://github.com/HealthCatalyst/Fabric.Cashmere/files/10690925/HealthCatalyst-light-echart-v4.x.project.txt)
[HealthCatalyst-dark-echart-v4.x.project.txt](https://github.com/HealthCatalyst/Fabric.Cashmere/files/10690923/HealthCatalyst-dark-echart-v4.x.project.txt)

_The above text files are json safe._

andrew-frueh:
@sheaivey the themes look absolutely incredible!  Amazing work.  Excited to review all this on Monday and get it all added to the site/library formally.

PR #2050, opened 2023-02-18
#1236 Adding Apache Echarts Documentation

sheaivey:
I think this is ready to go. My Angular is rough at best but everything appears to be working as expected.
I added an interactive chart demo to show off the theme and expose its configuration.

### DIFF
--- a/src/app/foundations/chart/chart-demo.component.ts
+++ b/src/app/foundations/chart/chart-demo.component.ts
@@ -1,0 +1,194 @@
+import {Component} from '@angular/core';
+import {SectionService} from '../../shared/section.service';
+import {BaseDemoComponent} from '../../shared/base-demo.component';
+import { SelectChangeEvent } from '@healthcatalyst/cashmere';
+import {ViewChild, ElementRef} from '@angular/core';
+import * as echarts from 'echarts';
+import * as hcChartLight from '../../../assets/foundations/charts/healthcatalyst-light-echart-v5.x.json';
+import * as hcChartDark from '../../../assets/foundations/charts/healthcatalyst-dark-echart-v5.x.json';
+
+const exampleData = {
+    simple: {
+        legend: undefined,
+        labels: ['Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun'],
+        series: [
+            {
+                name: "Week 1",
+                data: [150, 230, 224, 218, 135, 147, 260],
+            }
+        ],
+    },
+    grouped: {
+        labels: ['Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun'],
+        series: [
+            {
+                name: "Week 1",
+                data: [150, 230, 224, 218, 135, 147, 260],
+            },
+            {
+                name: "Week 2",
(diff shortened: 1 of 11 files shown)


src/app/foundations/chart/chart-demo.component.ts
```diff
@@ -0,0 +1,194 @@
...
+
+    changeChartType(chartType: SelectChangeEvent): void {
+        this.chartType = chartType.value;
+        this.updateChart();
+    }
+    
+    changeChartTheme(chartTheme: SelectChangeEvent): void {
+        this.chartTheme = chartTheme.value;
+        echarts.dispose(this.chartContainer.nativeElement);
+        this.ngAfterViewInit();
```
sheaivey, line 177: Calling `this.ngAfterViewInit()` seems like bad practice. It would be better to create a `initChart()` and call it here and inside `ngAfterViewInit()`.

src/app/foundations/chart/chart-demo.component.ts
```diff
@@ -0,0 +1,194 @@
...
+@Component({
+    selector: 'hc-chart',
+    templateUrl: './chart-demo.component.html',
+    styleUrls: ['./chart-demo.component.scss']
+})
+
+export class ChartDemoComponent extends BaseDemoComponent {
+
+    @ViewChild('chartContainer') chartContainer:ElementRef;
+    chartTheme = 'light';
```
sheaivey, line 87: After reading a little more the data binding to the input controls is not correct.  I think `@Input` is what I should have used.

src/app/foundations/chart/chart-demo.component.ts
```diff
@@ -0,0 +1,194 @@
+import {Component} from '@angular/core';
+import {SectionService} from '../../shared/section.service';
+import {BaseDemoComponent} from '../../shared/base-demo.component';
+import { SelectChangeEvent } from '@healthcatalyst/cashmere';
+import {ViewChild, ElementRef} from '@angular/core';
+import * as echarts from 'echarts';
+import * as hcChartLight from '../../../assets/foundations/charts/healthcatalyst-light-echart-v5.x.json';
+import * as hcChartDark from '../../../assets/foundations/charts/healthcatalyst-dark-echart-v5.x.json';
```
sheaivey, line 8: These JSON theme files are static and hard coded. In our react component library I set these chart themes up to be dynamically created based on the currently loaded theme. I was not sure of an easy way to provide that functionality from the docs. Maybe a charting component in the future would be a good candidate for dynamic chart themes.

andrew-frueh (approved):
Amazing work @sheaivey!  Good callouts in your notes, but most of those are stylistic things more than anything - and you did great with the Angular!  I added one commit with two things to callout:

- I moved the charts docs to the "Styles" section under the Web Apps tab.  Generally I think of the Foundations tab as the things that are universal across all our modalities.  This one is pretty specifically web development, so I think it fits well in Styles - I do like having it in its own section as you built it.  And someone can always search for it.
- The only tweak I made to your code is in the component styles, I loaded the color stylesheet and switched the hard-coded hex value to a sass variable.  We try to avoid hard-coded hex anywhere possible in the library just in case branding stuff ever changes in the future.

corykon (approved):

PR #2062, opened 2023-02-28
fix(tab-set): fix default tab with routing

### LINKED ISSUE #2061: Tabs component strips query params from URL
**Summary**

I've been noticing that query parameters are getting stripped from URLs in the app in two situations:

1. Reloading the page
2. Clicking from one tab to another

This only happens on pages where we use hc-tab-set and hc-tab with `[routerLink]` parameters. I believe these components to be the root cause after a couple of hours of debugging in our app. (We are using the `[queryParams]` parameter on every tab to preserve whatever query params are currently in the URL, so this is especially troubling.)

I'll work on an MCVE today.

**Reproduction**

1. Use hc-tab-set and hc-tab components with `[routerLink]` paths.
2. Manually set some query parameters when the user clicks a button.
3. Reload the page or click to another tab. The query parameters disappear.

Stackblitz reproduction: (WIP)

**Additional context**

Cashmere v13.7.0 and v13.7.1 both have this behavior.

andrew-frueh:
resolve several issues relating to setting the default tab based on url

@isaaclyman - these were some desperately needed fixes.  There was a lot of strange stuff going on in tabs with routing.  One huge win here is that on Cashmere site, for all the other pages in addition to Components, deep-linking now works (it doesn't always bounce you back to the first page in the list).

One very strange thing is we were using a `BehaviorSubject` to fire when `SelectTab` changed - but one property of BehaviorSubject is to fire on init.  So SelectTab was always getting called with a 0 value.  Seems like defaultTab would get called after it, but it was fixing something that shouldn't have been happening in the first place.  So switched that to `Subject` which is much better suited to this situation.

The one known issue is that `defaultToFirstRoute` is going to struggle with relative paths.  For example, on the Components page, all those tabs just have `Examples`, `API`, or `Usage` for their routerLink.  And the function doesn't know how to resolve that relative path into an absolute path to match with the current route.  I really tried to think of a way to fix that, but I was hitting a wall.  It's no worse than the existing state of the component, but I think with relative paths until we think of something better - we have to rely on the devs making good use of `defaultTab`.

Total side note that I noticed search wasn't indexing some content pages correctly, so that's the fix in this PR on `search-prep`. Also noticed we had a really strange example for tabs - `SelectedTabInputExample`.  So I cleaned up the code in that example to better reflect best practices.

closes #2061


andrew-frueh:
Good news @isaaclyman - I came back after dinner with fresh eyes and I think I've got it much more solid now.  It's now resolving relative urls, and it's only navigating if it's absolutely necessary.  These changes actually allowed me to undo a lot of code on the Cashmere site that was basically covering over this strange tab routing behavior.  So hopefully it'll allow you to do the same in your app as well (like removing the `defaultTab='none'` stuff.

Would feel a lot better though if you could give these changes a good look over and make sure everything makes sense to you.  Any suggestions to improve greatly appreciated!

andrew-frueh:
...and one last change, since that `BehaviorSubject` isn't automatically firing on load, we don't need the ability to disable `scrollIntoView` on `selectedTab` change events.  It was causing weirdness by trying to scroll the page on load with the old way, but now that it's only called when that event is triggered, it's appropriate to scroll the selected tab into view if it was programmatically selected.

projects/cashmere/src/lib/tabs/tab-set/tab-set.component.ts
```diff
@@ -499,29 +506,48 @@ export class TabSetComponent implements AfterContentInit {
...
-                return currentRoute === routerLink || currentRoute.indexOf(`${routerLink}/`) > -1;
-            });
+        const tabArray = this._tabs.toArray();
+        let routeTab = -1;
+
+        // Determine if there is a routerLink in our tabset that matches the current route
+        for ( let i=0; i < tabArray.length; i++ ) {
+            const routerLink = this.mapRouterLinkToString(tabArray[i].routerLink);
+            const currentRoute = this.router.url.split("?")[0];
+            if( currentRoute === routerLink ) {
```
isaaclyman, line 516: We should consider doing a case-insensitive comparison here.

andrew-frueh: Good call - probably just `toLowerCase()` on both sides?

projects/cashmere/src/lib/tabs/tab-set/tab-set.component.ts
```diff
@@ -499,29 +506,48 @@ export class TabSetComponent implements AfterContentInit {
...
+        // Combine url arrays
         if (routerLink instanceof Array) {
             routerLink = routerLink.join('/').replace('//', '/');
         }
+        // Resolve relative urls
+        if ( !routerLink.startsWith('/') ) {
+            const currentRoute = this.router.url.split("?")[0];
+            const routeParts = currentRoute.split('/');
+            routeParts.shift();
+            routeParts.pop();
```
isaaclyman, line 548: This bit of logic seems a little fragile. What if someone's using nested tabs?

andrew-frueh: I _think_ it shouldn't matter because what I'm trying to do there is construct a (mostly) full path from a relative path.  So I'm grabbing the current url and stripping the first segment ("https://") and the last segment (which gets combined with the relative path).  So nested tabs with relative paths are still relative to the current url, no?

isaaclyman: `this.router.url` doesn't yield the same thing as `window.location.href`. For example, testing on one of the pages in PowerPlanning:

window.location.href: http://localhost:4300/budgeting/expenses/general-ledger
this.router.url: /budgeting/expenses/general-ledger

So once you shift and pop (the next great dance move in America's urban clubs, probably), you'd end up with `['expenses']`.

Imagining we had an outer tab set that navigates between "expenses" and its siblings (e.g. "volume", "spread") and an inner tab set that navigates between "general-ledger" and its siblings (e.g. "summary", "department"), would this method work as expected for both of them?

andrew-frueh: I liked the shift and pop too 😆 Fortunately I was able to test this a bit last night because the Components section uses relative paths on the Examples, API, Usage tabs.  And because of that first `/` the first element of the split array was an empty string (hence the need for the shift).  So actually I think for PowerPlanning you'd get `budgeting/expenses` - which is what we want to put the relative path at the end of.

I'll try to set up the scenario you mentioned with inner and outer tabs to validate that.

isaaclyman: Oh, you're right. I messed up the mental calculation on that split.

isaaclyman: Have you looked into `ActivatedRoute.snapshot.url`? It's sensitive to the current component's position in the router tree, and returns only the route segment relevant to it.

andrew-frueh: Good thought - I was thinking `ActivatedRoute` might be a better option than `Router` here.  I'll give the snapshot url a look.

projects/cashmere/src/lib/tabs/tab-set/tab-set.component.ts
```diff
@@ -187,6 +187,24 @@ export class TabSetComponent implements AfterContentInit {
...
 
+        // If using routing and routerLinkActive hasn't set any active tabs, default to the first route
+        if ( this._routerEnabled ) {
+            setTimeout(() => {
+                let noActiveTabs = true;
+                this._tabs.forEach(t => {
+                    if ( t._active ) {
+                        noActiveTabs = false;
+                    }
+                });
```
isaaclyman, line 198: Doing some unnecessary iterations here. You could replace all of this with:

```ts
const noActiveTabs = !this._tabs.some(t => t._active);
```

projects/cashmere/src/lib/tabs/tab-set/tab-set.component.ts
```diff
@@ -405,6 +423,14 @@ export class TabSetComponent implements AfterContentInit {
         this._stopTabSubscriptionSubject.next();
         this._tabs.forEach(t => t.tabClick.pipe(takeUntil(this._stopTabSubscriptionSubject)).subscribe(() => this._setActive(t)));
 
+        // If using tabs with routing, listen for changes to the routerLinkActive state
+        if ( this._routerEnabled ) {
+            this._tabs.forEach(t => t._routerActiveChange.pipe(takeUntil(this._stopTabSubscriptionSubject)).subscribe(tab => {
```
isaaclyman, line 428: Does `_routerActiveChange` only fire when the tab becomes active? Or does it also fire when the tab becomes inactive?

isaaclyman: Never mind, I can see in the next file that it only fires when status is `true`. Carry on.

isaaclyman (approved):

PR #2067, opened 2023-03-02
feat(tab-set): selectedTab priority

andrew-frueh:
selectedTab takes priority on init over defaultTab or the current url

The work continues @isaaclyman 😁 some cleanup here for docs purposes, but primarily adjusting as you suggested to make `selectedTab` a write-only property that only the user can set, not something that the library will be setting on its own.  The only exception is calling the `selectTab()` function - which does fire a `selectedTabChange` event.  

`selectedTab` is now the highest priority input on init, it takes priority over `defaultTab` or the url (in routing tabs).

Appreciate your input as always!


bskeen (approved):

projects/cashmere/src/lib/tabs/tab-set/tab-set.component.ts
```diff
@@ -187,7 +188,7 @@ export class TabSetComponent implements AfterContentInit {
         this.setUpTabs(false);
         this.refreshTabWidths();
 
-        // If using routing and routerLinkActive hasn't set any active tabs, default to the first route
+        // If using routing, selectTab isn't set, and routerLinkActive hasn't set any active tabs, default to the first route
```
isaaclyman, line 191: Is `selectedTab` meant here? I don't know what `selectTab` refers to.

andrew-frueh: yup, that's a typo

PR #2093, opened 2023-03-24
feat(select): placeholder value

### LINKED ISSUE #2092: Show placeholder in hc-select for any falsy value
**Describe your idea**

`hc-select` should show its placeholder text when its value is _any_ falsy value (null, undefined, false, 0, or empty string), not just the empty string.

I created an hc-select with a placeholder, then added an `[(ngModel)]` to it with the initial value set to `null`. I was surprised when the placeholder didn't show up. I then tried setting the value to `undefined`, but the placeholder again didn't show. I had to dig into the Cashmere component template to learn that the only situation where the placeholder shows is if the value of the hc-select is the empty string `''`.

**Additional context**

I'm not completely sold on this, just putting it out there for discussion. Maybe there are situations where an `<option>` has a value of `null` or `undefined`, in which case that option should be selected instead of showing a placeholder. But it's a little counterintuitive to have to set an empty string instead of `null` on a `<select>` with no value.

andrew-frueh:
adds the ability to set the value of the placeholder option

What do you think of this approach @isaaclyman?  Avoids any potential for a breaking change, but also allows for a developer have control over which falsey value to support.  And most important, provides better visibility into how the component will behave.  To that end, I adjusted an existing example that wasn't providing much value and repurposed it to demo this functionality (because as we know, without an example, it's like the functionality isn't there).

closes #2092


isaaclyman:
Looks great. Thanks!

isaaclyman (approved):

PR #2094, opened 2023-03-24
fix(tooltip): disable empty tooltips

### LINKED ISSUE #2079: Make hcTooltip support conditional expressions
**Describe your idea**

It would be cool if `[hcTooltip]` could be conditionally rendered.

We have situations where a tooltip is needed in some states but not others. It's natural to try to do:

`[hcTooltip]="shouldShowTooltip ? 'You can't do that right now.' : null"`

Or 

`[hcTooltip]="shouldShowTooltip ? 'You can't do that right now.' : undefined"`

But in both cases, an empty tooltip is still shown (see below). It would be great if the tooltip only showed when there was content in it.

**Additional context**

image

andrew-frueh:
change trigger to none for tooltips with empty content

...and one more for you @isaaclyman - this was an easy fix but I like it a lot.  Gives the developer another option for disabling tooltips besides always having to control `trigger`.

closes #2079


isaaclyman (approved):
Amazing, thank you so much

Issue #1706, opened 2021-11-03
New component: Confirmation Modal


corykon:
Confirmation modals are super common in our apps. We already have generic modals, but we should set up a standard component specifically for this use case to make sure the confirmation experience is aligned.

Leading Wisely just made a nice confirmation modal component that we could port over:
![image](https://user-images.githubusercontent.com/12416432/140190806-c1f89ab0-cf3a-4099-a05e-cb32e6137c80.png)

![image](https://user-images.githubusercontent.com/12416432/140190961-9f225f42-413a-42f7-a003-69d0f0c1509e.png)

![image](https://user-images.githubusercontent.com/12416432/140190239-89a32bf1-d52b-47f8-8e0f-7d7135df3dde.png)

![image](https://user-images.githubusercontent.com/12416432/140190182-03c7fc7e-cded-4543-934e-61ddbd893038.png)

corykon:
@andrew-frueh thoughts on the styles? Wondering if the spacing/sizing could be dialed in. Some cashmere and non-cashmere variations

![image](https://user-images.githubusercontent.com/12416432/140191981-2188fd87-be53-4aed-9c14-f9bc3e4fac08.png)
![image](https://user-images.githubusercontent.com/12416432/140191522-f9a73f7a-6ec8-49a4-88f8-6db8beef3022.png)
![image](https://user-images.githubusercontent.com/12416432/140191550-68a55ece-cc52-4b4e-8259-899c832af871.png)

On this one - don't like that the buttons are both blue, but love how minimalist it is. Rather than header and a message, sometimes just a one liner and the action buttons are enough.
![image](https://user-images.githubusercontent.com/12416432/140191731-79e0e01f-f8bd-4f69-b60c-01295b69fa49.png)

corykon:
Note: 
- Option to have no header
- Double check styles work well for small screens

andrew-frueh:
@corykon - if you end up building this one, make sure to include the FocusTrap enhancements/options just added to modal in #1747

corykon:
Pick this up to work on now, for a CDP use case.

PR #2099, opened 2023-03-27
feat(confirmation modal): add confirmation modal service

corykon:
Provide a minimal code experience for common modal use cases.

Just inject the ConfirmationModalService, and write a bit of code:

```

        this.confirmService
            .confirm({message: 'Are you sure you want to do XYZ?'})
            .pipe(
                take(1),
                filter(confirmed => confirmed)
            )
            .subscribe(() => {
                // go do the thing
            });
```

Also has a `confirmDestructive()` function with reasonable defaults for a destructive actions.

image

re #1706


isaaclyman:
This is awesome. We've built a ConfirmationModalComponent for our app, but yours has all the same options we've got and then some.

Only one suggestion: could there be a `white-space: pre-wrap` on the modal body so that newlines in the `message` parameter are preserved, for easy multi-paragraph messages?

corykon:
great idea @isaaclyman!

andrew-frueh (approved):
Looking great, well done @corykon - aside from that one small style tweak everything is good to go by me.  I did have a question about services - is it standard practice to make something like this a separate service, as opposed to wrapping it into the existing ModalService?  Is it better to keep services more modular?  I can certainly see instances where I'd need to pull both ModalService and ConfirmationService in my constructor, so that's mainly why I ask.

andrew-frueh:
That's a good call @isaaclyman - I think that'd be a pretty simple addition to your `options` @corykon and it expands the use case here.

andrew-frueh:
> > Looking great, well done @corykon - aside from that one small style tweak everything is good to go by me. I did have a question about services - is it standard practice to make something like this a separate service, as opposed to wrapping it into the existing ModalService? Is it better to keep services more modular? I can certainly see instances where I'd need to pull both ModalService and ConfirmationService in my constructor, so that's mainly why I ask.
> 
> Yeah, great question and a valid argument. I like to keep services (and all files, really) smaller, more modular, and focused on one job. Angular style guide talks about it a bit - https://angular.io/guide/styleguide#single-responsibility-1

Ah that Angular Style guide is really helpful.  And I think following that reasoning, you've done the right thing here 👍

corykon:
@andrew-frueh @isaaclyman After adding support for alerts, should I rename the service to something more generic?

SimpleModalService
QuickModalService

Or just merge it into the standard ModalService like @andrew-frueh suggests?

corykon:
@isaaclyman FYI I added a `preserveMessageWhitespace` parameter. I was going to default to true, but decided against it because it's easy to accidentally introduce an extra space if you're not anticipating it. Felt like defaulting to false would be the less surprising to most devs

PR #2109, opened 2023-03-30
feat(modal): close icon

### LINKED ISSUE #2000: Simple Modal Setting
**Describe your idea**

We're using modals with only a title, x-icon, and small footer in some places (instead of buttons in the footer). It would be awesome to have a setting on our modals that automatically applies all of these styles/functions to a modal.

No properties or examples for creating simple modals in Cashmere.

**Additional context**
Example from HAS app 
Screen Shot 2022-11-28 at 10 10 14 AM

andrew-frueh:
adds option for modal close icon in top right

closes #2000


projects/cashmere/src/lib/modal/modal.component.ts
```diff
@@ -14,16 +13,6 @@ import { ModalHeaderComponent } from './modal-header.component';
...
-    private _tight = false;
-
-    /** If true, condense the default padding on all elements and reduce the title font size. *Defaults to `false`.*  */
-    @Input()
-    get tight(): boolean {
-        return this._tight;
-    }
-    set tight(value: boolean) {
-        this._tight = parseBooleanAttribute(value);
-    }
```
andrew-frueh, line 26: Turns out this input param actually does nothing 😬 looks like this was maybe added during development but never removed.  The only way to add tight to a modal is via `ModalOptions` like everything else.  So this is just confusing to developers to have a zombie input.

corykon (approved):

projects/cashmere/src/lib/modal/modal.component.ts
corykon: Ooh good catch

projects/cashmere-examples/src/lib/modal-overview/modal-overview-example.component.ts
```diff
@@ -18,7 +18,7 @@ export class ModalOverviewExampleComponent {
         const options: ModalOptions = {
             data: {
                 content: 'This is a size lg modal. If isDraggable is set to true, it places a drag handle in the top right corner which you can use to reposition the modal. If isResizable is set to true, a resize handle appears in the bottom right to change the dimensions of the modal. They may be used together or separately.',
-                tight: false
```
andrew-frueh, line 21: This is a case of confusing documentation.  Here we are passing an object to the example modal's component, and we're trying to tell it whether it's a tight modal or not to display appropriate content.  But we used a boolean flag with the same name (`tight`) as the actual setting - so it looks very confusing to a developer whether you set tight in the modal's data outside of it.  So that's why I just changed the object key name to be less misleading.

projects/cashmere-examples/src/lib/modal-simple/modal-simple-example.component.html
```diff
@@ -1,7 +1,6 @@
 <p>To implement simple modals with minimal code, use <code>alert()</code>, <code>confirm()</code>, or <code>confirmDestructive()</code>:</p>
 
-<pre><code>
-this.modalService.confirm(options)
```
andrew-frueh, line 4: @corykon `<pre>` is really strict with newlines 😄 so you were getting an extra line at the top of your codeblock that you didn't need

PR #2112, opened 2023-03-31
docs(tooltip): scroll behavior guide

### LINKED ISSUE #1371: Tooltip stays open on scroll
Within a scrollable div, tooltips stay open when you scroll away.

cashmere-tooltip-bug

andrew-frueh:
@corykon - added guidelines around the use of `cdkScrollable` to the pop usage doc and also the tooltips example.  I also added the `cdkScrollable` directive to our components container, so it looks to me like the tooltips example behaves much better on scroll now.  I can see the change in Chrome, but interestingly this isn't an issue in Firefox.  Let me know if you're seeing the same.

closes #1371

### DIFF
--- a/projects/cashmere-examples/src/lib/tooltip-overview/tooltip-overview-example.component.html
+++ b/projects/cashmere-examples/src/lib/tooltip-overview/tooltip-overview-example.component.html
@@ -1,7 +1,12 @@
-This is an example of a bottom-aligned
-<u hcTooltip="I'm a tooltip" verticalAlign="below">tooltip in action</u>.
+This is an example of a bottom-aligned <u hcTooltip="I'm a tooltip" verticalAlign="below">tooltip in action</u>.
 <br><br>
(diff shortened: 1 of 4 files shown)


corykon (approved):
This looks great :+1:

Documentation around tweaking animations is nice, too. Good add.

andrew-frueh:
Ah I can't take the credit on tweaking animations - that section was actually just out of order.  At some point when we added some new documentation that looks like it got bumped to the bottom of the page.  So in this case I was just shoving it back to where it belongs :-) 

One other useful thing to call out @corykon in case you ever need it (I had forgotten this) - to direct link to a section on usage docs, you can use the url param `section` and it'll scroll down after load.  For example:
`/web/components/pop/usage?section=scroll-behavior`.

corykon:
Haha oh gotcha!

I forgot about that as well - thanks for calling out. I tried to use the
#id syntax recently and was bummed to find it didn't work.

PR #2118, opened 2023-04-06
fix(modal): revised drag handle

### LINKED ISSUE #2110: Draggable Header on Modal
**Describe your idea**
Change the modal to support draggable via the header area instead of the icon.

When we have an x-icon in the righthand corner of the modal header, the draggable icon flips to the other side of the modal (to the left of the title). Instead, we should make the header area draggable with an appropriate cursor on hover.

andrew-frueh:
changes drag handle to a grip always positioned on the top left

closes #2110


corykon (approved):
ship it! 🚀

cikeddy (approved):
Checked it out and tested the draggable with tight. Looks good to me!

PR #2119, opened 2023-04-10
fix(input): mark as touched on blur

### LINKED ISSUE #2116: Input Field Validation is Delayed
**Summary**

If a required field is left blank or invalid information is added into an input field, tabbing or clicking out of the input box does not trigger the validation error message immediately. 

**Reproduction**

Steps to reproduce:

This can be tested using the [Cashmere Input-Required example](https://cashmere.healthcatalyst.net/web/components/input/examples)
1. Tab into or click inside the required input field.
2. Leave the field blank
3. Tab out of the field or click outside of it with your mouse

The validation message does not appear at that point. It requires clicking back into the input field and then clicking out of it again, or multiple mouse clicks outside of the field before the validation message activates.

**Additional Info**
From Andrew Frueh:
"In older versions of Cashmere, there was a constant DoCheck loop that would check for changes to error state.  Now it only listens for changes to the valid status or form submit"
See [input.directive.ts](https://github.com/HealthCatalyst/Fabric.Cashmere/blob/dev/projects/cashmere/src/lib/input/input.directive.ts)
"and then also if you look at line 217 there, you’ll see that the control needs to be touched before the control will throw an error...
unfortunately there isn’t a way to listen for changes to the touched state - so on line 129 you’ll see that onBlur there is a call to check the error state again"

andrew-frueh:
force an input element to mark its ngControl as touched on blur

@JDavidStevens @corykon  - this was an interesting one because it happens in Chrome but not Firefox.  The issue is that the `ngControl` on our inputDirective isn't being marked as `touched` on blur (or it's happening after the blur event).  And with our validation step, it won't mark the control as invalid unless it's been touched.  So what I've done here is force the control to be marked as touched on blur (which it should be doing anyway) before the validation gets called.  With that change, the test you mentioned checks out.

closes #2116

### DIFF
--- a/projects/cashmere/src/lib/input/input.directive.ts
+++ b/projects/cashmere/src/lib/input/input.directive.ts
@@ -128,6 +128,9 @@ export class InputDirective extends HcFormControlComponent implements AfterViewI
 
     @HostListener('blur')
     _onBlur(): void {
+        if ( this._ngControl && this._ngControl.control) {
+            this._ngControl.control.markAsTouched();
+        }
         this._changeFocus(false);
         this._updateErrorState();
     }


corykon (approved):
that's so odd. Did you find stuff online that mentioned this difference across browsers?

PR #2125, opened 2023-04-25
fix(form-field): error check on form reset

### LINKED ISSUE #2124: Form reset is not clearing the control error status
**Summary**

By using the form field and having specified validators, when executing `formGroup.reset()` or `formControl.reset()` the field is cleared but it continues validating so the control gets in an error state.

**Reproduction**

Steps to reproduce:

1. Create a form group with a control which will be associated to an input and specify the required validator.
2. In the template create a form and connect it to the form group and specify in the input the form control name.
3. Add a button and associate the click with a function that will reset the formGroup.
4. When pressing the mentioned button, the information will be cleared but the control will be outlined in red, because the form is invalid.

https://angular-qb5inf.stackblitz.io

andrew-frueh:
make sure error states are reset have form resets

closes #2124

@guille-gomez this one was a real mind bender, notes below (I've also updated an example on the inputs page with these notes):

- The big trick here is that you have to make sure after a form submit you are resetting your form to submitted = false.  As best as I can tell, there are two ways to do that - I show both of them in the example.  You can use resetForm() on NgForm.  Or you can include a button inside your form with `type="reset"`.  The second is old school HTML, but it still does the trick and removes submitted from your form.
- That said, I did have to add a `setTimeout` to all our ngControl.statusChanges subscriptions, because we need to be sure that the error state check is happening after submitted states have been updated.  I think this was tripping us up today because our calls were happening before the submitted update had been applied.

I've been able to validate the above works on the Cashmere site, and I think it matches your use case.  Let me know what you think.

### DIFF
--- a/projects/cashmere-examples/src/lib/input-required/input-required-example.component.html
+++ b/projects/cashmere-examples/src/lib/input-required/input-required-example.component.html
@@ -1,10 +1,13 @@
-<form>
+<form [formGroup]="exampleFormGroup" #exampleForm="ngForm">
     <div class="form-container">
         <hc-form-field>
             <hc-label>Validate an email address:</hc-label>
-            <input hcInput [formControl]="formDemo" required>
+            <input hcInput formControlName="exampleInput" required>
             <hc-error>A valid email address is required</hc-error>
         </hc-form-field>
     </div>
+    <button hc-button title="Reset Form" buttonStyle="secondary" type="reset">Form Reset</button>
     <button hc-button title="Submit Form" buttonStyle="primary" type="submit">Submit</button>
 </form>
+
+<button hc-button title="NgForm Reset" buttonStyle="secondary" (click)="resetForm()">NgForm Reset</button>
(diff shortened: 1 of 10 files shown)


guille-gomez (commented):
I tried both solutions suggested here both of them worked (without adding the setTimeout).

I'd also suggest to consider updating this
```       if ( this._ngControl && this._ngControl.statusChanges ) {
            this._ngControl.statusChanges.pipe(takeUntil(this._unsubscribe)).subscribe(() => {
                // setTimeout is necessary to make sure any form or control state changes have been applied before rechecking error states
                setTimeout(() => {
                    this._updateErrorState();
                });
            });
        }
```
with this
```
        if ( this._ngControl?.statusChanges ) {
            this._ngControl.statusChanges.pipe(
                delay(0), // This rxjs operator is similar to setTimeout
                takeUntil(this._unsubscribe)
            ).subscribe(() => (this._updateErrorState())
        }
```

andrew-frueh:
Appreciate the suggestion @guille-gomez!  Updated the PR to leverage the rxjs operator instead of `setTimeout`.  Can you give the PR one more look over and approve if all looks good?

guille-gomez (approved):
These changes look good to me, since I have already tried the two solutions and they both work. Thank you also @andrew-frueh for taking into consideration my suggestion.

PR #2137, opened 2023-05-03
docs(discovery-service): load cashmere icons

andrew-frueh:
@corykon now that we've got the base64 icons available, may as well simplify the mock discovery service to pull those in.  I had to string together three http requests, not sure if there is a better way to do that?


corykon (approved):
That's cool. I can't think of a better way to pull those in unless we had all the base64's merged into one JSON file. Probably not worth doing just for the sake of this example. I'd leave it as is!

PR #2141, opened 2023-05-04
fix(sidenav): fix active tab styles and add keyboard accessibility

corykon:
finding more little things to tweak on sidenav


andrew-frueh (approved):
Nothing like testing a component in the wild to find all the missing pieces.  I did the same thing with the file uploader as I was working it into the KU app.

corykon:
soooo true

PR #2148, opened 2023-05-10
fix(scrollnav): remove scrollintoview usage

### LINKED ISSUE #2145: Scroll Nav will scroll outside containers out of view
**Summary**

When clicking on an item in the Scroll Nav, everything outside the scroll nav scrolls up in addition to the scroll nav content. This can scroll tabs out of view with no way of scrolling back if scrolling is disabled in the container (or if the container is a fixed size). In the stackblitz reproduction below, clicking on any of the links in the scroll nav will scroll part of stackblitz off of the screen in addition to the content in the scroll nav.

**Reproduction**

Steps to reproduce:

1. Add a scroll nav to a page (probably one that has containers/content outside of the scroll nav)
2. Click on any of the links
3. The page will scroll to put the scroll nav at the top of the screen, regardless of where the scroll nav is.

https://stackblitz.com/edit/angular-cfjano?file=src/app/scroll-nav/scroll-nav-example.component.scss

**Additional context**

scrollnav
scroll_nav_stackblitz

andrew-frueh:
@bskeen @corykon - I untangled the two instances of `scrollIntoView` that the scrollNav component was using.  And Ben I 100% agree that this isn't the place to use that function - the scrollNav shouldn't ever be scrolling the page it's a part of, only itself.  The component is split into multiple pieces, so it required some additional emitters and listeners to get the information to the right place, but as best as I can tell it's working as expected.  I also updated the unit tests to account for the updated approach.

closes #2145

### DIFF
--- a/projects/cashmere/src/lib/sass/scroll-nav.scss
+++ b/projects/cashmere/src/lib/sass/scroll-nav.scss
@@ -96,6 +96,7 @@
 }
 
 @mixin hc-scroll-nav-content-container() {
+    position: relative;
     max-height: 100%;
     overflow: scroll;
     overflow-x: hidden;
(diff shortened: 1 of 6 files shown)


bskeen (approved):
It all looks good to me. Thanks for looking at this.

PR #2152, opened 2023-05-15
Resizable Container Component

### LINKED ISSUE #2001: Incorporate VW Splitter into Cashmere
**Describe your idea**

An engineer on the VitalIntegrity team built a splitter (re-sizing) component that we might be able to pull into Cashmere.

**Additional context**

https://user-images.githubusercontent.com/46426253/204389566-f75aa0f1-d5e9-4a5e-8159-12281b4fdd3d.mov

andrew-frueh:
Adds the resizable container component.

closes #2001


corykon (approved):
That's really cool. Fun to play with, feels smooth.

Is it worth adding a little more guidance for implementation? If I were to try using this component, it's not clear to me which chunks of code from the example are needed. Obviously I'd need the new directive for sure, but the other containers and classes would take some trial and error to sort through. Would be cool to add some comments in the examples, a guide .md file, or another even more dumbed down example.

cikeddy (approved):
Worked well for me! Thanks for making all these updates :)

andrew-frueh:
@corykon - I 100% agree with you that the implementation would probably be a little confusing.  I was considering making two additional directives for a "container" and "static element" so that you at least had three building blocks you could use to assemble the example.  Maybe that would make everything easier?  And I'll also add a usage dock to add some additional explanation.  Let's talk in our next 1:1.

PR #2153, opened 2023-05-16
feat(multiselect): ng-select pagination

### LINKED ISSUE #2022: Pagination Ng-Select Example
**What would you add or change?**
VitalCDM is using a version of ng-select that contains pagination and refresh functions on the footer. We want to add an example into cashmere for this.

Screen Shot 2022-12-13 at 3 33 31 PM

andrew-frueh:
@HaydenDalley35 @cikeddy @corykon would love to get your thoughts on this PR that adds a new example to multiselect (ng-select) demonstrating how to paginate a long list of dropdown options.  This is something Vitalware does, and they worked up a custom controller for it, but I think it can be done in Cashmere style with the existing `hc-pagination` component.  NgSelect is amazingly versatile and I think it works pretty well with this use case.  But please let me know what you think. 

Required adding a tight param to pagination to strip padding/margin and an additional button style, which are small library updates.

closes #2022


corykon:
Cool. Clever use of some built in Cashmere components. 

I'm curious what the refresh button is for though? A tooltip or something would help.

andrew-frueh:
Tooltip is a good idea to clarify that refresh button, and I'll make sure and add additional comments to the code also.

corykon (approved):

andrew-frueh:
Thanks again for looking this over w/ my @HaydenDalley35 and signing off on the example 👍

Issue #1254, opened 2020-05-13
Add wrapper to ng-select to enable form-field functionality


jacquie:
I am not sure this is a bug, but it's not a full-blown feature either - maybe an enhancement :)

We would like the ng-select component to have a wrapper component that adds form-field functionality and in particular standard styling on validation messages

corykon:
From a slack convo, wondering if Joe's work in IDEA here can be leveraged:
https://dev.azure.com/healthcatalyst/CAP/_git/CAP?path=%2FIDEA%2FFrontEnd%2FClientApp%2Fsrc%2Fapp%2Fshared%2Fdynamic-input%2Fdynamic-input.component.ts

corykon:
This came up again while implementing CDP UI. We should prioritize this one a little higher. Discussed w/ @andrew-frueh and the design team this morning.

PR #2154, opened 2023-05-17
feat(ng-select): create a wrapper for hc-form-field use

corykon:
this change allows ng-select to work more smoothly within a Cashmere form
multiselectv-wrapped

@andrew-frueh I followed your example with the wrapper you built for the slider.
re #1254


andrew-frueh (approved):
This is a fantastic update! Just another reason now to always use ng-select instead of normal select.

Issue #1902, opened 2022-06-16
Scrollable modals


corykon:
We should add a property for enforcing a max-height on modal-body, so that super tall modals are scrollable within the viewport. 

Victor R. showed me this modal that his team got bug reports on from users:
![image](https://user-images.githubusercontent.com/12416432/174150256-ab49899b-8ea4-4d39-9d57-8149b5b0627f.png)

We want the experience to feel more like this:
![image](https://user-images.githubusercontent.com/12416432/174150807-2e9b6378-67b5-4f18-ad5e-3ac8932dece1.png)
You can do this by adding a `max-height` with overflow-y": auto on the `hc-modal-body`. Not hard to add after the fact, but might as well have it in the library.

PR #2156, opened 2023-05-18
feat(modal): add default max height on modal body

corykon:
New default height so that tall modals won't bleed off screen. 

re #1902

image


andrew-frueh (approved):
Nice, I add this _all_ the time myself anyway to modal body, so it'll be great to have it in by default.  I can't imagine why someone wouldn't want this, but I suppose they could just override the style if they needed to?

corykon:
Yeah that was my thinking - override it if you have a reason not to have it.

Issue #2028, opened 2023-01-31
Drawer Style Refresh


corykon:
A couple apps (CDP, PopAnalyzer) have run into problems with drawer because of overly opinionated or outdated styles on the drawer. The box shadow is too large/heavy, the animations feel sluggish, and the built-in padding has gotten in the way.

PR #2157, opened 2023-05-18
fix(drawer): refresh styles

corykon:
quicker animations, lighter shadows, padding in examples

@andrew-frueh let me know how you feel about these (esp animation speed and shadows). This is what I'm doing in platform apps these days.

re #2028


andrew-frueh (approved):
Great fixes all around - it's kind of crazy to compare the current version side by side with your update.  The current version is SO sluggish, and the box-shadow is pretty weird too.  100% supportive of all these changes.

Issue #889, opened 2019-06-06
Sticky table headers aren't well-documented


isaaclyman:
**Where does our documentation fall short?**

It's hard to figure out how to make a table header/footer sticky.

**What would you add or change?**

A snippet like the following would clear things up:

```html
<!-- Regular table header -->
<tr hc-header-row *hcHeaderRowDef="columnList"></tr>
<!-- Sticky table header -->
<tr hc-header-row *hcHeaderRowDef="columnList; sticky: true"></tr>
```

An example of a sticky table header with a Stackblitz would be even better.

We should also note that (needs testing to confirm) this doesn't work when using `*hcHeaderRowDef` on an `<ng-container>` element, which is the common implementation of multiple header rows. It only seems to work on `<tr>` elements.

jacquie:
I am doing some work using Cashmere tables and needed to use the `when` predicate for the `hcRowDef` directive.  There is no documentation for that either and I had to go and look for it in the cdk documentation.

isaaclyman:
@jacquie do you have the link handy to where you found that information?

jacquie:
It's here https://material.angular.io/cdk/table/api in the `CdkRowDef` section.
I found an example of it's use here: https://stackblitz.com/edit/material2-rc0-6drmzf?file=app%2Fapp.component.html

Issue #685, opened 2019-01-30
Ability to freeze headers on a fixed height table


corykon:
**Describe the problem this feature request is solving**
Something we do in a table on PopBuilder, may or may not be something other teams need.

**Describe the solution you'd like**
Maybe two new parameters for the table:
maxHeight: number
freezeHeaders: boolean

Or, a new wrapper component that holds holds the table.

![freezecolumnheaders](https://user-images.githubusercontent.com/12416432/52016224-37710880-24a1-11e9-86f9-6a215557f533.gif)

andrew-frueh:
Review #734 related to this.  There is currently a `sticky` value on the columnDef.  Should make sure we follow similar syntax for a header row.

andrew-frueh:
#889 is also related

PR #2162, opened 2023-05-22
889 sticky table headers

### LINKED ISSUE #1207: Sticky headers should not overlap table
When using the `sticky` property on the `hc-header-row`, table data is still scrolling underneath the header.
image
The fix I have come up with includes adding a background color of white to `.hc-table th` (or pink).
image
And to remove the still visible top pixel and edge borders use `border-collapse: separate;` on the `.hc-table`.
image

corykon:
We had a few really old issues out there about sticky table headers being buggy and not well documented. Now that we're not supporting IE11, this become easier.

**Component**
image

**CSS-only Examples**
image

re #889 
re #685 
fix #1207

### DIFF
--- a/projects/cashmere-examples/src/lib/table-sticky/table-sticky-example.component.html
+++ b/projects/cashmere-examples/src/lib/table-sticky/table-sticky-example.component.html
@@ -1,31 +1,34 @@
 <div class="example-intro">
     <span>Example of a table with a sticky header & column.</span>
     <span class="example-checkboxes">
+        <hc-checkbox [(ngModel)]="hasGrayBg" [tight]="true">Has gray background</hc-checkbox>
         <hc-checkbox [(ngModel)]="hasBorders" [tight]="true">Has borders</hc-checkbox>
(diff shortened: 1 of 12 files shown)


corykon:
Also have to mention that Github copilot is really fun with generating examples - it magically autocompleted all the stats and facts about the cities for me.

andrew-frueh (changes requested):
Wow, nicely done - there is definitely some CSS alchemy in this one.  I know from past work on the table styles that it can be super finicky to pin down.  And I imagine it's even worse with the `sticky` display type.  Everything is looking good except when you have "borders" off, "header" on, and "column" off.  As you scroll, you can see the left border creeping up beside the top left cell (at least in Firefox).  See the attached screenshot.
 
Screenshot 2023-05-22 at 9 15 07 PM

corykon:
Fixed that issue. Anything else you noticed?
sticky-headers

I added some styles for table on $slate-gray-100 background, since that's pretty standard.
stickytable-bgs

andrew-frueh (approved):

PR #2177, opened 2023-06-19
feat(search-bar): add debounce input

waldnercharles:
I figured `debounce` should be configurable. I was going to add it to the example as well but, this implementation doesn't allow altering the debounce time after `ngOnInit` has fired. 

This didn't seem like a big enough reason for me to alter the implementation, but, let me know if you disagree and I'm sure I can come up with something.

### DIFF
--- a/projects/cashmere-examples/src/lib/search-bar-overview/search-bar-overview-example.component.html
+++ b/projects/cashmere-examples/src/lib/search-bar-overview/search-bar-overview-example.component.html
@@ -1,5 +1,6 @@
 <hc-search-bar
     [placeholder]="placeholder"
+    [debounce]="debounce"
     [disabled]="disabled"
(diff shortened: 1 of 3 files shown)


andrew-frueh (approved):
This looks good to me - but @corykon I'll let you have final say since you built this component.  Thanks for the contribution @waldnercharles !!

waldnercharles:
I added another commit to update the `searchStream` when `debounce` changes so that it can be included in the search-bar-overview-example.

This seems like the best solution for now, but, one drawback is that the initial subscription is cancelled if `debounce` is changed while the subscription is currently "debouncing". (We call `this.destroy$.next()` even if the `searchStream` is in the middle of debouncing.)

corykon (approved):
looks awesome. great add @waldnercharles

PR #2192, opened 2023-07-10
fix(input): error check on markAsTouched

andrew-frueh:
Because there is no way to listen for changes to the `touched` state on an Input, we use the blur event as a proxy.  But in instances where the developer is calling `markAsTouched` or `markAllAsTouched`, a blur event won't be fired, and the developer will expect error state checking to occur.  This update patches the native functions to call Cashmere error state error checking.

### DIFF
--- a/projects/cashmere-examples/src/lib/input-required/input-required-example.component.html
+++ b/projects/cashmere-examples/src/lib/input-required/input-required-example.component.html
@@ -23,3 +23,4 @@
 </div>
 
 <button hc-button title="NgForm Reset" buttonStyle="secondary" (click)="resetForm()">NgForm Reset</button>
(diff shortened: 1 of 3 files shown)


alexdow-hc:
This looks like a perfect solution, great idea! We'll also probably need this applied to the other controls like select, checkbox and radio.

Is there any risk of null reference at `self._updateErrorState();` if we use `ngIf`` to remove the control?

andrew-frueh:
Thanks for the review @alexdow-hc!  I don't think the null reference is a concern, because in this case `self` is a pointer to the component (not the control).  It's necessary because the scope in that monkey patched function changes to the control, so like `this` doesn't point to any component functions anymore.  But if I'm wrong, let me know and we can always add the check :-)

Issue #2205, opened 2023-07-21
add search icon click event


corykon:
Came from a request on a CDP config ui page where we need manually trigger searches, and it felt intuitive to some user to click the search icon.

PR #2206, opened 2023-07-21
feat(searchbar): ability to trigger search on icon click

corykon:
image

re #2205


andrew-frueh (approved):
Nice addition!  I think a lot of users probably do expect to be able to trigger search on an icon click, and I agree that the default behavior should be to have this turned on.

Issue #2244, opened 2023-10-06
Descenders getting clipped on SideNav


andrew-frueh:
**Summary**

Just noticing that on the sidenav component, any values with descenders (g,y,etc) are getting clipped.  Probably just need to allow for overflow or expand the height.

![Screenshot 2023-10-06 at 11 06 38 AM](https://github.com/HealthCatalyst/Fabric.Cashmere/assets/22795893/6dfef13b-4665-4b81-a982-c05dc65441c7)

corykon:
This reminds me that I really want to see the product catalog app. Demo
next 1-on-1?

PR #2246, opened 2023-10-09
fix(sidenav): fix dangling character cutoff

bradcypert:
set overflow to visible for tab name

re: #2244

### DIFF
--- a/projects/cashmere/src/lib/sass/sidenav.scss
+++ b/projects/cashmere/src/lib/sass/sidenav.scss
@@ -89,7 +89,7 @@ $sidenav-width: 260px !default;
     white-space: nowrap;
 
     .hc-sidenav-tab-name {
-        overflow: hidden;
+        overflow: visible;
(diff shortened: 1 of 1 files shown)


bradcypert:
There's a chance that more work is required here. Upon inspection it seems that there's two style declarations for the same class name, one for the standard scss and one for component specific css. Just wanted to call out that the contents of this style block is not in sync between the two. Maybe that's okay, maybe not but I wanted to mention it :)

corykon (changes requested):
Thanks for submitting Brady! We want to make sure we handle cases where the tab name is too long by truncating and adding ellipsis. Setting overflow to visible will break this. 

I would recommend setting the overflow back to hidden, and then adding something like this: `line-height: 1.5;`

bradcypert:
@corykon great call! I've updated the scss to reflect your suggestions

corykon (approved):

PR #2255, opened 2023-10-20
fix(input directive): fix - error not resetting when validation=change

### LINKED ISSUE #2186: Input validation types have different reset behavior
- When using "on blur", a reset of the form gets rid of the red error message.
- When using "on change" a rest of the form **doesn't** get rid of the red error message.

form-reset

I also found it a bit confusing that there were two reset buttons... seems like those are two different methods of resetting the form, but it would be cool to explain the difference between the two, and why you'd use one over the other.

JakeSaintG:
fix #2186

### DIFF
--- a/projects/cashmere-examples/src/lib/input-required/input-required-example.component.html
+++ b/projects/cashmere-examples/src/lib/input-required/input-required-example.component.html
@@ -6,7 +6,8 @@
             <hc-error>A valid email address is required</hc-error>
         </hc-form-field>
     </div>
-    <button hc-button title="Reset Form" buttonStyle="secondary" type="reset">Form Reset</button>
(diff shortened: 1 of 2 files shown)


JakeSaintG:
A lot of the minor tweaks like spacing came from running `npm run prettier`. The bulk of the fix involves altering how error state is checked with regards to "onChange". When in the `newState` variable, it was always coming back as truthy. Moving this check to the input event preserved the error state updating on typing but allowed the reset event to remove it.

Additionally, Cory mentioned in the issues that the example was a little confusing with the two types of reset buttons. When deciding between one that handled resetting the default way (type=reset) and Angular's way to resetting, Cory and I chose to keep the later around.
image

andrew-frueh (approved):
Thanks @JakeSaintG and @corykon for the work on this one - these error states are tricky to manage!

Issue #2236, opened 2023-09-27
Angular 16


corykon:
Angular v16 enter LTS on Nov 11, 2023. Only about a month away!

https://angular.io/guide/releases#support-policy-and-schedule

PR #2258, opened 2023-11-09
2236 angular 16

corykon:
**Gets cashmere up to Angular 16.** (v16 enter LTS in a couple days)
- Updated our dependencies as per usual and made sure everything is working
- Updated a bunch of unit tests to get rid of all warnings. Removed tests that were turned off, but we can bring these back from git history if we ever want them back.
- Biggest gotcha was the ngx-slider dependency, which does not work with angular 16. Had to move to [a fork of that project](https://github.com/danisss9/ngx-slider) that does.

**Longer version of the story:** Angular 16 requires you to use Ivy, and ngx-slider isn't compatible. The [ngx-slider project](https://github.com/angular-slider/ngx-slider) hasn't been updated in over a year, and the maintainers haven't yet responded to requests to upgrade. [Someone forked and made the changes needed for it to work](https://github.com/angular-slider/ngx-slider/issues/355), but for whatever reason hasn't requested to merge as far as I can tell. I'm open to other approaches for getting over this hurdle (submitting a PR to the original ngx-slider, or changing to a new dependency), but jumping to the fork seemed easiest for now. The big risk in my mind would be having to jump off this fork and then onto something else later. That would be annoying, especially for any apps using `hc-slider`.

re #2236


andrew-frueh (approved):
Thanks so much for jumping on this.  Nice work fixing all those unit tests too.  Glad this upgrade went more smoothly than 15 did 😅

PR #2264, opened 2024-01-17
fix(draglist): make height flexible

isaaclyman:
Previously hard-coded to `max-height: 300px`. Now by default grows to content height, but a max-height can be set on the root element and each column will scroll independently.

### DIFF
--- a/projects/cashmere/src/lib/drag-list/drag-list.component.html
+++ b/projects/cashmere/src/lib/drag-list/drag-list.component.html
@@ -13,69 +13,71 @@
                 <div class="preserve-whitespace" style="white-space: pre">{{ targetTooltip }}</div>
             </hc-pop>
         </div>
-        <div *ngFor="let assignment of _modifiedAssignments" class="hc-drag-list-assignments">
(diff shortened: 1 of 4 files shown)


projects/cashmere/src/lib/drag-list/drag-list.component.html
```diff
@@ -13,69 +13,71 @@
...
-                class="hc-drag-list-assignment"
-                (dragenter)="_dragenter($event)"
-                (dragleave)="_dragleave($event)"
-                *ngIf="!!assignment.assignment && !assignment.locked"
-                [attr.data-target-id]="assignment.target[targetIdField]"
-                [attr.data-option-id]="assignment.assignment[optionIdField]"
-                (drop)="_makeAssignment($event)"
-                (dragover)="_dragover($event)"
-            >
+        <div class="hc-drag-list-targets-container">
```
isaaclyman, line 16: The only actual change to this file was the addition of this wrapper div to allow the body of the list to scroll independently from the header.

anderslyman (approved):

andrew-frueh (approved):

PR #2275, opened 2024-02-16
docs(app-branding-demo): adding an app title generator for navbar

sheaivey:
I created a tool to generate the application title text people frequently ask for.

I was not sure the exact font used to create these titles. But I found one that is pretty close. `Carnas Light`
It should be pretty easy to plug in the actual font-family found in `app-branding-demo.scss` and then updating all the references to use it.

The new title generator can be found in Foundations -> Brand -> App Title Generator

Screen Shot 2024-02-15 at 5 54 06 PM

### DIFF
--- a/src/app/foundations/app-branding/app-branding-demo.component.html
+++ b/src/app/foundations/app-branding/app-branding-demo.component.html
@@ -1,0 +1,52 @@
+<div class="demo-content">
+    <h1>App Title Generator</h1>
+    <h6>Last updated February 15, 2024</h6>
+
+    <hc-tile>
+        <h5 id="overview">Overview</h5>
+        <p>
+            The App Title Generator is a tool to create the application
+            title followed just after the Health Catalyst icon in the navbar.
+            For example "Cashmere" in the top navbar of this page is the
+            application title.
+        </p>
+        <p>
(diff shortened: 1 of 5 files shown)


sheaivey:
Updating to use `Flexo-Regular` font. It appears to be the actual font used in the past.

corykon:
@sheaivey THANK YOU! This is awesome. Looks like an great solution.

Couple points of feedback
- The primary normal text should be a slightly gray color - #f4f3f3. The secondary bold text is just plain old #fff
- Would it be possible to base64 encode as png instead of a gif? The analytics teams always ask for pngs, but a gif may be just fine. I'll ask ASO about that
- An awesome bonus would be if we could spit out SVGs for webapps, but that sounds like a much taller order.

aaronbrethorst:
This repo purports to convert text to svg paths: https://github.com/shrhdk/text-to-svg

sheaivey:
- I updated the normal text color to be `#f4f3f3` and the bold text to `#ffffff`. 
- I also added output format selector. Currently canvas only supports png or jpeg.

It would be fun in the future to give the `text-to-svg` library a try that @aaronbrethorst mentioned.

corykon (approved):

PR #2276, opened 2024-02-17
fix(login): remove font-family

andrew-frueh:
Shouldn't define the font-family here - was causing some odd Times font fall-backs.  Should just be using what's defined in the app's `body` style.


aaronbrethorst (approved):
lgtm 👍

PR #2282, opened 2024-02-29
fix(tab-set): fix issue on selectTab

juanslinger:
Prevent an error from being thrown when the activeTab is null or undefined.


corykon (approved):
I just noticed this PR that's been hanging out awhile. My apologies for not seeing it sooner. Feel free tag @andrew-frueh or I on these!

andrew-frueh (approved):

PR #2293, opened 2024-04-30
feat(sidenav): use anchor tags for native context menu support

saleryan:
There are a lot of formatting changes but all I did is change the tabs in the side Nav to use anchors.

image


corykon (approved):
Great addition, thanks @saleryan! I think it'll be good to get one more review from @aaronbrethorst or @andrew-frueh since I worked on this with you. After that, feel free to merge

PR #2367, opened 2024-12-19
fix(example creation script): alter example project's package handling

JakeSaintG:
# Cashmere New Example Issues

## Default Peer Dependencies
It was noticed that the `npm run new:example` command failed with error "Could not run Object.Keys over undefined". 

image

---
I checked the target package.json that and there was no peerDependencies property for it to read from. A default empty array was added in the case of a non-existent or empty peerDependencies object. After that change was added, the script moved along and created a new example. As a test, the component was created with the `is-number` package.

image

image

image

image

## Issues With Subsequent Runs
After the creation of the "foobar" component, it was noticed that the new-example script made few alterations to the example project's package-lock file. 
image

Subsequent runs of the new-example script or `npm i` caused the below error.
image

The error message led to me updating the version of cashmere that was in the package-lock, reverting the other changes, and trying again. I ran into the same issue where the **script would work the first time but fail on the next run.**

After some tinkering, it was noticed that **removing the packages object from the package-lock.json was needed to avoid the package conflicts.**

## Testing
The foobar example with the package `is-number` (above) was left in the project as a first run.

Running another example creation, "foo", with the `is-odd` package succeeded.
image
- Note that `is-number` is listed on the "specify packages" line.

Running **a subsequent** example creation, "bar", with the `is-even` package succeeded.
image
- Note that `is-number` and `is-odd` are listed on the "specify packages" line.

---
These examples were tested by temporarily implementing them and pulling in the downloaded packages.
image

Output
image

## Additional thoughts
I am not 100% certain that this is the best possible solution to this problem. I have tested it locally and it appears to be functional on first run as well as subsequent runs. I'm open to feedback and comments!

### DIFF
--- a/projects/cashmere-examples/package-lock.json
+++ b/projects/cashmere-examples/package-lock.json
@@ -5,13 +5,11 @@
     "requires": true,
     "dependencies": {
         "@healthcatalyst/cashmere": {
-            "version": "12.7.2",
-            "resolved": "https://registry.npmjs.org/@healthcatalyst/cashmere/-/cashmere-12.7.2.tgz",
-            "integrity": "sha512-SoMceeRtZYQByBbBsOZEwQLZcWsDu7PnDDDQ3rzKVwhr2GgDsQUNOA8VleTSKK5dEpEAwMbgwL+dGV1LET3Kzw==",
+            "version": "17.2.3",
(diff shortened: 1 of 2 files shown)


corykon (changes requested):
Generally speaking it's not a good idea to manually edit a pack-lock.json file. It's really meant to only ever be edited by npm. 

I really feel like it would be better to find out WHY there are dependencies conflicts and try to solve that, rather than stepping around the problem by tinkering with the lock file

Issue #2377, opened 2025-03-13
Angular v18 (due 11/19/24)


corykon:
Angular 18 entered LTS on 11/19/24. We should update Cashmere to v18 ASAP, as we're behind by several months now.

https://angular.dev/reference/releases

PR #2384, opened 2025-03-20
angular 18

corykon:
re #2377


corykon:
We've got a bunch of extra commits here because I merged master back into dev, so it's a little noisier! My apologies. Nothing that should cause issue.

Issue #2279, opened 2024-02-26
HcIcons font import update


guille-gomez:
**Describe your idea**

In Angular 17 we have the option of using new builders `esbuild`, `application`. 

**What problem are you trying to solve?**

When trying to use them we are seeing some errors in cashmere.css like
`
 Could not resolve "node_modules/@healthcatalyst/cashmere/hcicons/hcicons.eot"
`
which happen here
`
@font-face {
  font-family: "hcicons";
  src: url("node_modules/@healthcatalyst/cashmere/hcicons/hcicons.eot");
  src: url("node_modules/@healthcatalyst/cashmere/hcicons/hcicons.eot") format("embedded-opentype"), url("node_modules/@healthcatalyst/cashmere/hcicons/hcicons.woff2") format("woff2"), url("node_modules/@healthcatalyst/cashmere/hcicons/hcicons.woff") format("woff"), url("node_modules/@healthcatalyst/cashmere/hcicons/hcicons.ttf") format("truetype"), url("node_modules/@healthcatalyst/cashmere/hcicons/hcicons.svg") format("svg");
  font-weight: normal;
  font-style: normal;
}
`
Replacing `node_modules/@healthcatalyst/cashmere/` with `@healthcatalyst/cashmere/` solves the issue and it works with the current builder.

![image](https://github.com/HealthCatalyst/Fabric.Cashmere/assets/42807289/81354f94-138c-4128-a175-f00ab8752447)

PR #2386, opened 2025-03-26
build(static css): remove node_modules ref in static css file

corykon:
re #2279


guille-gomez (approved):
This looks like is going to fix the issue. Thanks @corykon

